PR #126, opened 2026-05-04
feat(core): add support for attributes

Guikingone:
Adds support for PHP 8.0 attribute syntax (#[Name], #[Name(args)], #[A, B], stacked groups, fully-qualified names) at every declaration site PHP allows them: classes, interfaces, traits, enums, enum cases, top-level functions, methods, and properties.

`Lexer`:
- new Token::AttrOpen for the literal "#[" sequence
- bare '#' (not followed by '[') is now treated as a PHP-style line comment, matching PHP's lexer; the existing // and /* */ paths are unchanged

`Parser`:
- new attribute helper parser/attributes.rs walks one or more groups, optional argument lists, and qualified names
- consume_attribute_lists is invoked at the top of parse_stmt, ahead of every member in class/trait/interface bodies, ahead of every enum case, and at the top-level loop before extern blocks

For v1 the parser only validates the syntactic shape — attributes are not preserved in the AST, so codegen runs unchanged and runtime introspection (ReflectionAttribute) is not yet supported. This is the minimum needed to compile real PHP code that uses attributes for DI or documentation, including Symfony's #[Required] on ClockAwareTrait.

Tests cover lexer (#[ vs # comment, qualified names, multi-attribute groups), parser (attributes parse to the same AST as the bare declaration), codegen (attributed program runs identically to the bare one), and error reporting (unterminated, empty, malformed attribute groups). Adds examples/attributes/ and updates docs/php/classes.md.


Guikingone:
@nahime0 I'm currently working on the local rebase / optimisations for this branch, should be ready for review tomorrow 👍🏻

nahime0:
Thanks @Guikingone, I'm closing another area, in a few hours I'll get back to this and check out your PRs.

PR #154, opened 2026-05-09
feature/first class callable targets

nahime0:
- **feat: support method first-class callables**
- **docs: document callable target support**


nahime0:
Needs to be completed after merging #155

PR #276, opened 2026-05-21
fix: prevent non-definite control-flow assignments from polluting TypeEnv

nahime0:
### Motivation
- The type checker previously used the same mutable `TypeEnv` when checking `if`, `while`, `for`, and `foreach` bodies, causing assignments inside non-definite paths to be treated as definitely-initialized afterward and enabling uninitialized stack reads in generated code. 
- This could yield information disclosure or crashes (e.g., `<?php if (0) { $s = "secret"; } echo $s;`).

### Description
- Update the control-flow checker so bodies are validated against cloned environments instead of mutating the outer `TypeEnv`, preventing branch/loop-local assignments from leaking; changes live in `src/types/checker/stmt_check/control_flow.rs` and affect `if`/`elseif`/`else`, `while`, `for`, `foreach`, and related body checks.
- Use `env.clone()` and pass the clone to `check_body`/`check_break_continue_target_body` for non-definite bodies to preserve the caller environment.
- Add three regression tests in `tests/error_tests/type_system.rs` asserting variables assigned only inside `if (0)`, `while (0)`, and `for (; 0; )` remain undefined after the control statements.

### Testing
- Ran `cargo build`, which completed successfully.
- Ran the focused test group via `cargo test assignment_is_not_definite`, and the three new regression tests passed.
- Ran `git diff --check` locally to verify no comment/alignment issues (no problems reported).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a0d31158088832688eff80f24bbb972)

### DIFF
--- a/src/types/checker/stmt_check/control_flow.rs
+++ b/src/types/checker/stmt_check/control_flow.rs
@@ -78,7 +78,8 @@ impl Checker {
                         "by-reference foreach over Iterator/IteratorAggregate objects or iterable-typed values is not supported; use an array source or remove &",
                     ));
                 }
-                let errors = self.check_break_continue_target_body(body, env);
+                let mut body_env = env.clone();
+                let errors = self.check_break_continue_target_body(body, &mut body_env);
                 if errors.is_empty() {
                     Ok(())
                 } else {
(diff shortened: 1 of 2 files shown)


nahime0:
This is fix is too aggressive. Will fix later on.

PR #327, opened 2026-06-02
feat(core) array builtin parity + indexed-scalar inputs and ownership fixes

Guikingone:
15 PHP array builtins, dual-arch (ARM64/x86_64), tested and GC-clean: array_is_list, array_key_first/last, array_replace, array_replace_recursive, array_diff_assoc, array_intersect_assoc, array_merge_recursive, array_walk_recursive, array_find/any/all, array_udiff/uintersect, array_multisort.

Correctness fixes:
- array_merge_recursive: persist combined-list string values (fixes a use-after-free when the temporary scalar wrappers are released).
- Hash-based builtins persist string values into results instead of incref-sharing (fixes a latent use-after-free when an input is freed before the result).

Scalar indexed-array inputs for the 5 hash-based builtins via the new __rt_array_to_hash converter; result key/value widen to Mixed for heterogeneous inputs.

Validated on macOS ARM64, Linux x86_64, and Linux ARM64.


nahime0:

PR #333, opened 2026-06-04
feat(core): continue support for streams

Guikingone:


Guikingone:
Hi @nahime0 👋🏻 

Don't know if you planned to work on the CI but maybe it could be a good idea to split the CI in multiple parallels steps that triggers specific tests, would be "faster" as the test suite evolve 🤔

nahime0:
Hi @Guikingone!

That's a good suggestion. As the test suite grows, parallelizing parts of the CI will likely become necessary. I'll take a look at it. If you have any specific ideas or examples, I'd love to hear them.

Guikingone:
One approach would be to split "per topic": `codegen`, `optimize`, etc, not sure how Rust can split it that was during tests, found solutions locally when running tests but it was "at runtime", not in CI, I'll take a look on my side to see if I find a way to dot it 😅

nahime0:
Hi @Guikingone I’ve tried an approach with nextest and sharding. The required test time is considerably lower than before. It's already on main.

PR #343, opened 2026-06-05
feat(datetime): Add support for DateTime, DateInterval, DatePeriod, strtotime

Guikingone:
Builtin date/time classes implemented as synthetic PHP over date()/mktime()/ strtotime(): DateTimeInterface, DateTimeZone, DateTime, DateTimeImmutable (construct/format/getTimestamp/getTimezone/set*/add/sub/modify/diff), DateInterval (ISO-8601 parse, calendar y/m/d fill from diff(), format()), and a DatePeriod (start, interval, end) Iterator.

date()/gmdate(): all non-timezone format specifiers + backslash escapes + ISO-8601 week (W) and ISO year (o). gmdate() is the UTC sibling of date().

strtotime(): full relative/absolute grammar — `@epoch`, American M/D/Y slash dates, textual dates (D Month Y / Month D, Y), "first/last day of <mod> month", nth weekday of month, plus the 2-arg base-timestamp form; DateTime::modify() builds on it.

Includes the prerequisite codegen fixes the above depend on: Mixed/Union operand unboxing in mktime()/intdiv()/the remaining int-arg builtins, cast-vs-binary operator precedence in the parser, and array_fill() storing every element for string values.

Docs (docs/php/datetime.md) + a runnable example added.

Also excludes the datetime classes from the dynamic `new $var()` AOT dispatcher candidate list: they are synthetic builtin classes whose methods are emitted on demand, so referencing their constructors unconditionally would leave undefined symbols at link time (consistent with how other on-demand builtins are handled).


nahime0:

Guikingone:
@nahime0 Sorry for the delay, I noticed some gaps in the current implementation after migrating it to EIR, PR should be ready by the end of week 🙂

nahime0:
Absolutely no problem @Guikingone. Take your time.

Guikingone:
@nahime0 And we're good for review, sorry for the delay 😅

PR #347, opened 2026-06-08
fix(core): audit-remediation suite, re-ported to the active EIR backend

Guikingone:
This branch remediates the findings from a full compiler audit (correctness, memory safety, PHP compatibility, performance), grouped by theme.

It was originally authored against the legacy AST backend. Since **EIR is now the only active backend**, every fix has been **re-ported to EIR** so it actually takes effect on emitted code — the audit commit carry both the original legacy fix and its EIR implementation, and a final commit regenerates the builtin docs/registry to match.

## Fixes (by theme)

- **Critical** — foreach-var heap release gated on ownership (double-free); x86_64 NaN-compare parity guard; `PHP_INT_MIN % -1` fold panic.
- **High** — cast precedence + name resolution; Mixed/Union coercion for math builtins / `floatval` / `min`/`max` / `is_a`/`is_subclass_of`; `substr` negative length; float sort by numeric value; variadic `array_merge`/`diff`/`intersect`; `in_array`/`array_search` `$strict`; `SORT_*` flags.
- **Medium** — DCE keeps fatal-capable builtins; division/modulo-by-zero fatals; `array_merge` string-array corruption; infer↔emitter type alignment; `round()` `$mode` (HALF_UP / HALF_EVEN); `array_slice` `preserve_keys`; `number_format` multi-byte separators.
- **Low** — `array_sum`/`array_product` over float arrays; `settype`; `str_split`; negative-key auto-increment; NaN-ordering docs.
- **Perf** — `@src` comment allocation; `is_subclass_of` clone.
- **Feature** — `array_map($cb, $a, $b)` two-array form (int + string element arrays, every callback shape: named fn, capture-less/capturing closure, closure-holding variable).

## EIR re-port notes

A recurring root cause worth calling out for reviewers: EIR lowering (`src/ir_lower/`) computes its **own** builtin result-type inference, independent of the type checker. So several fixes (`array_sum` float, `array_slice` `preserve_keys`, two-array `array_map`) had to correct the result type **there** as well as in the checker — otherwise the value flowed as `Int`/`Mixed`, the backend took the wrong code path, and `echo`/`implode` printed garbage. The legacy emitters under `src/codegen/` are the frozen backend and are left as-is.

## Verification

Build is warning-free and whitespace-clean. Each re-ported feature's focused suite passes on **macOS-aarch64** locally (division 17, strict 55, min 154, round 78, number_format 11, array_merge 7, sorting 13, is_a 14, instanceof 27, slice 15, array_map 53, aggregates 8, substr 19, floatval 3) with no cross-feature regression (broad `arrays::` 286/0).

**Linux x86_64 / arm64 are validated by this PR's codegen matrix.** The x86-only paths that local macOS testing can't exercise (the `__rt_mixed_unbox` payload register `rdi`/`x1`, the `number_format` ptr+len argument ABI, the target-aware argument-register marshalling for `array_map2`) were reviewed explicitly against the established runtime ABIs.

Rebased clean onto the latest `main`.

### DIFF
--- a/docs/internals/builtins/array/array_all.md
+++ b/docs/internals/builtins/array/array_all.md
@@ -10,7 +10,7 @@ sidebar:
 ## Where it lives
 
 - **Signature**: [`src/builtins/array/array_all.rs`](https://github.com/illegalstudio/elephc/blob/main/src/builtins/array/array_all.rs)
-- **Lowering**: [`src/codegen_ir/lower_inst/builtins/arrays.rs`:1536](https://github.com/illegalstudio/elephc/blob/main/src/codegen_ir/lower_inst/builtins/arrays.rs#L1536) (`lower_array_all`)
+- **Lowering**: [`src/codegen_ir/lower_inst/builtins/arrays.rs`:1707](https://github.com/illegalstudio/elephc/blob/main/src/codegen_ir/lower_inst/builtins/arrays.rs#L1707) (`lower_array_all`)
 - **Function symbol**: `lower_array_all()`
 
 
--- a/docs/internals/builtins/array/array_any.md
+++ b/docs/internals/builtins/array/array_any.md
@@ -10,7 +10,7 @@ sidebar:
 ## Where it lives
 
 - **Signature**: [`src/builtins/array/array_any.rs`](https://github.com/illegalstudio/elephc/blob/main/src/builtins/array/array_any.rs)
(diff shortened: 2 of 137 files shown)


nahime0:

nahime0:
Closing this draft as superseded.

The audit items already landed later, usually more correctly (`#678` and the later ownership/foreach PRs). This SHA still targets the deleted `codegen_ir` / `codegen/runtime` layout and the old `builtin!` API. The PR body also advertises files that are not in `1f50514b`.

Please do not rebase this branch. Remaining completeness gaps (`array_map` 3-arg, `SORT_*`, float `array_sum`, …) belong in new focused PRs on current builtins/runtime, one surface at a time.

PR #353, opened 2026-06-10
feat: --emit cdylib produces loadable shared libraries on all supported targets

luthermonson:
## Summary

Implements the `--lib` flag + `.so` / `.dylib` output items from the
[v0.27.x "Shared and static libraries (C ABI)" track](https://github.com/illegalstudio/elephc/blob/main/ROADMAP.md#v027x--shared-and-static-libraries-c-abi).
Lets embedding hosts (the original ask is from [ePHPm](https://github.com/ephpm/ephpm))
`dlopen` an elephc-compiled `.so` / `.dylib` and call individual PHP
functions through a stable C ABI.

The scope is intentionally narrow — a working v1 that exercises every
piece of the integration path end-to-end (CLI → codegen → linker →
dlopen → call → return), but only for the marshaling shapes that need no
new runtime machinery. Auto-generated headers, pkg-config, multi-file
libraries, `.a` archives, string-return, exception bridging, and the rest
of the v0.27.x checklist remain follow-ups.

### PHP-side surface

```php
#[Export]
function validate_token(string $token): int {
    return strlen($token) >= 8 ? 0 : 1;
}
```

Functions marked `#[Export]` (bare or fully-qualified `\Elephc\Export`)
are exposed under their unmangled PHP name with C-ABI linkage. v1
parameter types: `int`, `float`, `bool`, `string`. v1 return types:
`int`, `float`, `bool`, `void`. Anything outside that set is rejected at
typecheck time with a localized error pointing at the offending parameter.

### C-side surface

```c
int32_t     elephc_init(void);         // returns 0 on success
void        elephc_shutdown(void);
const char *elephc_last_error(void);   // NULL when no error pending
void        elephc_free(void *ptr);    // for elephc-owned host-returned ptrs
```

Plus one exported symbol per `#[Export]` function, with the PHP function
name as the symbol name (`_validate_token` on macOS per Mach-O underscore
mangling, `validate_token` on Linux ELF).

### CLI

```
elephc --emit cdylib examples/cdylib/auth.php
# → examples/cdylib/libauth.so (Linux) or libauth.dylib (macOS)
```

`--emit executable` (the default) preserves today's standalone-binary
behavior exactly — byte-identical asm output for unchanged programs.

## Architecture

The implementation lands in three logical commits to make each piece
reviewable on its own:

### Commit 1 — `codegen: route global symbol references through the GOT in PIC mode`

Adds a `pic_data_refs: bool` flag to `Emitter`. When set, the four
symbol-access helpers in `codegen/abi/symbols.rs`
(`emit_load_symbol_to_reg`, `emit_store_reg_to_symbol`,
`emit_store_zero_to_symbol`, `emit_symbol_address`) route through the
existing extern-symbol helpers, which already encoded the GOT-relative
sequence: `@GOTPCREL[rip]` on x86_64, `adrp :got: + ldr :got_lo12:` on
AArch64.

`generate_runtime_with_features` grows a `_pic` companion so the runtime
object can be assembled twice — once for executables (cached as today),
once PIC for cdylib. The two builds produce different asm text, so the
existing asm-hash cache key naturally separates them into distinct
cached object files.

Executable emission is unchanged: the flag defaults off, every existing
call path bypasses the GOT, asm output is byte-identical to pre-change.

### Commit 2 — `feat: --emit cdylib produces loadable shared libraries (v1, AArch64-only)`

The main feature commit:

- **CLI** (`src/cli.rs`): adds `--emit <kind>` taking `executable` /
  `cdylib` plus a few aliases (`exe`/`bin`, `dylib`/`shared`). The
  `Emit` enum is defined in `codegen/mod.rs` so both binary and library
  consumers can reach it.
- **Pipeline** (`src/pipeline.rs`): runs an exports-collection pass
  after typechecking, threads the `Emit` value into codegen and the
  linker, computes the output filename (`lib<stem>.{so,dylib}` for
  cdylib), warns when `#[Export]` is used in executable mode, and
  errors out cleanly when cdylib is requested on x86_64.
- **Linker** (`src/linker.rs`): in cdylib mode, macOS invokes
  `ld -dylib -install_name @rpath/<file>` (drops the `-e _main`
  requirement), Linux invokes `gcc -shared` and skips the `-static`
  flag that would otherwise apply when no extra libs are present.
- **Codegen entry** (`src/codegen/mod.rs`): swaps in `Emitter::new_pic`
  for cdylib emission so the new symbol-access flag fires, then after
  user functions are emitted calls into a new `codegen::cdylib` module
  for the export trampolines + lifecycle stubs.
- **`src/codegen/main_emission.rs`**: when `emit == Cdylib`, early-returns
  to `finish_user_asm` so the artifact has no `_main` label and no
  `exit(0)`. Top-level executable PHP statements are silently dropped
  (they would have lived inside `_main`); a future typecheck-time warning
  for them is straightforward but out of scope.
- **`src/codegen/cdylib.rs`** (new): emits one trampoline per exported
  function plus the four lifecycle symbols. The trampoline is a single
  tail-branch into the internal `_fn_<name>` symbol on both
  architectures — elephc's existing internal calling convention already
  routes integer/scalar params through the same SysV/AAPCS integer-arg
  registers a C caller populates, and PHP `Str` params arrive as
  `(ptr, len)` in two consecutive integer registers, exactly matching
  `const char*, size_t` from C. No register shuffling is needed for any
  v1 signature. Lifecycle stubs: `elephc_init` returns 0 (allocator state
  is BSS-zero-init from the runtime object), `elephc_shutdown` returns,
  `elephc_last_error` returns NULL, `elephc_free` returns.
- **`src/exports.rs`** (new): walks the post-typecheck program AST,
  detects `#[Export]` (last-segment match so both bare and
  `\Elephc\Export` work), looks up the resolved `FunctionSig`, and
  rejects anything outside the v1 marshaling set with a per-parameter
  error. Variadic and by-reference exports are rejected for v1.
- **`src/runtime_cache.rs`**: takes a new `pic` argument and forwards
  to `generate_runtime_with_features_pic`. The cache key is unchanged
  (the runtime-asm-hash differs between PIC and non-PIC builds so the
  two cached objects land in different files automatically).

### Commit 3 — `examples: add cdylib end-to-end PHP -> C dlopen demo`

Drops `examples/cdylib/{auth.php, host.c, README.md}` exercising both
exported functions through `dlopen` + `dlsym`. The harness validates the
v1 marshaling contract concretely:

```
elephc cdylib demo OK: add_i64(40,2)=42, validate_token long=0 short=1
```

Doubles as a CI smoke test once a workflow wires it up.

## What I am asking you to validate

I built this on a Windows host with WSL Linux for verification, so the
matrix coverage I could actually run is partial:

| Target | `cargo check` | cdylib build | dlopen + call |
|---|:-:|:-:|:-:|
| Windows host build of compiler | yes | n/a | n/a |
| linux-x86_64 (in WSL) | yes | errors cleanly per design | n/a |
| linux-aarch64 | yes | **unverified** (no cross toolchain) | **unverified** |
| macos-aarch64 | yes | **unverified** (no macOS host) | **unverified** |
| macos-x86_64 | yes | errors cleanly per design | n/a |

The AArch64 paths use `adrp + add` / `adrp :got:` which are naturally PIC,
and my emitter-flag plumbing routes through the existing extern-symbol
helpers that AArch64 already uses elsewhere in the compiler. So my prior
expectation is that AArch64 just works — but your CI is the first real
test. The two most likely places for trouble:

1. The runtime might have hand-written AArch64 assembly using direct
   `adrp` (not `adrp :got:`) for cross-section globals. Same fix
   pattern as the x86_64 helpers if so.
2. The export trampolines (`b _fn_<name>`) assume same-object linkage.
   If the linker wants PLT routing under default symbol-interposition
   rules, easy to add `-Bsymbolic` or switch the branch to go through
   the PLT.

Either would be a small follow-up, not a structural rework.

## The x86_64 caveat (and why it is a v1 cliff, not a bug)

The runtime emits roughly 700 hand-written x86_64 instructions with
direct `[rip + sym]` addressing to its own globals
(`_rt_diag_suppression`, `_socket_key_str`, `_concat_off`, ...). Those
land as `R_X86_64_PC32` against cross-section symbols, which the dynamic
linker cannot resolve at `dlopen` time. `-Wl,-z,notext` does not save us
— this is a fundamental shared-object constraint, not a warning that
got promoted.

Routing all 700 sites onto the GOT helpers in `codegen/abi/symbols.rs`
is the right fix and the path is now in place after commit 1, but it is
a mechanical sweep across 30 codegen files that I felt was wrong to
bundle into a "structural feature foundation" PR. Pipeline gates
x86_64+cdylib with a clear error pointing at the AArch64 targets:

```
--emit cdylib is not yet supported on x86_64 targets (the runtime object
emits direct RIP-relative references that cannot resolve in a shared
library). Use --target macos-aarch64 or --target linux-aarch64 for v1,
or stay on --emit executable for x86_64 targets.
```

Happy to do that runtime sweep as a follow-up PR if you would like, or
you may prefer to drive it yourself given how much of the runtime it
touches.

## Out of scope for v1 (intentional)

- x86_64 cdylib (see above)
- C header auto-generation (`--emit header`)
- pkg-config generation
- Multi-file library compilation
- Symbol visibility flags beyond export/non-export
- `.a` static archive output (the third item in the roadmap line)
- String return values (waiting on `elephc_free` doing real work)
- Array / object / closure / `null` marshaling
- Exception propagation
- Thread-safety guarantees beyond single-threaded hosts
- A typecheck-time warning when `#[Export]` is used with a top-level
  executable PHP statement (silently dropped today)

## Background

Original motivation thread: [ephpm/ephpm](https://github.com/ephpm/ephpm)
is a single-binary PHP application server in Rust that embeds the Zend
Engine via FFI. With cdylib support, ephpm can dlopen elephc-compiled
modules and call them as host functions from regular PHP — turning
"compile your hot paths with elephc" into a one-line config change for
application authors. v1 unlocks that integration; the rest of v0.27.x
turns it into a polished story.

Happy to iterate on any of this — scope, attribute name, file layout,
trampoline shape, whichever calls feel off. You know the compiler far
better than I do; treat this PR as a starting position, not a
finished design.


luthermonson:
i also cherry picked these changes against EIR and fixed the merge conflicts https://github.com/illegalstudio/elephc/compare/feature/the-eir...luthermonson:elephc:feat/emit-cdylib-on-eir?expand=1

nahime0:
Hi @luthermonson 

Thanks. I've looked at the implementation and it's solid.

Right now I'm doing the sweep to remove the x86 gate. Once that's done I'll clean up the docs and update the ROADMAP.

My plan is to complete the sweep and get full target support before merging. Once that's done and I'm happy with the result, I'll probably merge it so we have a working prototype to experiment with dynamic libraries and gather feedback for the broader track.

Thanks

nahime0:
## Scope update: x86_64 support landed — no longer AArch64-only

The original description above reflects the v1 scope (AArch64-only, with the GOT sweep listed as a follow-up). The branch has since been completed so that `--emit cdylib` works on **every supported target**: macOS aarch64, Linux aarch64, and Linux x86_64. What changed:

### GOT sweep (the prerequisite called out in the original write-up)

- ~830 hand-written global data references converted onto the `codegen::abi::symbols` helpers: ~719 x86_64 `[rip + sym]` sites, ~105 AArch64 `adrp`/`add_lo12` pairs, plus the `format!`-interpolated sites converted by hand. In PIC mode they all route through the GOT; references to local labels stay direct (already PIC-legal).
- Safety net: the runtime assembly in non-PIC (executable) mode is **byte-identical** to the pre-sweep output on all three targets, so executable codegen is provably unchanged.
- The AArch64 pair conversion matters beyond x86_64: those sites bypassed the `pic_data_refs` flag, producing direct `ADR_PREL_PG_HI21` relocations against preemptible globals — which means the linux-aarch64 cdylib path was broken in v1 as well (it only worked on Mach-O, where same-image references cannot be preempted). It is now GOT-routed and covered by tests.

### Clobber-free PIC helper paths

- PIC loads resolve the GOT entry through the destination register itself; stores borrow a push/pop-protected scratch. Swept call sites keep their exact non-PIC register footprint, so no per-site liveness audit was needed.
- Fixed a latent bug in `emit_store_zero_to_symbol`: the x86_64 PIC path zeroed r11 and then reused it as the GOT scratch, storing the symbol address instead of zero.
- New helpers for the remaining shapes: `emit_cmp_reg_to_symbol`, `emit_dec_symbol`, `emit_store_imm_to_symbol`, `emit_load_symbol_to_reg_via_page`.

### Hidden visibility on ELF (`src/codegen/visibility.rs`)

Every internal `.globl`/`.comm` symbol in PIC runtime and cdylib user assembly gets a `.hidden` directive on Linux; only the `#[Export]` trampolines and the four lifecycle entry points keep default visibility. This is a correctness requirement, not just hygiene: with default visibility, two elephc cdylibs loaded into one process would bind their internal runtime state (concat buffers, JSON state, I/O handles) to whichever `.so` loaded first via dynamic-symbol preemption.

### Tests (the branch previously had none)

- `tests/cdylib_tests.rs`: compile `#[Export]` functions → dlopen from a C host → assert behavior; ELF dynamic-symbol-table assertions (public ABI only); v1 scalar-set rejection diagnostic; executable-mode ignore warning.
- 7 ABI unit tests pinning the PIC instruction sequences, 3 unit tests for the visibility pass.
- Full suites green on macOS aarch64 (incl. `--include-ignored`), Linux x86_64, and Linux aarch64; the cdylib end-to-end host run verified on all three targets.

### Docs

`docs/beyond-php/cdylib.md` (flag, `#[Export]`, v1 marshaling table, lifecycle contract, symbol visibility), updated example README, and the shipped v0.27.x roadmap items marked done.

Still follow-ups (unchanged from the original plan): string returns + `elephc_free` ownership, `.a` output, auto-generated headers, exception bridging, multi-threaded hosts.

PR #373, opened 2026-06-16
fix(core): balance heap ownership for borrowed passthrough and mixed-to-string casts

Guikingone:
Two related single-owner-with-copy ownership fixes in EIR lowering, each with its own regression coverage.

1. Borrowed by-value parameter passthrough (self-reassign use-after-free)

When a user function returns one of its own never-rebound by-value parameters (e.g. `function ident($s) { return $s; }`), the call result aliases the caller's argument storage. `$x = ident($x)` released the old `$x` and also freed the call result, double-freeing the same block; `$y = ident($x)` freed `$x`'s storage while `$x` was still live. A new AST pass (borrow_passthrough.rs) marks such call results `Ownership::Borrowed` (only for lifetime-tracked types, so ints stay NonHeap). `value_is_owning_temporary` returns false for Borrowed values (no spurious source release), and `store_local` retains/copies the new value before releasing the previous owner (no read-after-free).

2. Owned string copies from mixed-to-string casts

Coercing a `mixed`/union value to string -- concatenation, interpolation, or an explicit `(string)` cast -- lowers as `Op::Cast` with a `Str` target, whose codegen calls `__rt_mixed_cast_string`. For a boxed string payload that helper persists a fresh owned copy, but `value_is_owning_temporary` did not recognize `Op::Cast` results as owning, so the consuming concat or store never freed the copy. Each such cast leaked one heap block, scaling with cast count and accumulating across loops. `cast_result_owns_string_storage` now treats a `Cast -> Str` result as an owning temporary when its source is `mixed`/union. This is source-aware on purpose: string and array casts alias their source and int/float casts use scratch storage, so only mixed/union string casts own releasable heap storage; releasing a mixed-int/float cast is a safe no-op because `__rt_heap_free_safe` ignores out-of-heap pointers.

Both changes are in target-independent EIR lowering; the freeing paths are already dual-arch.

Tests: borrowed-passthrough self-reassign, cross-variable, and array regressions, plus mixed-to-string concat, explicit `(string)` cast, no-accumulation loop, and a mixed-int scratch-pointer release no-op guard, all under tests/codegen/runtime_gc. docs/internals/memory-model.md documents both owning-temporary rules.


nahime0:
Superseded by the post-#540 ownership work (ReturnAliasSummaries / value_is_borrowed_user_call_result, value_is_owning_mixed_string_cast, #544 and follow-ups). Original UAF and mixed-concat cast-leak fixtures pass on current main. Closing as obsolete rather than rebasing ~2k commits of conflicts. Residual return (string)$mixed leak is a separate return-path issue and should be a new focused PR if still wanted.

PR #394, opened 2026-06-19
feat: add image support — GD, Exif/IPTC, Imagick/Gmagick, Cairo

Guikingone:
Image support

  Adds full PHP image support to elephc, through a pure-Rust bridge crate (crates/elephc-image/, 17 files) and a pure-PHP prelude (src/image_prelude.rs) that is only injected
  when an image class is referenced. No system GD, libpng, or ImageMagick is required — image decoding/encoding uses the pure-Rust image crate and the rasterizers are
  reimplemented.

  Four areas are covered:

  - GD — image create/load/write, resize, crop, rotate, filters, color handling, and built-in (font8x8) text rendering.
  - Exif / IPTC — exif_read_data, iptcparse, iptcembed, and thumbnail extraction.
  - Imagick and Gmagick — an OOP subset is actually implemented (63 Imagick methods, 49 Gmagick), and every other method in the PHP surface is declared as a throwing stub.
  The whole API stays callable: exotic operations with no pure-Rust equivalent throw ImagickException / GmagickException("… not supported in elephc") at runtime, matching
  PHP's "method exists but unsupported" behavior. Implemented behavior is unchanged.
  - Cairo — a procedural plus OOP subset, deliberately partial. Undefined exotic methods are a compile error by contract, pinned by a #[should_panic] test.

  The "API-surface complete" stance — declaring throwing stubs for roughly 450 otherwise-absent methods — is the point of this PR. A program that calls e.g. swirlImage(),
  quantizeImage(), or identifyImage() now compiles and throws at runtime, instead of failing to compile with "Undefined method".

  Generator and snapshots

  scripts/gen_image_api_stubs.py reads the php.net class synopses (snapshots under scripts/image_synopsis/*.html) and the set of methods already implemented, and emits the
  stub declarations plus a per-stub coverage fixture. The generated output was hand-verified before commit.

  Tests, examples, docs

  - tests/codegen/image/ — imagick_api_surface.rs, gmagick_api_surface.rs, cairo, basics, draw, pixel. Every stub compiles when called and throws the matching exception.
  - tests/error_tests/image.rs — diagnostics for malformed calls.
  - Seven examples under examples/ (image_basics, image_cairo, image_imagick, …).
  - docs/php/image.md — reframed from "unsupported = absent" to "present, throws".

  Bundled general EIR fixes

  Three independent fixes surfaced during the image work and are not yet on origin/main, so they travel with this PR rather than being dropped:

  1. Chained-receiver temporary leak. $o->a()->b() and new X()->m() leaked the owning receiver temporary. Method-call sites now release it through
  release_owning_receiver_temporary; the receiver and the call arguments are disjoint value sets, so releasing both is safe.
  2. int/bool → float param widening. An int or bool argument to a float parameter was read as garbage on x86_64. It is now widened at the call boundary. This is
  x86-sensitive and was verified on Docker x86_64.
  3. DCE try/catch tail-sink blowup. Many sequential try/catch blocks in one function made the optimizer clone the tail into every fall-through path, compounding to 2ⁿ and
  overflowing the assembler at around 8 blocks. The tail is now kept as a single shared after-block when more than one path falls through, so growth is linear.

  Two small follow-on changes were needed because origin/main added AST variants since the fork: image_prelude/detect.rs gained arms for TypeExpr::Intersection and
  ExprKind::IncludeValue, so the exhaustive detection walker stays complete.


Guikingone:
Will fix the text in the description tomorrow, strange behavior from gh on my machine 😅

Guikingone:
@nahime0 Fixed, ready for review when you have time 🙂

PR #403, opened 2026-06-23
feat(core): zval pack/unpack bridge builtins + release owned call-argument temporaries

Guikingone:
## Summary

  This PR adds the **zval bridge** (ROADMAP v0.28.x) — four compiler-extension
  builtins that convert elephc runtime values to and from PHP-shaped `zval`
  structures — and bundles a general EIR ownership fix that the bridge benefits
  from.

  ## zval bridge (`eb759b83b`)

  Four builtins, EIR-only, for FFI / PHP-extension interop (no standard-PHP
  equivalent):

  | Builtin | Signature | Behavior |
  |---|---|---|
  | `zval_pack` | `(mixed): ptr` | Deep-copies an elephc value into a freshly allocated 16-byte PHP `zval` |
  | `zval_unpack` | `(ptr): mixed` | Rebuilds an elephc value from a `zval` |
  | `zval_type` | `(ptr): int` | Returns the PHP `IS_*` type byte at `zval+8` |
  | `zval_free` | `(ptr): void` | Releases the `zval` and the PHP-shaped storage it owns |

  Covered value kinds: `int` / `float` / `bool` / `null`, strings (copied into an
  owned `zend_string`), packed arrays (`nTableMask = HT_MIN_MASK`), hash arrays
  (DJBX33A hashing, `arHash` + Buckets), and nested arrays. Values are deep-copied
  into PHP-shaped heap objects so a linked PHP extension can read them directly;
  `zval_free` recurses to free `zend_string` keys and child values.

  All four are wired through the standard builtin path (catalog, signatures,
  case-insensitive lookup, namespace fallback, first-class-callable support) and
  the EIR backend; the frozen legacy AST backend is untouched.

  ## Release owned call-argument temporaries (`af3bd3207`)

  Plain builtin / user / extern calls never released owned **argument**
  temporaries the way method calls already did, so a fresh owned temp passed as an
  argument leaked one heap value per call — **unbounded in a loop** (e.g.
  `array_sum([$i, 20, 30])`, `count([$i, …])`, `implode(",", [$i, …])`).

  `lower_function_call` now releases owned argument temporaries after the call,
  and the lazy `isset` fallback releases its re-derived operand. The alias guard
  `call_result_may_alias_arg` is extended to treat container/iterable pairings in
  either direction as a possible alias, so a function returning its array argument
  typed as `iterable` (`function id(iterable $x): iterable`) does not have that
  still-referenced payload freed. As a side benefit this also makes
  `zval_pack([…])` leak-clean, since zval builtins route through the same path.

  ## Testing

  - **3-target** across `macos-aarch64`, `linux-aarch64`, `linux-x86_64`.
  - zval: 65 codegen tests (scalar/string/null round-trips, packed + hash + nested
    array content round-trips, `zval_type`, `zval_free` leak regressions, arg-count
    errors, embedded-NUL strings, `PHP_INT_MAX`/`MIN`, dynamic-callback link guard).
  - Call-arg fix: full macOS suite green (0 double-frees) + `runtime_gc` green on
    both Linux targets; added an iterable-passthrough correctness regression test.
  - Full macOS suite green on the combined branch.


Guikingone:
@nahime0 Not sure if this was what you're looking for when talking about `zval` support, let me know if not, can dive into it to improve it.

nahime0:
@Guikingone wihout looking at the implementation, and based only on the description, yes, that's exactly what we need to be able to produce a real php extension. Thanks!

I'll take a closer look at the code asap.

nahime0:
@Guikingone I've updated this branch to main and added a plan of what I'm hoping to achieve:

a) consume php libraries
b) produce php libraries 

I'd like to focus con (b). 

I think I will merge this after #496, given that has no side effects. We can continue implementing later on.

Issue #408, opened 2026-06-24
Releasing a string-keyed associative array leaks its keys (per-request heap exhaustion under --web)


nahime0:
## Summary

Releasing a **string-keyed associative array** does not free its key strings (and/or entry cells): each release leaks heap proportional to the number of string keys. Integer-keyed arrays (lists) release correctly. This is a steady per-release leak that exhausts the fixed heap.

It is especially severe under `--web`: the request superglobals (`$_SERVER`, `$_GET`, `$_POST`, `$_COOKIE`, `$_REQUEST`, `$_ENV`, `$_FILES`) are string-keyed assoc arrays rebuilt and released **every request**, so every worker leaks a fixed amount per request and the heap is exhausted after a few hundred requests — even for a trivial `<?php echo "hi";` handler. Because `$_ENV` copies the whole process environment, a large environment makes workers die proportionally sooner.

## Verified on

- `elephc 0.25.0`, 2026-06-24, target `macos-aarch64` (default EIR backend)

## Minimal reproduction (no `--web` needed)

```php
<?php
$g = [];
for ($n = 0; $n < 200000; $n++) {
    $g = [];          // releases the previous string-keyed array
    $g["a"] = "x";    // string key
}
echo "done";
```

## Expected

Prints `done` (the previous array is fully freed each reassignment, so the heap stays flat — exactly like the integer-key version below, which works).

## Actual

```
Fatal error: heap memory exhausted
```

## Isolation

These do **not** leak, which pins the leak to releasing a *string-keyed* array:

```php
// integer key -> no leak
$g = [];
for ($n = 0; $n < 200000; $n++) { $g = []; $g[0] = "x"; }
echo "done";              // done
```
```php
// no reassignment (accumulate into the same array) -> no leak
$g = [];
for ($n = 0; $n < 200000; $n++) { $g["a"] = "x"; }
echo "done";              // done
```

Computed keys/values leak the same way as literal ones, so it is the key *strings* (or the hash entry cells), not the values, that are not freed when the assoc array is released.

## `--web` impact (measured)

A trivial `<?php echo "hi";` server on one worker dies after ~860 requests (8 MiB default heap), ~137 requests with a ~50 KB env var present, and ~6930 requests with `--heap-size=67108864` — a clean linear per-request leak. Workarounds until this is fixed: `--max-requests N` (recycle each worker before it exhausts) and/or a larger `--heap-size`.

nahime0:
[...]

Issue #407, opened 2026-06-24
Variable string-key access on a typed array property is rejected ("Array index must be integer")


nahime0:
## Summary

Reading a typed `array` property by a **variable** string key (`$this->prop[$key]`) is rejected at compile time with `Array index must be integer`. Literal string keys on properties work (fixed in #250); variable string keys on **local** arrays and on **free-function parameters** also work. Only the *property + variable-key* combination fails.

## Verified on

- `elephc 0.25.0`, 2026-06-24, target `macos-aarch64` (default EIR backend)

## Minimal reproduction

```php
<?php
class Bag {
    public array $data = ['a' => 1];
    public function get(string $key): int {
        return $this->data[$key] ?? 0;
    }
}
$b = new Bag();
echo $b->get('a');
```

## Expected

Prints `1`.

## Actual

Compilation fails:

```
error: Array index must be integer
```

## Notes

These all work, isolating the failure to *property + variable key*:

- Literal key on a property: `$this->data['a']` — works (see #250).
- Variable key on a **local** array: `$m = ['a' => 1]; $k = 'a'; echo $m[$k];` — works.
- Variable key on a **free-function parameter**: `function f(array $a, string $k): int { return $a[$k] ?? 0; }` — works.

This looks like a remaining facet of #250 (typed array properties): the associative element type is not used when the key is a variable rather than a literal.

Issue #406, opened 2026-06-24
array-typed method parameters lose associative shape (string keys rejected; free-function params work)


nahime0:
## Summary

An `array`-typed **parameter of a method** (instance or static) is treated as an integer-indexed list: string-key access is rejected at compile time with `Array index must be integer`, and `json_encode()` of the parameter emits a JSON list with wrong values instead of a JSON object. The identical `array` parameter on a **free function** preserves the associative shape and works.

## Verified on

- `elephc 0.25.0`, 2026-06-24, target `macos-aarch64` (default EIR backend)

## Minimal reproduction

```php
<?php
class W {
    public static function first(array $d): string {
        return $d['a'];          // <- rejected
    }
}
echo W::first(['a' => 'hello']);
```

## Expected

Prints `hello`.

## Actual

Compilation fails:

```
error: Array index must be integer
```

When the string-key access is replaced with `json_encode($d)`, the method instead returns a JSON **list with garbage values** (e.g. `['a' => 1, 'b' => 2]` encodes to `[12,5]`) rather than `{"a":1,"b":2}`.

## Notes

- A **free function** with the same signature works correctly:
  ```php
  function first(array $d): string { return $d['a']; }
  echo first(['a' => 'hello']);          // hello
  ```
  ```php
  function enc(array $d): string { return json_encode($d); }
  echo enc(['a' => 1, 'b' => 2]);        // {"a":1,"b":2}
  ```
- Both **instance** and **static** methods are affected.
- The associative shape of the argument is known at the call site, so the method parameter should be typed/lowered the same way the free-function parameter is.

Issue #405, opened 2026-06-24
Appending a foreach value into a new array and returning it corrupts the array (dangling elements / heap exhaustion)


nahime0:
## Summary

Building an array by appending the **value variable of a `foreach` loop** and then returning that array from a function produces a corrupted array: indexing the result yields empty strings, and iterating it exhausts the heap. Appending a function *parameter* (instead of a `foreach` value), or returning `explode(...)` directly, both work — which points at the appended `foreach` value being stored as a borrow into the (soon-freed) iterated array rather than an owned copy.

## Verified on

- `elephc 0.25.0`, 2026-06-24, target `macos-aarch64` (default EIR backend)

## Minimal reproduction

```php
<?php
function collect(string $csv): array {
    $out = [];
    foreach (explode(',', $csv) as $item) {
        $out[] = $item;          // append the foreach value
    }
    return $out;
}
$r = collect('a,b,c');
echo $r[0];                       // expected: a
```

## Expected

Prints `a`. Iterating `$r` (`foreach ($r as $e) echo $e;`) prints `abc`.

## Actual

`echo $r[0];` prints nothing. Reading the returned array back with `foreach` aborts at runtime:

```
Fatal error: heap memory exhausted
```

## Notes

These variants all work, isolating the cause to the appended `foreach` value:

- `return explode(',', $csv);` (return the split directly) → works.
- `function mk(string $a, string $b): array { $o = []; $o[] = $a; $o[] = $b; return $o; }` (append parameters) → works.

So `$out[] = $item;` appears to store a borrow of the `foreach` element instead of an owned copy; when the iterated `explode()` temporary is freed at function return, the stored elements dangle.

mirchaemanuel:
I'm working on this one — status so far.

**Re-confirmed on current `main` `84095e0ee`** (local debug build, macOS ARM64), and the symptom has gotten worse than the original report: the minimal repro now dies with `Fatal error: heap memory exhausted` already at `echo $r[0];`, without any foreach over the result.

**Diagnosis so far** (from EIR dumps + an lldb watchpoint session):

[...]

PR #409, opened 2026-06-24
feat(web): compile a PHP program to a standalone prefork HTTP server (--web)

nahime0:
## Summary

Adds `--web`: a compiler mode that turns a standard PHP file into a **standalone, prefork HTTP server binary** instead of a CLI executable. The PHP you compile is ordinary PHP — the same file runs under `php-fpm` or `php -S` — but the compile-and-serve mechanism is elephc-specific. The produced binary has no PHP runtime dependency.

```bash
elephc --web app.php
./app --listen 127.0.0.1:8080 --workers 4
```

## Architecture

- A small Rust bridge crate (`crates/elephc-web`, hyper + tokio current-thread) is linked into `--web` binaries. The process entry forks **N `SO_REUSEPORT` workers**; the kernel load-balances connections across them.
- Each request re-runs the program's top-level code from clean state (PHP-FPM model). `echo`/`print` output is captured into the response body; status and headers are driven by `http_response_code()` / `header()`.
- Per-request state reset (`__rt_web_reset`) clears globals, function statics, static properties, and the request superglobals between requests so no data leaks across requests.

## Request & response surface

- **Superglobals**, rebuilt fresh per request: `$_SERVER` (method/URI/query, `HTTP_*` headers, `REMOTE_*`/`SERVER_*`, `REQUEST_TIME`, …), `$_GET`, `$_POST` (urlencoded **and** multipart text fields), `$_COOKIE`, `$_REQUEST`, `$_ENV`, `$_FILES` (multipart uploads with `tmp_name`).
- `file_get_contents('php://input')` returns the raw request body.
- **Response control**: `http_response_code()`, `header()` (full PHP semantics incl. `Location`/`Status`), `setcookie()` / `setrawcookie()`. Output is buffered, so setting headers after output is fine (no "headers already sent").

## Server operations

Runtime flags on the produced binary: `--listen`, `--workers`, `--max-body-size` (413 on overflow), `--max-requests` (worker recycling), `--max-execution-time` (SIGALRM watchdog kills runaway handlers), `--access-log`, `--gzip`, `--help`/`--version`.

Robustness: graceful SIGINT/SIGTERM shutdown, worker respawn + crash-loop guard, clean bind-error UX, HTTP/1.1 keep-alive with a header-read timeout, and per-connection `spawn_local` tasks so a slow/idle keep-alive connection no longer blocks the accept loop.

## Notable shared-infrastructure fix

`fix(linker): co-link multiple whole-archive Rust bridges without duplicate std symbols` — a `--web` program that also pulls in a second whole-archive bridge (e.g. `elephc-tls`, via `file_get_contents`/`move_uploaded_file` on a dynamic path) previously failed to link because each bridge bundles its own `std`/`core`. Resolved generally for any number of whole-archive bridges (Linux: `--allow-multiple-definition`; macOS: symbol/name-based member dedup).

## Examples & docs

- Examples: `web-hello`, `web-request`, `web-response`, `web-router`, `web-demo`, and `web-framework` (a tiny Laravel-style framework: namespaced `Request`/`Response`/`Router`, single-action controllers, a middleware onion, `:param` routes, JSON).
- Docs: `docs/beyond-php/web.md` (full guide), `docs/compiling/cli-reference.md`, internals notes; `ROADMAP.md` / `CHANGELOG.md` updated.

## Testing

- `tests/web_tests.rs`: 39 end-to-end tests (compile → serve → assert), covering superglobals, cookies, `$_REQUEST`/`$_ENV`, multipart `$_FILES` (incl. reading an upload back), `--max-requests`/`--max-execution-time`, `--gzip`, body-size 413, keep-alive, graceful shutdown, worker respawn, uncaught-exception → 500, and namespaced programs.
- Full suite green on `macos-aarch64`: `web_tests` 39/39, `codegen_tests` 4450/0, bridge unit tests passing. Supported targets: `macos-aarch64`, `linux-aarch64`, `linux-x86_64`.

## Known limitation (tracked separately)

- **Per-request memory leak under sustained load** — releasing a string-keyed associative array does not free its key strings ([#408](https://github.com/illegalstudio/elephc/issues/408)). Because the superglobals are string-keyed assocs rebuilt every request, a worker's fixed heap is exhausted after a few hundred requests. This is a pre-existing runtime bug surfaced by `--web`, not introduced here. Mitigations until it lands: `--max-requests N` (recycle workers) and/or a larger `--heap-size`.
- Building the `web-framework` example also surfaced three type-checker bugs, tracked at [#405](https://github.com/illegalstudio/elephc/issues/405), [#406](https://github.com/illegalstudio/elephc/issues/406), [#407](https://github.com/illegalstudio/elephc/issues/407).
- Not yet supported: graceful in-flight drain on shutdown, response streaming, Unix-socket `--listen`, sessions, in-process TLS/HTTP2-3 (front with a reverse proxy).


Guikingone:
> Each request re-runs the program's top-level code from clean state (PHP-FPM model). echo/print output is captured into the response body; status and headers are driven by http_response_code() / header().

Am I wrong when thinking about bringing a "worker" mode like FrankenPHP? 🤔 👀

nahime0:
> > Each request re-runs the program's top-level code from clean state (PHP-FPM model). echo/print output is captured into the response body; status and headers are driven by http_response_code() / header().
> 
> Am I wrong when thinking about bringing a "worker" mode like FrankenPHP? 🤔 👀

You're definitely right. Right now I'm focused on stabilizing this shared-nothing implementation (a few memory leaks here and there) that isolates each request. Worker mode is a very valid follow-up, but it needs a rethink of how, and what, to isolate between requests.

PR #417, opened 2026-06-27
feat(core): remove --ast-backend CLI flag and legacy pipeline entry point

Guikingone:
Phase 1 of legacy backend removal. The EIR backend is now the only codegen path reachable from the CLI:

- Remove CodegenBackend::Ast variant and --ast-backend flag parsing
- Remove the legacy branch in pipeline.rs (always lower through EIR)
- Migrate tests/codegen/support/projects.rs from codegen::generate to compile_via_eir with prelude injection and autoload setup
- Remove TestCodegenBackend::Legacy from compiler.rs (always EIR)
- Delete tests/ir_backend_parity.rs (compared legacy vs EIR, no longer relevant)
- Mark 2 SPL filesystem iterator tests as #[ignore] (they only passed via the legacy backend, EIR crash to be fixed separately)
- Update README.md, cli-reference.md, optimization.md to reflect the removal
- Mark CheckResult fields as #[allow(dead_code)] (were only read by legacy backend; will be removed in Phase 2)

The frozen legacy emitter modules under src/codegen/expr/, src/codegen/ stmt/, etc. remain in-tree but are no longer reachable. Phase 2 will delete them after disentangling shared ABI/runtime helpers.


nahime0:
Closed. Implemented in #481

PR #428, opened 2026-06-29
fix(core): release owning call-result/argument temporaries when a call throws

### LINKED ISSUE #399: Owning call-result/argument temporary leaks when the call throws
## Summary

When a method/function call throws, an owning *temporary* produced earlier in the same expression (e.g. the intermediate result of a chained call, or an inline `new X()` receiver/argument) is not released during exception unwinding, so its `__destruct` never runs and the object leaks.

## Reproduction

```php
<?php
class N {
    public function __destruct() { echo "dtor\n"; }
    public function make(): N { return new N(); }
    public function boom(): void { throw new Exception("x"); }
}
function run(): void {
    $t = new N();
    try { $t->make()->boom(); } catch (Exception $e) {}
    echo "caught\n";
}
run();
echo "end\n";
```

## Expected (PHP) — 2 destructors

```
dtor      # the make() intermediate, destroyed during unwind
caught
dtor      # $t at end of run()
end
```

## Actual (elephc) — 1 destructor (intermediate leaks)

```
caught
dtor      # only $t
end
```

The object returned by `make()` is never destructed.

## Root cause

The releases for owning receiver/argument temporaries (`release_owning_receiver_temporary` / `release_owned_call_arg_temporaries` in `src/ir_lower/expr/`) are emitted in straight-line order *after* the call. If the call throws, unwinding bypasses them, so the temporary is leaked. The normal (non-throwing) path balances correctly.

## Scope / notes

- Pre-existing; identified during the EIR-fix review bundled into the image-support work. Not a regression.
- Same gap affects an owning argument temporary when the callee throws (`$t->take(new N())` where `take` throws).
- Fixing it needs the temporary releases to be part of the exception-cleanup path (landing pads), not just the straight-line fall-through.

Guikingone:
When a chained method call like $t->make()->boom() throws, the intermediate owning temporary from make() was leaked because longjmp bypassed the straight-line release code. This adds a runtime pending- cleanup stack (TLS global array) that records owning temporaries before a potentially-throwing call. If the call throws, __rt_throw_current drains the stack via __rt_eh_drain (calling __rt_decref_any on each entry) before longjmp. On the normal path, eh_pop removes the entry after the call returns.

New EIR ops: EhPush (1 operand) and EhPop (0 operands). New runtime helpers: __rt_eh_push, __rt_eh_pop, __rt_eh_drain (AArch64 + x86_64).
New globals: _eh_cleanup_stack (2048 bytes), _eh_cleanup_top (8 bytes).

Fixes #399

### DIFF
--- a/src/codegen/runtime/data/fixed.rs
+++ b/src/codegen/runtime/data/fixed.rs
@@ -86,6 +86,12 @@ pub(crate) fn emit_runtime_data_fixed(heap_size: usize, target: Target) -> Strin
     out.push_str(".comm _exc_handler_top, 8, 3\n");
     out.push_str(".comm _exc_call_frame_top, 8, 3\n");
     out.push_str(".comm _exc_value, 8, 3\n");
+    // Exception cleanup stack: array of owning-temporary pointers that must be
+    // released if a call throws. __rt_eh_push stores a pointer before a
+    // potentially-throwing call; __rt_eh_pop removes it after the call returns
+    // normally; __rt_throw_current drains the stack before longjmp.
+    out.push_str(".comm _eh_cleanup_stack, 2048, 3\n");
+    out.push_str(".comm _eh_cleanup_top, 8, 3\n");
     out.push_str(".comm _fiber_current, 8, 3\n");
     out.push_str(".comm _fiber_main_saved_sp, 8, 3\n");
     out.push_str(".comm _fiber_main_saved_exc, 8, 3\n");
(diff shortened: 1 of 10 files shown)


nahime0:
Closing this draft. **Do not close `#399`.**

The leak is still real on `main`. This patch is not the fix: the cleanup stack is process-global (not TLS), would not survive fiber switch, already failed Linux PIE link, and the codegen paths it edits no longer exist.

Please do not rebase `e61819bb`. Salvage is the three destructor tests and the problem statement. A real fix belongs on current `main`, registered into the existing activation / fiber story, not a second global stack. `#775` stays separate.

PR #432, opened 2026-06-29
fix(core): numeric string arithmetic coercion

### LINKED ISSUE #362: Numeric string arithmetic is rejected instead of following PHP coercion
## Description

PHP accepts arithmetic on numeric-prefix strings, emits a warning for trailing junk, and still performs the coercion.

The current Elephc tree rejects the program at compile time with `Arithmetic operators require numeric operands`, which is not PHP-compatible behavior for this case.

## Minimal reproduction

```php
<?php
var_dump("  +12foo" + 3);
```

## Steps to reproduce

1. Save the program above as `main.php`.
2. Run `php main.php`.
3. Compile it with Elephc: `target/debug/elephc main.php`.

## Expected behavior

PHP 8.4.19 emits a warning and then prints:

```text
Warning: A non-numeric value encountered in /path/to/main.php on line 2
int(15)
```

## Actual behavior

Elephc fails during compilation with:

```text
error[2:21]: Arithmetic operators require numeric operands
```

## Environment

- Elephc crate version: `0.23.8`
- Git commit: `ca334f141b930f8075d36e5074e49934aeb4efb4`
- PHP baseline: `PHP 8.4.19 (cli)`
- OS: `Darwin 25.5.0 arm64`
- Reproduced on: `2026-06-11T07:01:17Z`

Guikingone:
Allow Str operands in arithmetic (+, -, *, /, %, **) by treating them as numeric at runtime, matching PHP semantics. Pure integer-form strings coerce to int; float-form strings (containing a dot or exponent) coerce to float. Leading-numeric strings emit a compile-time warning and use the numeric prefix.

Type checker: accept Str in is_numeric_operand_type, produce Mixed result type when either operand is Str, and warn for constant non-numeric strings. Lowering: coerce Str to float for Div via StrToF, route Add/Sub/Mul through MixedNumericBinop with Str operands.

Runtime: new __rt_str_is_float_form helper (ARM64 + x86_64) parses the string with strtoll + strtod and reports whether strtod consumed more bytes (float-form). MixedNumericBinop now checks string operands for float-form on both targets and routes to the float path when needed, fixing "1.5" + 3 producing int(4) instead of float(4.5).

9 regression tests added. Docs and arithmetic example updated.

Fix #362

### DIFF
--- a/docs/php/operators.md
+++ b/docs/php/operators.md
@@ -17,6 +17,13 @@ sidebar:
 | `**` | `$a ** $b` | Exponentiation (right-associative, returns float) |
 | `-$x` | `-$x` | Unary negation |
 
+Numeric strings are accepted as arithmetic operands and coerced at runtime,
+matching PHP semantics. Pure integer-form strings (e.g. `"123"`) coerce to
+`int`; float-form strings (e.g. `"1.5"`) coerce to `float`. Leading-numeric
+strings (e.g. `"12abc"`) emit a compile-time `Warning: A non-numeric value
+encountered` and use the numeric prefix. The result type is `Mixed` when either
+operand is a string, since the runtime type depends on the string's content.
+
 ## Comparison
 
 | Operator | Example | Notes |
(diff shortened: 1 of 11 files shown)


nahime0:
Thanks for this — the float-form routing insight here is the core of #944, which reimplemented it on current main without widening the shared compare/spaceship operand check and clipping through __rt_php_num_scan instead of raw strtoll/strtod; closing as superseded now that #944 has landed.

Issue #382, opened 2026-06-17
Recursive closures using `use (&$self)` are rejected as undefined


nahime0:
## Summary
Recursive closures that capture themselves by reference with `use (&$f)` are rejected as undefined variables, even though PHP allows this pattern and it is a standard way to express self-recursive anonymous functions.

## Minimal repro
```php
<?php
$f = function (int $n) use (&$f): int {
    return $n <= 1 ? 1 : $n * $f($n - 1);
};
echo $f(5);
```

## Steps to reproduce
1. Save as `main.php`.
2. Run `php main.php`.
3. Run `elephc main.php`.

## Expected behavior
PHP 8.4.20 prints:
```text
120
```

## Actual behavior
Elephc rejects the program during compilation:
```text
error[2:6]: Undefined variable in use(): $f
error[3:6]: Undefined variable: $f
```

## Backend note
This reproduces on both the default EIR backend and `--ast-backend` on current `main`.

## Environment
- elephc commit: `9a65c244f8b5b553f483fb326f9e8eace26611cb`
- elephc version: `0.24.1` (`Cargo.toml`)
- OS: `Darwin 25.5.0 arm64`
- PHP reference: `PHP 8.4.20 (cli)`
- Reproduced on: `2026-06-17`

nahime0:
Closed by #435 (merged).

Issue #381, opened 2026-06-17
Default EIR foreach over arrays visits appended elements during value iteration


nahime0:
## Summary
With the default EIR backend, `foreach` by value over an array incorrectly visits elements appended during the loop. PHP does not iterate the appended element in this case.

## Minimal repro
```php
<?php
$a = [1, 2];
foreach ($a as $v) {
    echo $v;
    if ($v === 1) $a[] = 3;
}
echo '|' . count($a);
```

## Steps to reproduce
1. Save as `main.php`.
2. Run `php main.php`.
3. Run `elephc main.php && ./main`.

## Expected behavior
PHP 8.4.20 prints:
```text
12|3
```

## Actual behavior
The elephc-compiled binary prints:
```text
123|3
```

## Backend note
This reproduces on the default EIR backend on current `main`. The legacy backend matches PHP for the same testcase.

## Environment
- elephc commit: `9a65c244f8b5b553f483fb326f9e8eace26611cb`
- elephc version: `0.24.1` (`Cargo.toml`)
- OS: `Darwin 25.5.0 arm64`
- PHP reference: `PHP 8.4.20 (cli)`
- Reproduced on: `2026-06-17`

nahime0:
Closed by #435 (merged).

Issue #359, opened 2026-06-11
Static closures wrongly reject `isset($this)`


nahime0:
## Description

PHP allows `isset($this)` inside a `static` closure created in an instance method. The expression evaluates to `false` because the static closure is not bound to the object.

The current Elephc tree rejects this at compile time with `Cannot use $this inside a static closure`, which is stricter than PHP and breaks compatible code.

## Minimal reproduction

```php
<?php
class C {
    function go() {
        $f = static function() {
            echo isset($this) ? "yes" : "no";
        };
        $f();
    }
}
(new C())->go();
```

## Steps to reproduce

1. Save the program above as `main.php`.
2. Run `php main.php`.
3. Compile it with Elephc: `target/debug/elephc main.php`.

## Expected behavior

PHP 8.4.19 prints:

```text
no
```

Elephc should compile the program and the generated binary should print the same output.

## Actual behavior

Elephc fails during compilation with:

```text
error[4:25]: Cannot use $this inside a static closure
error[6:9]: Undefined variable: $f
```

## Environment

- Elephc crate version: `0.23.8`
- Git commit: `ca334f141b930f8075d36e5074e49934aeb4efb4`
- PHP baseline: `PHP 8.4.19 (cli)`
- OS: `Darwin 25.5.0 arm64`
- Reproduced on: `2026-06-11T07:01:17Z`

nahime0:
Closed by #435 (merged).

Issue #339, opened 2026-06-05
Typed property uninitialized access is not catchable as `Error`


nahime0:
## Description

Accessing an uninitialized typed property inside a `try` block should throw a catchable `Error`. The current elephc binary instead terminates with a fatal error, so the `catch (Error $e)` block never runs.

## Minimal reproduction

```php
<?php
class Box { public int $v; }
$b = new Box();
try { var_dump($b->v); } catch (Error $e) { echo $e->getMessage(); }
```

## Steps to reproduce

1. Save as `main.php`.
2. Run `php main.php`.
3. Compile and run with elephc.

## Expected behavior

PHP 8.4 prints:

```text
Typed property Box::$v must not be accessed before initialization
```

and exits successfully because the `Error` is caught.

## Actual behavior

The elephc-compiled binary prints a fatal error to stderr:

```text
Fatal error: Typed property Box::$v must not be accessed before initialization
```

and exits with status `1`.

## Environment

- elephc 0.23.5
- PHP 8.4.19
- macOS arm64
- reproduced on 2026-06-05

## Impact

This changes exception/fatal control flow and makes a standard PHP `catch (Error $e)` path unreachable.

nahime0:
Confirmed again on current `main` (`9a65c244f8b5b553f483fb326f9e8eace26611cb`) during the 2026-06-17 stress run, on macOS arm64 with PHP 8.4.20 and elephc 0.24.1.

Fresh repro:

```php
<?php
class Box { public int $x; }
try {
    $b = new Box();
    var_dump($b->x);
} catch (Error $e) {
    echo 'err';
}
```

Observed today:

- PHP 8.4.20: `err`
- elephc current main: stderr `Fatal error: Typed property Box::$x must not be accessed before initialization`, exit status `1`

So uninitialized typed property access is still not catchable as `Error` on the current tree.

nahime0:
Closed by #435 (merged).

PR #435, opened 2026-06-29
fix(quick-wins): static closure isset($this), recursive closures, foreach snapshot, typed property Error

Guikingone:
#359: static closures wrongly reject isset($this). The checker now exempts bare $this arguments to isset() from the 'Cannot use $this inside a static closure' error. The EIR lowering emits a constant false when isset($this) targets a missing 'this' local slot. Non-isset $this usage remains rejected.

#382: recursive closures using use(&$f) were rejected as undefined. The checker now pre-declares the assignment target as Callable before inferring the closure type when a self-referential by-ref capture is detected, matching PHP's behavior of pre-initializing the variable to null.

#381: foreach by value over indexed arrays visited elements appended during iteration. The iterator state now snapshots the array length at IterStart and compares against the snapshot in IterNext instead of re-reading the live length each iteration. Both static indexed and dynamic iterable/mixed paths snapshot the length.

#339: typed property uninitialized access was a fatal exit instead of a catchable Error. The guard now constructs an Error object, publishes it to _exc_value, and branches to __rt_throw_current so surrounding try/catch blocks can observe and catch it. Both instance and static property guards are updated. PropertyAccess and StaticPropertyAccess effects are now marked may_throw so the optimizer preserves try/catch wrappers around property reads.


Guikingone:
@nahime0 I'm pushing PRs with fixes for issues by batches, I'll let you handle the issue state once it's merged and if it fully fix the issue, I don't want to interfere with your tracking of bugs / lacking features.

To ease the process, I'm not moving the PR from Draft until the CI is working 🙂

nahime0:
Great, thanks @Guikingone!

I'm working on the oldest PR first. As soon as a PR is ready, I'll review it, make any changes if needed, and merge it.

Issue #425, opened 2026-06-28
String-key read on a foreach-rebuilt Array(Mixed) result is rejected ("Array index must be integer")


nahime0:
### Summary

After a `foreach ($src as $k => $v) $dst[$k] = $v;` rebuild, the destination is typed `Array(Mixed)`. At runtime its storage may be promoted to a hash (when string/sparse/negative keys are written), but because the **static** type stays `Array(Mixed)` (an indexed-array type), a direct **string-key read** on the result is rejected at compile time.

### Reproduction

```php
<?php
function rebuild(array $src): array {
    $dst = [];
    foreach ($src as $k => $v) { $dst[$k] = $v; }
    return $dst;
}
$r = rebuild(["name" => "Alice", "age" => 30]);
echo $r["name"];   // <-- compile error
```

elephc:

```
error: Array index must be integer
```

PHP: `Alice`

Iterating the result (`foreach ($r as $k => $v)`) and integer-index reads (`$r[0]`) work; only **string-key reads** fail. Array comparisons on the result also fail — that is the separate, general gap in #424.

### Root cause

- `src/types/checker/inference/expr/mod.rs` — the `PhpType::Array(elem_ty)` read arm rejects any non-integer index, including `Array(Mixed)`.
- The EIR read lowering (`src/ir_lower/expr/mod.rs`) coerces the index to int for `Heap(Array)` receivers; there is no runtime read helper that dispatches on heap kind (indexed vs promoted-hash) + key type for a raw `Array(Mixed)` pointer.

### Why not part of the Mixed-key write fix

A correct fix needs both:

1. relaxing the checker to allow non-int reads on `Array(Mixed)` — a global strictness change affecting every `Array(Mixed)` read (e.g. plain mixed-value lists `[1, "a", 3.0]`), returning `null` for missing keys like PHP; and
2. a new dual-arch runtime read helper (e.g. `__rt_array_get_mixed_key`) mirroring the existing write-side `__rt_array_set_mixed_key`, dispatching on `__rt_heap_kind` and the Mixed key tag.

That is a feature addition beyond the write-path bug fix.

### Context

Found while reviewing the foreach Mixed-key array-write path (`__rt_array_set_mixed_key`). The write side (string/sparse/negative keys) is correct and read-back via `foreach` works; this issue tracks enabling direct string-key reads of the rebuilt result.

Issue #398, opened 2026-06-22
Scalar-union builtin returns collapse false to ""/0 (e.g. image_type_to_extension)


nahime0:
## Summary

A builtin declared as `string|false` (and similar scalar-union returns) collapses its `false` failure value to the scalar type's empty value at runtime, so callers cannot distinguish failure with `=== false`.

## Reproduction

```php
<?php
$r = image_type_to_extension(IMAGETYPE_UNKNOWN); // PHP: false
var_dump($r === false);
var_dump($r);
```

## Expected (PHP)

```
bool(true)
bool(false)
```

## Actual (elephc)

```
bool(false)
string(0) ""
```

The `false` return is materialized as the empty string, so `$r === false` is wrong and `$r` is `""`.

## Scope

- Affects `string|false` (and likely `int|false` / other scalar-union) builtin returns. `array|false` works (`getimagesize()` on a missing file correctly returns `false`), so the gap is specific to scalar unions.
- Image functions touched: `image_type_to_extension`, and the `exif_*` family per the image docs. The same scalar-union machinery is general, so non-image builtins are likely affected too.
- Already noted as a known limitation in `tests/codegen/image/foundation.rs` and `docs/php/types.md`.

## Notes

Needs a runtime representation that can carry `false` distinctly from a scalar's empty value for scalar-union returns (the "scalar-union value-runtime" work referenced in the image tests).

nahime0:
Draft #436 is closed. Do not rebase that SHA.

The bug here is still real on current `main`: hint-less image prelude functions still collapse `false` to `""` / `0`, so `image_type_to_extension(IMAGETYPE_UNKNOWN) === false` is false. `#425` (string-key read on `Array(Mixed)`) already landed separately and is not this issue.

The #436 patch is the wrong shape now. `PhpType::False` and `string|false` exist (`getenv` in #801). Do not teach `wider_type` to send Bool+scalar to Mixed; that would fight declared unions. A fix is a new focused PR on current `main`: declare `: string|false` on `image_type_to_extension`, `exif_tagname`, `exif_thumbnail` (same pattern as #801), and flip the tests that currently pin `""`.

PR #436, opened 2026-06-30
fix(core): scalar-union returns preserve false, Array(Mixed) string-key reads

Guikingone:
#398: wider_type let Str/Float absorb Bool, collapsing string|false and int|false return types to Str/Int and coercing false to ""/0. Now Bool+ scalar and Mixed+anything widen to Mixed (boxed tagged cell), so false preserves its bool tag. image_type_to_extension(IMAGETYPE_UNKNOWN) now returns false instead of "".

#425: after foreach rebuilds an array as Array(Mixed), string-key reads were rejected at compile time ('Array index must be integer'). The checker now allows any PHP array key on Array(Mixed). A new EIR op ArrayGetMixedKey dispatches on heap kind at runtime (indexed vs hash) via __rt_array_get_mixed_key, returning an owned boxed Mixed cell.

### DIFF
--- a/src/codegen/runtime/arrays/array_get_mixed_key.rs
+++ b/src/codegen/runtime/arrays/array_get_mixed_key.rs
@@ -1,0 +1,288 @@
+//! Purpose:
+//! Emits the `__rt_array_get_mixed_key` runtime helper for reads from a
+//! statically `Array(Mixed)` indexed local whose key is a boxed `Mixed` cell
+//! (notably PHP `foreach` loop keys, which are always `Mixed` in EIR).
+//!
+//! Called from:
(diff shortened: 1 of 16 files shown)


nahime0:
Closing this draft.

`#425` already landed (PR `#434`, later closed by `#678`) with a stricter `ArrayGetMixedKey` on current `main`. `#398` is still a live bug, but this patch is the wrong shape now: `PhpType::False` and `string|false` exist, and widening Bool+scalar to Mixed would fight that.

Please do not rebase `06bbb3a5`. A `#398` fix should be a new focused PR on current `main`, following the `#801` `getenv` pattern (`Union(Str, False)` on the image prelude), not this SHA.

Issue #424, opened 2026-06-28
Array/hash equality (== and ===) is unsupported by the EIR backend


nahime0:
### Summary

Comparing two arrays with `==` or `===` is not supported by the EIR backend — for **any** array type (indexed or associative). This is a general gap, not tied to a specific array shape.

### Reproduction

```php
<?php
$a = [1, 2, 3];
var_dump($a === [1, 2, 3]);             // unsupported
var_dump(["x" => 1] === ["x" => 1]);   // unsupported
var_dump([1, 2, 3] == [1, 2, 3]);      // unsupported
```

elephc:

```
EIR backend error: unsupported EIR backend feature: strict_eq for PHP type Array(Int)
EIR backend error: unsupported EIR backend feature: strict_eq for PHP type AssocArray { key: Str, value: Int }
EIR backend error: unsupported EIR backend feature: loose_eq for PHP types Array(Int) and Array(Int)
```

PHP: `true` / `true` / `true`

### Root cause

- `src/codegen_ir/lower_inst/comparisons.rs` — the supported type combinations for `strict_eq` / `loose_eq` do not include `Array(..)` or `AssocArray`, so they fall into the "unsupported EIR backend feature" branch.
- No runtime equality helpers exist (no `__rt_array_eq` / `__rt_hash_eq`).

### Scope / semantics

PHP array comparison is non-trivial and recursive:

- `===`: same key/value pairs in the **same order**, with identical value **types**.
- `==`: same key/value pairs regardless of order, with value type-juggling.

Implementing this needs new dual-arch runtime routines plus lowering for both operators (and boxed-`Mixed` element comparison for heterogeneous arrays). Filing as a standalone feature.

### Context

Found while reviewing the foreach Mixed-key array-write fix: a rebuilt `Array(Mixed)` could not be compared against an array literal — but the gap turned out to be general (plain `[1,2,3] === [1,2,3]` fails too), so it is unrelated to that fix and tracked separately here.

nahime0:
This looks **resolved on current `main`** — worth verifying and closing.

`main` gained `__rt_array_strict_eq` (`src/codegen_support/runtime/arrays/array_strict_eq.rs`,
merged with PR #559). I hit it while merging `main` into #678: before the merge every one of
these was `EIR backend error: unsupported EIR backend feature: strict_eq for PHP type Array(Int)`,
and after it they all pass.

Verified against PHP 8.4 on macOS ARM64:

```php
<?php
[...]

Issue #388, opened 2026-06-18
var_dump prints NULL for nested arrays/objects instead of recursing


nahime0:
## Description

`var_dump()` now renders flat heterogeneous indexed arrays and flat associative/sparse arrays correctly (see #328, which is effectively resolved by the EIR Mixed/hash walkers). However, when an element is **itself an array or object**, the walker emits `NULL` for that element instead of recursing into a nested `array(N) { … }` / object dump.

The runtime walkers (`src/codegen/runtime/io/var_dump_walk.rs`) are flat: both the Mixed indexed walker and `__rt_var_dump_hash` fall back to `NULL` for boxed array/object element tags (the file preamble documents this limitation). `print_r` already recurses correctly (`print_r_walk.rs`), so it can serve as the reference model.

## Steps to reproduce

```php
<?php
var_dump([1, [2, 3]]);
var_dump(["a" => [1, 2]]);
```

## Actual behavior

```text
array(2) {
  [0]=>
  int(1)
  [1]=>
  NULL
}
array(1) {
  ["a"]=>
  NULL
}
```

## Expected behavior (PHP 8.4)

```text
array(2) {
  [0]=>
  int(1)
  [1]=>
  array(2) {
    [0]=>
    int(2)
    [1]=>
    int(3)
  }
}
array(1) {
  ["a"]=>
  array(2) {
    [0]=>
    int(1)
    [1]=>
    int(2)
  }
}
```

## Notes

- Requires threading a depth/indentation parameter through the var_dump walkers (PHP indents nested levels by 2 spaces each) and dispatching boxed array/object element tags back into the array/hash/object walkers, on both ARM64 and x86_64.
- `print_r_walk.rs` is a working recursive reference.
- Flat cases (covered by `codegen::io::printing::test_var_dump_mixed_indexed_array`) already work.

nahime0:
Additional current-`main` evidence for the object side of this issue: top-level object dumping has no body at all, so a complete object walker is also a prerequisite for the nested-object recursion already described here.

Verified on `17adb3a09c8c2f79d1851298d9472ea1a729b86c` (macOS ARM64):

```php
<?php
class Box {
    public int $answer = 42;
    public string $name = "box";
}

$value = new Box();
var_dump($value);
```

PHP 8.4.19:

```text
object(Box)#1 (2) {
  ["answer"]=>
  int(42)
  ["name"]=>
  string(3) "box"
}
```

elephc:

```text
object(Box)
```

[...]

PR #438, opened 2026-06-30
fix(array|debug): array/hash strict equality and var_dump nested recursion

Guikingone:
#424: array/hash strict equality (===) was unsupported by the EIR backend, causing a compile error. New runtime helpers __rt_array_strict_eq and __rt_hash_strict_eq compare indexed arrays and hashes element-by-element following insertion-order chains. Pointer-identity short-circuit handles aliases and cycles. String elements compare via __rt_str_eq; Mixed elements via __rt_mixed_strict_eq. Loose equality (==) deferred.

#388: var_dump printed NULL for nested arrays/hashes instead of recursing. New __rt_var_dump_value recursive renderer dispatches on runtime tags: arrays/hashes recurse with 2-space-per-level indent, Mixed cells unbox and redispatch. Depth cap at 128 prevents stack overflow on cyclic arrays. Rodata literals updated to remove baked-in 2-space prefix; indentation is now driven by __rt_var_dump_spaces. Objects (tag 6) remain NULL (documented limitation).


nahime0:
Closing this draft as superseded.

Both claimed bugs are already on `main` with a stricter implementation: `#424` via `#559` (`__rt_array_strict_eq`), `#388` via `#626` (`var_dump` walkers). This SHA still targets deleted `src/codegen/runtime/` / `src/codegen_ir/` paths, and landing it would regress object dumps and packed-vs-hash `===`.

Please do not rebase `56b2c01e`. Nothing unique remains.

PR #443, opened 2026-06-30
fix: strtotime x86_64 weekday scan uses min(remaining, 16) instead of fixed 16

### LINKED ISSUE #391: strtotime x86_64 weekday-modifier scan uses fixed length 16 instead of min(remaining, 16)
## Description

Low-severity robustness inconsistency between targets in the `strtotime` weekday parser. Not a confirmed miscompilation today — flagged during review and could not be exercised on a Linux host.

In `src/codegen/runtime/system/strtotime/weekdays.rs`, the **x86_64** modifier path computes a remaining-length in `r11` then discards it and hard-codes `mov rcx, 16` before `match_word` (a comment acknowledges "simpler: pass len-capped"). The **ARM64** path correctly passes `min(remaining, 16)`.

Passing a fixed `16` lets `match_word` scan into the zero-padded tail of the lowercased 16-byte buffer. It is *functionally* safe for the current tables because `match_word` requires an exact full-length match and treats the trailing `\0` as a word boundary, but it diverges from the ARM64 logic and would become fragile if any table entry length or boundary rule changed.

The sibling x86_64 paths in `offsets.rs`, `textual_date.rs`, and `first_last_day.rs` all correctly compute `min(remaining, 16)`, so this is an isolated inconsistency.

## Suggested fix

Replace `mov rcx, 16` with the `min(remaining, 16)` computation used by the other x86_64 paths and by the ARM64 weekday path, so both targets match.

## Notes

- Per the supported-target policy, target-specific lowering paths should stay consistent.
- Verify on Linux x86_64 (Docker scripts) since it cannot be exercised on macOS/arm64.

MADEVAL:
The x86_64 weekdays sub-routine passed a hardcoded 16 as the available byte count to match_word, instead of min(remaining, 16). The ARM64 version and every other x86_64 call site already computed the cap correctly.

Replaced the commented-out dead computation and fixed mov rcx,16 with the same end-minus-cursor pattern used in offsets.rs:353-357, capped via cmovae.

Fixes #391

Verification:
- cargo check: clean
- 76 strtotime codegen tests passed on Linux x86_64 (WSL)

### DIFF
--- a/src/codegen/runtime/system/strtotime/weekdays.rs
+++ b/src/codegen/runtime/system/strtotime/weekdays.rs
@@ -180,11 +180,11 @@ fn emit_weekdays_linux_x86_64(emitter: &mut Emitter) {
     emitter.instruction("mov QWORD PTR [rsp + 112], rdi");                      // save weekday cursor before match
     emitter.instruction("lea rdi, [rbp - 64]");                                 // candidate ptr = lc16 base
     abi::emit_symbol_address(emitter, "rsi", "_strtotime_keyword_tab");         // keyword table base
-    emitter.instruction("mov rcx, r10");                                        // copy end
-    emitter.instruction("mov r11, QWORD PTR [rbp - 80]");                       // reload original ptr
-    emitter.instruction("add r11, rax");                                        // add modifier-consumed
-    // Compute remaining: end - (orig_ptr + modifier_consumed + ws_skipped) — simpler: pass len-capped
-    emitter.instruction("mov rcx, 16");                                         // cap to 16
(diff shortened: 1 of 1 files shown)


Guikingone:
Hi @MADEVAL 👋🏻 

Looks like we're trying the same issue twice (see https://github.com/illegalstudio/elephc/pull/429), @nahime0 any thoughts?

nahime0:
Hi @Guikingone, @MADEVAL. Yes the implementation is the same. Given that here are missing regression tests and that this came later, I'm gonna close this and review the other one.

PR #446, opened 2026-07-01
feat(platform): add full Windows x86_64 PE32+ cross-compilation support

### LINKED ISSUE #379: Support PE file (windows)
Hello, first of all,
 congratulations on this compiler, it’s absolutely amazing!

 Could you tell me if it is (or will be) possible to create Windows PE files (i.e., .exe files)?

Thanks

Guikingone:
Closes #379.

## Status

This PR adds the experimental `windows-x86_64` target and the native Windows
runtime/bridge path needed to produce standalone PE32+ executables.

- Rebased on `origin/main` at `574105c407` after reconciling the core-align
  integration.
- Exactly one squashed feature commit: `1036aa7da2`.
- A fresh correctness CI run has been triggered for this exact SHA.
- Windows remains experimental until its native Windows, Wine, bridge, eval,
  TLS, web, and strict-codegen gates are green on one revision.

The accepted spellings are `windows-x86_64` and
`x86_64-pc-windows-gnu`. `x86_64-pc-windows-msvc` is rejected explicitly:
generated programs use the GNU/MinGW Windows ABI, not the MSVC ABI.

### Target, PE/COFF, ABI, and toolchain

- GNU MinGW-w64 PE/COFF assembly/linking, plus opt-in Clang integrated
  assembler and LLD support.
- Microsoft x64 imports, shadow space, stack arguments, FP/GP classes,
  variadics, callbacks, indirect bridge calls, unwind metadata, exceptions,
  fibers, runtime data, DLL exports, and worker-stack setup.
- Cross-host bridge archive discovery/building, including PCRE2's MinGW
  Autoconf tuple/sysroot path and Windows `libiconv` linkage.

### Win32 runtime and PHP-visible behavior

- Win32/CRT/Winsock implementations for memory, clocks, entropy, environment,
  processes, streams, sockets, DNS, console I/O, filesystem metadata, directory
  iteration, and temporary files.
- UTF-8/UTF-16 path conversion with long-path, drive/UNC root, canonicalisation,
  and path-sized-buffer handling.
- `proc_open()` and its sparse descriptors, process status/termination/close,
  fresh replacement of the by-reference pipes array, pipe polling and sharing.
- `tempnam()`, stream socket server/client requirements, Windows `unsetenv()`,
  `number_format()` x86_64 varargs, and target-aware stream transport exposure.
- Target-aware omission of PCNTL/POSIX surfaces on Windows across direct calls,
  callable lookup, reflection, `function_exists()`, extension reporting, and
  Magician/eval.

### Windows SAPI, GD, and TLS

- Full `sapi_windows_*` AOT/eval catalog: code-page aliases/get/set/conversion,
  UTF-8 detection, VT100 mode, and safe deferred console-control callbacks.
- Windows-only `imagegrabscreen()` / `imagegrabwindow()` through GDI.
- Native trust roots, client/server TLS and listener wrapping; stream-context
  verification, CA, ALPN/SNI, tickets, fingerprints, local cert/key, protocol,
  and rustls-backed TLS 1.2 cipher-subset policies.
- TLS option ABI v6, with retained v3/v4/v5 decoding; encrypted PKCS#8 and
  legacy encrypted RSA/EC PEM support fail closed on incorrect passphrases.

## PHP source audit

The implementation is compared with frozen PHP 8.5.6 source
[`fcc29c8d6d6ee6f5ba2d941f0a2a6ea6aa6ee633`](https://github.com/php/php-src/tree/fcc29c8d6d6ee6f5ba2d941f0a2a6ea6aa6ee633),
including Win32 path/codepage/process behavior, GD capture, Windows SAPI,
OpenSSL stream contexts, and PCNTL build guards.

The audit closed the active gaps around native PE call boundaries, callback
stack arguments, TEB stack metadata during fiber switches, process-pipe array
replacement, Windows SAPI value boxing, PCRE2 cross-compilation, iconv,
`unsetenv()`, `tempnam()`, and target-specific builtin/transport visibility.

## Local validation

- `cargo build`
- `cargo test -p elephc --lib -- --format terse` — 2,667 passed
- `cargo test --test builtin_parity_tests -- --format terse` — 7 passed
- `cargo test -p elephc-magician --features curl --lib -- --format terse` —
  1,392 passed after resolving the catalogued Windows path constants in eval
- generated builtin docs: 1,119 builtins, 2,160 pages, zero audit errors;
  compatibility coverage is 924 functions / 142 classes / 1,106 constants
- focused PE compile/link, native-dependency, TLS, callback, process, stream,
  SAPI, number-format, and Windows visibility regressions
- PE disassembly validates every emitted `__rt_setjmp` boundary stages its
  jump buffer in the internal SysV `rdi` register on Windows x86_64
- GNU/MinGW CI stages a checksum-verified static `libiconv.a`; Windows
  `proc_open()` duplicates default standard handles as inheritable child copies
- Windows iconv FFI resolves GNU/MinGW's exported `libiconv_*` symbol names
- Windows `proc_open()` preserves optional cwd arguments before its fresh-pipes
  allocation can clobber volatile input registers
- Windows `proc_open(["socket"])` uses CRT-compatible pipe endpoints for child
  standard streams instead of unreadable raw Winsock `STDIN`/`STDOUT` handles
- CI failure probes now report the failing stage in Fiber/GC and expose the web
  request URI/query in response headers; a pipe-stdin control accompanies the
  socket round-trip test
- The pipe control now reports the parent's `fwrite` byte count and a marker
  from the child before its `fread`; the web probe inherits server diagnostics
  in CI and enables per-request access logging, separating process-handle
  wiring from request dispatch/output capture.
- The web probe now resolves Wine's fault address back to a PE symbol when
  shutdown fails; the proc probes also capture child stderr through descriptor
  2 to distinguish an unwired stdout/reader from a child that never ran.
- Web fault resolution falls back to the nearest MinGW PE symbol when the CI
  executable lacks DWARF line data.
- The proc helper records its stdin result to a file as a transport-independent
  witness in case every inherited standard handle is lost, and reports whether
  the child started before it blocks on stdin; it also records `fwrite`'s
  return values for child stdout and stderr. The socket case compares an
  unfiltered write to the filtered write on the same CRT-backed endpoint.
- PE callback-method and `proc_open()` compile gates pass locally. The eval
  bridge overflow PE reaches `-liconv`; local Homebrew MinGW lacks that archive,
  while the Windows CI MSYS2 jobs install and explicitly verify
  `mingw-w64-x86_64-libiconv`.
- Native-Windows CI exposed a platform-boundary bug shared by many array,
  callback, hash, cleanup, serialization, generator, SPL, property, and eval paths: hand-written `__rt_*`
  helpers use the internal SysV register order, while generated PHP functions
  use MS x64. The lowering/runtime call sites now use the dedicated internal
  helper register map; generated callback and C-ABI calls retain MS x64.
- The ABI sweep also covers generator/Fiber and SPL constructors, mixed-property
  fetches, dynamic-property hash probes, reference cells, clone, numeric-string
  coercion, and error-handler dispatch; direct generated/native callback paths
  remain on the platform ABI.
- The Windows ABI register-order regression passes locally, as do PE compile
  tests for `array_sum` and the SPL/ArrayAccess/GC callback matrix. Runtime
  acceptance for these cases remains assigned to native Windows/Wine CI.
- On native Win32 I/O, the standard CRT descriptors now fall back to
  `GetStdHandle` only when `_get_osfhandle(0..2)` is invalid; nonstandard CRT
  descriptors and raw socket handles keep their existing paths.
- `git diff --check` and assembly-comment alignment checks

Local Windows execution is not claimed: this macOS host has neither Wine nor
KVM. Native Windows and Wine execution remain CI acceptance gates.

## Remaining promotion gates

This PR does not yet claim 100% closure. Before promoting Windows into the
first-class target matrix, the same revision still needs:

- a fully green native Windows/Wine CI run;
- full OpenSSL cipher-string/security-level parity (`DEFAULT`/`HIGH`,
  exclusions, `@SECLEVEL=`, and TLS compression are intentionally fail-closed
  where rustls cannot represent them);
- the final php-src ledger review after any CI-driven fixes.

CI runtime/caching/sharding optimisation will be reviewed only after these
correctness gates are green.


nahime0:
Heads-up from #730 (feat/curl): this branch carries the same `static CELL: OnceLock<Mutex<CString>>` addition in `crates/elephc-tz/src/abi.rs` as #632 (possibly inherited from its tz work) — a process-global buffer cell handing out `as_ptr()`.

On #730 that pattern was root-caused as an intermittent cross-thread use-after-free (the PDO live-suite garbage-SQLSTATE failure) and removed from all bridge crates in favor of `thread_local!` cells; the rule is now in CONTRIBUTING.md (§ "Expose a stable C ABI") and enforced by a CI tripwire (`tests/ffi_buffer_hygiene.rs`) that will flag this line after #730 merges. When this branch rebases, please adapt the cell to the `thread_local!` shape the surrounding `abi.rs` cells use post-#730.

Guikingone:
Rebased onto current `main`. `crates/elephc-tz/src/abi.rs` now uses per-thread `RefCell<CString>` cells for returned strings, so the process-global buffer race from this comment is no longer present. CI is running on the rebased head; separate Windows process/web failures remain under investigation.

PR #451, opened 2026-07-01
feat: PHP eval() support via the Magician interpreter

nahime0:
## Summary

This PR adds full `eval()` support to elephc through **Magician** (`crates/elephc-magician`), a PHP interpreter that ships as an optional static library. Programs that never call `eval()` are completely unaffected: the interpreter is linked only when eval is actually used, and native-only binaries keep their current size and dependencies.

1,207 commits · 2,128 files · +268k/−11.5k lines

## How eval is executed

Eval sites are compiled with a three-tier strategy, picked per call site by coordinated classifiers in the plan, lowering, and codegen layers:

1. **Literal AOT** — string-literal fragments that the compiler can prove static (scalar expressions, static builtin calls, array literal reads, switch/match/ternary, user-function calls, …) are compiled ahead of time into native EIR functions. No interpreter, no bridge, no runtime cost beyond a direct call.
2. **Scope AOT** — fragments that read or write caller locals but are otherwise static run as native functions with direct read parameters or scope write-back, still without linking the interpreter.
3. **Bridge** — everything else (dynamic code, declarations, object semantics) executes in the Magician interpreter through the `__elephc_eval_*` FFI bridge, with bidirectional scope synchronization between native locals and the eval scope.

## Native ↔ eval interop

- Caller locals are synchronized across the eval boundary in both directions, including type-changing writes (scalar slots widen to boxed Mixed storage) and container values.
- eval'd code can call AOT functions, methods, and constructors with full PHP visibility semantics: private/protected access checks by calling scope, private-method shadowing over child overrides, magic `__call`/`__callStatic` fallbacks, and late static binding.
- eval-declared functions, classes, closures, and constants are callable from subsequent eval fragments and observable through the dynamic probes (`function_exists`, `defined`, reflection, …).

## The Magician interpreter

- 415+ PHP builtins implemented through a declarative `eval_builtin!` registry (inventory-collected, one home file per builtin), covering strings, arrays, math, regex (PCRE + mb_ereg), JSON, filesystem, date/time, SPL, and reflection.
- Static-vs-eval builtin parity is enforced by dedicated parity gates (`tests/builtin_parity_tests.rs`).

## Documentation

- New `docs/php/eval.md` covering supported semantics and the AOT tiers.
- The generated builtin docs now render a **dual support matrix**: every builtin page and index shows whether it is available in compiled (AOT) code and inside `eval()`, with links to both implementations. The exporter reads both registries, so the matrix cannot drift from the parity tests.

## Compiler improvements landed along the way

- PHP 8.3 auto-key semantics for array literals (bool/float/numeric-string and leading-negative keys).
- `array_key_exists()` over boxed Mixed containers with runtime tag dispatch; float key normalization.
- `is_array`/`is_integer` dispatchable as runtime string callbacks.
- ReflectionParameter promoted-parameter metadata, `newInstanceArgs()` with indexed argument lists, attribute array-literal arguments.
- First-class-callable fixes (by-ref builtin params, magic fallbacks).

## Testing

- 935 eval codegen tests plus 7 dedicated eval test suites (closures, callables, reflection invocation, …), all green on macOS.
- 928 Magician unit tests; full codegen/error/parser/lexer suites green.
- Branch is rebased on latest main (0.26.1 + `declare` support) with the generated docs byte-identical to CI regeneration.


nahime0:
This is almost ready for a first merge. It's still experimental, but it's a good candidate for the next release

PR #454, opened 2026-07-02
conf(ci): cancel superseded runs, build once via nextest archive, gate benchmark to main

Guikingone:
## What

Three CI-time optimizations. No change to what is tested or on which targets.

### 1. Cancel superseded runs
Adds a `concurrency` group with `cancel-in-progress: true`. A force-push/rebase now cancels the in-flight run for that ref instead of letting ~55 stale jobs run to completion.

### 2. Build once, run from a nextest archive
Previously every one of the **3 + 48** test jobs rebuilt the whole compiler + 7 bridge crates before running its slice of tests. Now:
- a new **`build-archive`** job compiles the workspace **once per platform** and packages the test binaries with `cargo nextest archive`;
- the **16 codegen shards** and the **non-codegen** job download that archive and run from it (`cargo nextest run --archive-file …`), no recompilation.

The bridge staticlibs (`libelephc_*.a`) are linked into compiled PHP programs at **run time** (via `ld`), so nextest can't discover them automatically — they're added to the archive via an `archive.include` list in `.config/nextest.toml`. Doctests (which can't run from an archive) run in `build-archive` where the workspace is already built.

### 3. Gate the benchmark job to `main`
The benchmark suite does a full `--release` build; it now runs on pushes to `main` only, not on every PR.

## Validation (local, macOS aarch64)
- Archive captures the 7 bridge staticlibs (`7 extra paths`) + the `elephc` CLI.
- Codegen **shard 1/16 run from the archive**: 315/315 pass, incl. image tests linking `libelephc_image.a`.
- Non-codegen suite (incl. the `ir_backend_parity` FCC case) runs from the archive under `--profile ci`.
- `ci.yml` YAML + `nextest.toml` parse.

The Linux shards can only be confirmed by this PR's own CI run (staticlib filenames are identical across targets, so it should carry over).

## Note
Also raises the nextest slow-timeout for the ~65–70s `ir_backend_parity` first-class-callable case (it sits just over the 60s cap and was terminated spuriously). This same one-test override also appears in #439; dedupe on whichever merges first.


nahime0:
@Guikingone I think we really need this one. I'm gonna merge main, fix the CI and merge if everythink is green.

nahime0:
Changed the requirement matrix to: 

  - build-archive-linux-x86_64
      - non-codegen-tests-linux-x86_64
      - codegen-tests-linux-x86_64

  - build-archive-linux-aarch64
      - non-codegen-tests-linux-aarch64
      - codegen-tests-linux-aarch64

  - build-archive-macos-aarch64
      - non-codegen-tests-macos-aarch64
      - codegen-tests-macos-aarch64
      
So that linux tests can start as soon as the archive is available, without waiting the build of macos (which tipically has a long queue).

nahime0 (approved):
Ok

PR #459, opened 2026-07-05
feat: accept declare(strict_types=1) and declare directives

chadmandoo:
## What

Accepts `declare(strict_types=1);` (and `declare(...)` directives generally) as a top-of-file statement instead of erroring. Every strict-typed PHP file opens with it, so this unblocks compiling any codebase that uses `declare(strict_types=1)`.

## How

Parses the `declare(directive=value)` form as a directive statement rather than treating `directive=value` as an assignment expression (which produced *"Invalid assignment target"*). `strict_types` is accepted and does not alter runtime behavior in the compiled output (the compiler is already strict), so byte-parity with PHP is preserved.

## Verification

Byte-parity vs PHP 8.5 for files opening with `declare(strict_types=1);` + a regression test.


nahime0:
This will be discussed in the next core team meeting. It touches an important area about static / dynamic behaviour
@Guikingone

nahime0 (approved):
Syntax aligned to php. Ok

PR #460, opened 2026-07-05
feat: implement mb_ereg_match() builtin (start-anchored regex)

chadmandoo:
## What
Implements `mb_ereg_match($pattern, $string): bool` — an **anchored-at-start** regex test (verified vs PHP 8.5: `mb_ereg_match('bc','abc')` = false, `mb_ereg_match('ab','abc')` = true; `\z`/`$` force end-anchoring).

## How
Adds `__rt_mb_ereg_match` (AArch64 + x86_64) reusing the existing PCRE2 regex runtime:
- Materializes the bare (delimiter-less) pattern verbatim as a C string via `__rt_cstr2` (no `__rt_preg_strip`/`__rt_pcre_to_posix` — mb_ereg patterns carry no delimiters).
- Compiles with `pcre2_regcomp` flags `0` (the POSIX wrapper uses Perl syntax, so `\z` etc. work).
- Executes `pcre2_regexec` and enforces the start-anchor by requiring `regmatch[0].rm_so == 0`.

Registered via the `builtin!` registry (`returns: Bool`) and added to the regex runtime-feature trigger so PCRE2 links only when needed.

## Verification
Byte-parity vs PHP 8.5 (prefix, `^…$`, `[a-z]+\z`, `^view(-[a-z0-9-]+)?\.php$`) + a regression test. Builds standalone on `main`.


chadmandoo:
Pushed a follow-up commit: mb_ereg_match must be excluded from the runtime callable-wrapper dispatch table (it is registry/EIR-only — no frozen legacy-backend emitter), otherwise any program that instantiates the string-callable dispatch table fails to link with an unresolved `_fn_mb_u_ereg_u_match` (same pattern and rationale as the existing serialize/unserialize exclusions). Caught by the full codegen suite (array_map/usort/array_filter tests).

PR #461, opened 2026-07-05
feat: htmlspecialchars/htmlentities ENT_* flags + ENT_* constants

chadmandoo:
## What
Adds the `ENT_*` flag constants and extends `htmlspecialchars()` / `htmlentities()` to accept the optional `flags` + `encoding` arguments, so the common `htmlspecialchars($s, ENT_QUOTES)` form compiles (previously: *"Undefined constant: ENT_QUOTES"* + single-arg only).

## How
- Registers `ENT_QUOTES`(3), `ENT_COMPAT`, `ENT_NOQUOTES`, `ENT_HTML401`, `ENT_HTML5`(48), `ENT_XHTML`, `ENT_XML1`, `ENT_SUBSTITUTE`, `ENT_IGNORE` in the checker constant map (`driver/init.rs`) and the codegen prescan value table (`prescan.rs`).
- Extends both `builtin!` registrations to `params: [string, flags: Int = 11, encoding: Str = "UTF-8"]` (PHP's default flag set).
- An arity-tolerant `lower_html_escape` uses the subject (operand 0) and reuses `__rt_htmlspecialchars`.

## Note
`__rt_htmlspecialchars` implements ENT_QUOTES behaviour, which matches PHP's default and the ENT_QUOTES call form (byte-parity verified: `&lt; &quot; &#039; &amp;`). The flags are accepted but not yet distinguished — a flag-aware runtime (ENT_HTML5 `&apos;` vs `&#039;`, ENT_NOQUOTES) is a follow-up. Builds standalone on `main`.


chadmandoo:
Pushed a follow-up commit syncing the `legacy_builtin_call_sig` parity golden: this PR gives htmlspecialchars/htmlentities 3-param registry signatures, so their golden arms move from the 1-param `fixed(&["string"])` group to an `optional()` arm mirroring the registry defaults — otherwise `derived_signatures_match_legacy` fails ("param count differs (derived=3, legacy=1)").

PR #463, opened 2026-07-05
fix: narrow `$x === false` guard (int|false returns as int after a throw-guard)

chadmandoo:
## What
`if ($v === false) { throw; } return $v;` on an `int|false` value declared `: int` was rejected — the return was inferred as `Union([Int, Bool])`.

## Why
The if/else flow-narrowing already keeps the guard complement after a diverging body (`throw`), but `type_guard_narrowing` only recognized `is_*()`/`instanceof` guards — not the `$x === false` strict comparison — so the complement never stripped the false-ish `Bool` member.

## Fix
Recognize `$var === false` / `false === $var` (`BinOp::StrictEq`): narrow to `Bool` in the then-branch, strip it in the else-branch (`int|false` → `int`). Byte-parity regression test; builds standalone on `main`.


chadmandoo:
Pushed two follow-up commits extending the same narrowing seam: (1) property-access flow-narrowing across ternary branches (`$this->raw instanceof Message ? $this->raw : new Message($this->raw)` with a union-typed property) via synthetic env keys the property-read inference consults; (2) a unified guard_narrowing() entry that adds is_null() to the predicate set and applies the same guards to simple property receivers at statement level (the `if (!$this->x instanceof C) { throw; } return $this->x;` pattern), with conservative purge-on-call/write/loop-entry soundness. Byte-parity + regression tests per fix.\n\nNote: this PR's tests (pre-existing and new) use `declare(strict_types=1)` in fixtures, so it depends on #459 landing first.

PR #464, opened 2026-07-05
feat: in_array() accepts the optional strict (3rd) argument

chadmandoo:
## What
`in_array($needle, $haystack, true)` failed to compile — *"in_array() takes exactly 2 arguments"*. The builtin capped arity at 2 (`max_args: 2`) and `lower_in_array` asserted exactly 2 operands.

## Fix
Remove the cap; accept 2–3 args. `lower_in_array`'s per-type comparison (int/float/bool by value, string by byte-equality) is already type-specific, so for homogeneously-typed arrays the result matches strict `===` membership (byte-parity verified: string[] + int[], present + absent). The `strict` flag itself is not separately consulted yet — a heterogeneous-array type-juggling distinction would need it (documented follow-up). Builds standalone on `main`.


nahime0:
I'll take over this to implement the strict / loose comparison. We have some helpers that we can reuse (like __rt_str_loose_eq).

nahime0 (approved):
All done, ready to merge

PR #465, opened 2026-07-06
feat: parse typed class constants (PHP 8.3 `const TYPE NAME = ...`)

chadmandoo:
PHP 8.3 typed class constants (`private const string X = '…'`) failed to parse ("Expected '=' after class constant name"). Accept an optional type expression between `const` and the constant name in class-like bodies; the declared type is validated against the initializer.

Found compiling real-world strict-typed code (a PHP 8.5 codebase with ~42 such constants across 15 files). Byte-parity vs PHP 8.5 + parser regression tests.


nahime0:
Maintainer follow-up context:

The original contribution introduced the PHP 8.3 typed class-constant syntax and its initial parser coverage. During review, we found that the parsed type annotation was discarded instead of being carried through the AST, name resolution, class schema, type checker, and Reflection metadata. As a result, the compiler accepted the syntax without enforcing initializer or inheritance compatibility, while `ReflectionClassConstant::hasType()` and `getType()` still reported every constant as untyped.

We intervened to complete the feature end to end before merge:

- retain and resolve declared constant types for classes, interfaces, traits, and enums;
- validate initializer types, PHP-forbidden constant types, and covariant parent/interface overrides;
- use declared types when inferring class-constant access;
- expose named, nullable, union, and intersection metadata through `ReflectionClassConstant`;
- preserve safe untyped defaults for enum cases and dynamically declared eval constants;
- add focused parser, diagnostic, codegen, Reflection, and eval regressions;
- update the class documentation, example, and `CHANGELOG.md`;
- merge the current `main` branch and verify the result with an offline build, focused tests, the classes example, and `git diff --check`.

The original contribution remains the basis of the feature; the maintainer commits complete the compiler-wide semantics and regression coverage required for it to land safely.

PR #467, opened 2026-07-06
feat: resolve the object pseudo-type and Closure type-hint

chadmandoo:
Two name-resolution completeness fixes for type positions: the `object` built-in pseudo-type resolved as a namespaced class name ("Unknown type: App\object"), and the `Closure` type-hint was unrecognized in param/return/property positions. `object` maps to the top object type; `Closure` maps to the callable representation but stays permitted as a PROPERTY type (PHP allows a Closure-typed property while forbidding bare `callable` there). Byte-parity vs PHP 8.5 + regression test.

### DIFF
--- a/src/name_resolver/names.rs
+++ b/src/name_resolver/names.rs
@@ -82,7 +82,7 @@ pub(super) fn resolve_type_expr(
         }
         TypeExpr::Named(name) => {
             let raw = name.as_str();
-            if matches!(raw, "array" | "mixed" | "callable" | "void") {
+            if matches!(raw, "array" | "mixed" | "callable" | "void" | "object" | "Closure") {
                 TypeExpr::Named(name.clone())
             } else {
                 TypeExpr::Named(resolved_name(resolve_special_or_class_name(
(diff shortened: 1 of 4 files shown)


nahime0:
Thanks a lot for digging into this and for the regression coverage; the namespaced-object failure and the missing Closure hint were real gaps, and the PR helped surface them clearly.

We're going to supersede this with a follow-up from current main rather than land this branch as-is. A few things shifted under us:

• main already maps object to PhpType::Object("") (accept any object). Mapping it to Mixed would widen the type incorrectly (e.g. accepting scalars) and regress that.
• In PHP, object is a pseudo-type (must not be namespace-prefixed), but Closure is a class name — so unqualified Closure inside a namespace means App\Closure, not \Closure. Treating Closure like mixed/callable in name resolution isn't PHP-compatible.
• Property rules also need ?Closure / Closure|null, not only bare Closure.

We'll keep the intent (and credit the report) in the replacement PR and link back here when it's up. Really appreciate the contribution.

PR #468, opened 2026-07-06
fix: type enum-case parameter defaults as the enum, not the backing scalar

chadmandoo:
A parameter default of an enum case (`function f(Level $l = Level::Low)`) was typed as the case's BACKING scalar, so the declared enum parameter type rejected its own default. Resolve the default's type as the enum object. Byte-parity vs PHP 8.5 (defaulted + explicit call) + regression test.

### DIFF
--- a/src/types/checker/type_compat/declarations.rs
+++ b/src/types/checker/type_compat/declarations.rs
@@ -9,7 +9,7 @@
 //! - Rules here define accepted programs, so PHP covariance, inheritance, and extension-specific constraints must stay explicit.
 
 use crate::errors::CompileError;
-use crate::parser::ast::{Expr, TypeExpr};
+use crate::parser::ast::{Expr, ExprKind, StaticReceiver, TypeExpr};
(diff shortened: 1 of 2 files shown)


nahime0:
Thanks for identifying the enum-case default issue and for the initial implementation.

I’m closing this PR because the current approach is not safe to merge:

- Matching only the scoped receiver allows undefined cases such as `A::Nope`.
- It also accepts scalar class constants such as `Foo::BAR` as defaults for `Foo`-typed parameters.
- The regression test already passes on the base behavior when the free function is called, so it does not exercise the failing declaration-validation path.
- `self::`, `static::`, and `parent::` receivers are not covered.

Since this branch was opened, `main` has gained a post-schema mechanism for deferred object-default validation. The correct fix should build on that mechanism and resolve scoped constants semantically once enum/class metadata is available, rather than bypassing validation during syntactic inference.

The base branch has also moved substantially, so we will prepare a clean replacement PR from the current `main`, reference this PR, and preserve credit for the original report and reproducer.

Closing as superseded by the forthcoming implementation. Thanks again for surfacing the problem.

PR #469, opened 2026-07-06
fix: keyword-spelled member names — declarations and case-insensitive access

chadmandoo:
Three commits making keyword-spelled members work end-to-end. The lexer folds keywords case-insensitively, so a member named like a keyword loses its source lexeme: (1) semi-reserved keywords are accepted as enum-case NAMES (`case Match;`, `case Default;`); (2) the two `::`-member parse sites stop guessing spellings (a hardcoded `Token::Match => "MATCH"` disagreed with the lowercase bareword mapping) and the checker's scoped-constant lookup retries case-insensitively after an exact miss, accepting a UNIQUE fold-match (ambiguity keeps the miss; genuinely undefined cases still error); (3) the LOWERING's scoped_constant_value/interface_constant_value mirror the same retry (builtin tables declare e.g. `RegexIterator::MATCH` uppercase — pinned by the existing spl regex tests).

Byte-parity vs PHP 8.5 (case values + identifier-named-case ->name). Known documented limit: `->name` of a KEYWORD-named case reports the folded spelling — recovering the source lexeme needs spelling-preserving keyword tokens.

### DIFF
--- a/src/ir_lower/context.rs
+++ b/src/ir_lower/context.rs
@@ -297,6 +297,10 @@ impl<'m, 'f> LoweringContext<'m, 'f> {
     }
 
     /// Returns a class or interface constant expression resolved with PHP lookup order.
+    ///
+    /// After an exact miss at each level, retries case-insensitively and accepts a UNIQUE
+    /// fold-match: the lexer folds keyword spellings, so `X::MATCH` reaches lowering as
+    /// "match" while the table declares "MATCH" (mirrors the checker's scoped lookup).
     pub(crate) fn scoped_constant_value(
         &self,
         class_name: &str,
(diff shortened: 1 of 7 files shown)


nahime0:
Thanks for digging into this and for the clear write-up: the enum-case declaration path (case Default;, case Match;) plus rejecting case class is the right direction, and the observation that the lexer folds keyword spellings is spot on.

We're going to close this PR and land the fix in a follow-up, for technical reasons rather than process ones:

1. Root cause is in the lexer, not in lookup. Keywords are classified via to_ascii_lowercase() and lose the source lexeme, so MATCH / Match / match all become the same token and then the same bareword ("match"). Case-insensitive retries in the checker and EIR lowering paper over that loss; they don't restore PHP's case-sensitive member names.

2. PHP compatibility. Class constants and enum cases are case-sensitive. A unique fold-match makes spellings that PHP would reject still resolve, and ->name for keyword-named cases reports the folded form (match / default) instead of the source spelling (Match / Default). The proper fix is spelling-preserving keyword tokens (or equivalent lexeme retention) so declaration, :: access, and ->name all share one canonical source spelling — without ignore-case fallbacks.

3. Current head isn't merge-ready as-is. Removing the Token::Match => "MATCH" special case breaks test_parse_scoped_constant_named_like_keyword (AST is "match", test still expects "MATCH"), and Non-Codegen is red across the matrix. Updating that expectation would only lock in the folded spelling rather than fix it.

We'll open a new PR on current main that preserves keyword member spellings end-to-end and keeps the good parser pieces from this work (enum-case barewords + class rejection). Happy to credit this investigation there.

PR #470, opened 2026-07-06
feat: mixed/union values at narrower declared boundaries + typed callbacks over unknown-element arrays

chadmandoo:
Two related type-checker completeness fixes, both runtime-enforced PHP the checker statically rejected:

1. `types_compatible` accepts a Mixed VALUE into any declared type (param and return) and a UNION value with at least one member the declaration accepts (e.g. an `int|false` seek result passed to an `int` param on its success path — a union with NO compatible member stays rejected). Scoped to the arg/return check only; interface conformance and property assignment are unaffected. The boxed-Mixed representation already funnels at the boundary — verified byte-identical.

2. usort/uasort/array_map callbacks keep their DECLARED parameter contract when the array's element type has no literal form: Object elements (array_map's scalar dummy builder had no Object arm and fabricated Int) and Mixed/Never elements (unknown-element `array`-hinted values) now use a synthetic element binding.

Byte-parity vs PHP 8.5 + 4 regression tests. Known limit noted in the commit: array_map over OBJECT-element arrays now passes the checker but the EIR lowering doesn't support object elements yet (pre-existing wall).


chadmandoo:
Withdrawing. This was cut from an early-July base and no longer builds against current `main` (Build & Test is red across the matrix), and it's a broad change across `array_map`/`usort`/`uasort` plus callable validation. The mixed/union-at-narrower-boundary and typed-callback-over-unknown-element-array gaps it targeted overlap work that has since moved substantially in this area. Rather than rebase a stale wide-surface PR, I'd prefer to re-cut a focused one from current `main` if a specific gap remains. Happy to reopen if you'd rather iterate here.

nahime0:
I kept this PR and updated its original head instead of replacing it with a new PR.

### What changed

- Merged current `main` into the contributor branch in `f2c09e1fc`.
- Reworked the original type-boundary approach in `1e0aac900`:
  - removed the global `Mixed -> any narrower type` compatibility rule;
  - kept `Mixed` boundary acceptance only for scalar types backed by the existing runtime cast funnels;
  - removed the unsound `Union -> T` “any compatible member” rule, preserving the existing narrowing and strict return-boundary contracts;
  - added regressions proving `Mixed -> Object/Array` is rejected without a real runtime tag check.
- Corrected `fseek()` to return `Int`, matching PHP’s `0` / `-1` contract, rather than weakening union compatibility.
- Replaced fabricated `Int` callback arguments with centralized contextual callback checking:
  - unknown `Mixed`/`Never` element types preserve explicit callback declarations and use those declarations to type closure bodies;
  - known element types are still checked normally, with a negative regression for incompatible declarations;
  - applied consistently to `array_map`, `array_filter`, `array_walk`, `array_walk_recursive`, `array_reduce`, `array_find`, `array_any`, `array_all`, `array_udiff`, `array_uintersect`, `usort`, and `uasort`.
- `array_map()` now explicitly rejects known object-element arrays until the EIR callback runtime supports that layout, keeping checker/backend behavior aligned.
- Added focused codegen/error regressions and documented the callback-boundary policy.

The separate pre-existing `array_walk()` lvalue/key divergence remains outside this change and is already tracked in #537.

### Validation

Local validation is green:

- `cargo build`
- builtin signature parity
- focused callback/codegen tests
- array-builtin, callable, narrowing, type-system, and union/false-return error tests
- `git diff --check`

CI run [29724273042](https://github.com/illegalstudio/elephc/actions/runs/29724273042) now has:

- Build & Archive green on macOS ARM64, Linux ARM64, and Linux x86_64;
- Non-Codegen green on all three supported targets;
- builtin docs and generated API parity gates green;
- no failed jobs at the time of this update.

The remaining codegen/eval/web shards are still being processed by the workflow.

PR #472, opened 2026-07-06
feat: implement PHP clone — parser, checker, EIR shallow object copy

chadmandoo:
Implements `clone $expr` end-to-end:

- PARSER: contextual recognition — an identifier spelled `clone` in prefix position followed by an operand-start token parses as a unary operator; `clone` stays fully usable as a method name and `clone($x)` lands on the operator like PHP.
- AST: `ExprKind::Clone` with arms across every expression walker (resolver, optimizer passes, preludes, usage collectors, parser dependency walks); effects mirror `ObjectNew`.
- CHECKER: the operand must be a statically-typed object; the result is the same class. Classes declaring `__clone()` are rejected explicitly ("__clone() hooks are not supported yet") rather than silently skipping the hook.
- EIR: `Op::ObjectCloneShallow` — the backend allocates a same-class payload (heap-kind stamp + class-id word) and copies the statically-sized payload wordwise. Reference-property slots copy their CELL POINTER, matching PHP (a reference property stays shared with the original); dynamic-properties classes are rejected (a payload copy would share the dynamic-prop hash).

Byte-parity vs PHP 8.5: the PSR-7 `withX()` immutability round-trip (original untouched, clone mutated, object slot shared) and contextual-keyword probes. 2 regression tests.

### DIFF
--- a/src/codegen/lower_inst.rs
+++ b/src/codegen/lower_inst.rs
@@ -167,6 +167,7 @@ pub(super) fn lower_instruction(ctx: &mut FunctionContext<'_>, inst_id: InstId) 
         Op::BufferGet => buffers::lower_buffer_get(ctx, &inst),
         Op::BufferSet => buffers::lower_buffer_set(ctx, &inst),
         Op::ObjectNew => objects::lower_object_new(ctx, &inst),
+        Op::ObjectCloneShallow => objects::lower_object_clone_shallow(ctx, &inst),
         Op::DynamicObjectNew => objects::lower_dynamic_object_new(ctx, &inst),
         Op::DynamicObjectNewMixed => objects::lower_dynamic_object_new_mixed(ctx, &inst),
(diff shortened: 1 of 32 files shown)


nahime0:
Closing as superseded by main.

`clone` already landed on main (notably `5d0536c00` “Add object clone support”) with the rigorous path this PR only sketched: retained heap-backed property slots, `stdClass`/dynamic-property hash cloning, and `__clone()` invocation with visibility checks.

Relative to current main this branch does not add a useful implementation delta. The shallow wordwise payload copy without retain, plus hard-rejecting `__clone()` and dynamic-properties classes, are workarounds rather than the ownership/runtime model elephc needs. Main already covers the intended semantics (and more); the only possibly reusable bits are a couple of test fixtures (PSR-7 wither round-trip / contextual `clone($x)` + method named `clone`), which can be ported separately if desired.

Thanks a lot for the contribution.

PR #473, opened 2026-07-06
feat: covariant self-returns, use-const/enum-name parse edges, exception $previous, unknown-array foreach keys

chadmandoo:
Five bounded completeness fixes found compiling real strict-typed PHP:

- INTERFACE COVARIANCE: an implementation may return a narrower type than the interface declares (the PSR-7 `withX(): static` shape). `type_accepts` can never see it — the class under validation is not yet registered mid-construction — so conformance accepts a self-typed return from its own context: the validated interface (or one it extends), or any interface the class declares (or one those extend).
- PARSE: `use const PHP_INT_MAX;` — the lexer eagerly tokenizes such constants; the use-declaration prefix accepts the dedicated tokens by canonical spelling (imports are inert — expression uses lower to literals). `class Enum` / `new Enum` / `Vendor\Enum` — `enum` is a soft keyword (vendor precedent: marc-mabe/php-enum); parse_name and the class-decl name accept it.
- EXCEPTIONS: builtin exception constructors accept PHP's third `$previous` parameter (positional or named). Not stored — the compact throwable payload has no previous slot and `getPrevious()` is already synthesized as null — wrap-and-rethrow code compiles with byte-identical message/code round-trips.
- FOREACH KEYS: iterating an unknown-element array (`array`-hinted param/property) types keys as Mixed, not Int — the value may be associative at runtime.

Byte-parity vs PHP 8.5 on all five regression tests.


nahime0:
Follow-up review fixes have been prepared locally on top of the currently published PR head (`ea5b32659`). These commits are **not pushed yet**; they will appear in the PR diff after the next branch push.

What changed relative to the published version:

- Merged the current `origin/main` into the branch and verified a clean merge tree.
- Removed the unsound `with*` return-type heuristic. Interface calls now keep their declared return type instead of assuming the receiver interface, preventing invalid descendant-only method chains from compiling.
- Completed real covariant self-return validation for both instance and static interface methods, and added coverage for static parent overrides.
- Completed `enum` soft-keyword handling across enum/class-like declarations, type hints, import aliases, qualified names, scoped expressions, and statement-position disambiguation such as `Enum::X;`.
- Unified `use const` name parsing so lexer-tokenized constants work in single imports, comma-separated imports, grouped imports, and aliases.
- Finished the `Throwable::$previous` integration across compact payload allocation and all target-specific throw paths, including retain/null handling, `getPrevious()`, JSON/SPL/error allocation sites, and the `?Throwable` type contract.
- Kept unknown-array `foreach` keys as mixed integer/string keys instead of forcing them to integers.
- Fixed the related guardrails and handoff material: assembly-comment alignment, the stale 32-byte JsonException payload comment, changelog wording, Rust docblocks, and the exceptions example showing a wrapped exception chain.

Validation completed locally:

- `cargo build` and `cargo fmt --check`
- 15 focused end-to-end codegen regressions
- focused parser and diagnostic regressions
- `examples/exceptions` compiled and ran with `wrapped failure <- root cause`
- focused Linux x86_64 Docker regression for `Exception::$previous`
- `git diff --check`, diff-aware assembly-comment validation, and a clean merge tree against `origin/main`

The full supported-target matrix is intentionally left to CI after the branch is pushed.

PR #475, opened 2026-07-06
fix(types,ir): loop-grown arrays widen to mixed before the body is processed (#452)

### LINKED ISSUE #452: Growing a Mixed-promoted array in a loop then iterating+consuming it corrupts the heap
## Summary

Growing an array that has been promoted to mixed-element storage (`array<mixed>`) across loop iterations, then iterating it and consuming each value, corrupts the heap: the program prints wrong/empty output or segfaults. No enum, closure, or object is involved — a plain `intval()` / `(int)` cast / `+=` on the `foreach` value is enough.

## Verified on

`main` (macOS ARM64, default EIR backend). No `#349`/`#449` changes involved — reproduces on a clean `main` build.

## Minimal reproduction

```php
<?php
$vals = [];
for ($i = 0; $i < 2; $i++) {
    $vals[] = 1;      // int
    $vals[] = 2.0;    // float → promotes the array to array<mixed>, then keeps growing
}
$sum = 0;
foreach ($vals as $v) {
    $sum += intval($v);
}
echo $sum, "\n";
```

## Actual

Prints an empty line (and sometimes segfaults, `exit 139`) — the result is nondeterministic across runs but consistently wrong. Deterministically wrong across repeated runs in practice.

## Expected

```
6
```
(matches PHP: `1 + 2 + 1 + 2`)

## Notes / scope

- The trigger is a `mixed`-promoted array that is **grown further after promotion inside a loop**, then iterated and consumed. A single-shot build (`$a = []; $a[] = 1; $a[] = 2.0;` then iterate) works; a **literal** `[1, 2.0, 1, 2.0]` works. The corruption appears when the promoted array is repeatedly appended-to across `for` iterations and later iterated with per-element consumption (`intval($v)`, `(int)$v`, `$sum += $v`).
- The symptom is heap free-list corruption (wrong/empty output, or `SIGSEGV` in `__rt_heap_alloc`), so it is an ownership/refcount problem in the `array_to_mixed` promotion + subsequent `array_push` + `foreach` value lifetime, not a codegen arithmetic bug.
- Surfaced while reviewing the enum `from(mixed)` work (#449): an int-backed `Enum::from($v)` over such an array crashes, but the crash reproduces identically with `intval($v)` and no enum, confirming it is orthogonal to enums.

mirchaemanuel:
## Summary

Fixes #452. Growing an array inside a loop across heterogeneous push sites (`$vals[] = 1; $vals[] = 2.0;`) corrupted the heap: reading the affected element produced garbage or a SIGSEGV, and the string/float flavour silently corrupted values and leaked.

## Root cause

Both the type checker and EIR lowering process loop bodies in a **single pass**, so each push site is typed and lowered against the types in force the *first* time through:

- site A (`$vals[] = 1`) is lowered against `array<never>` → a **raw scalar** `array_push`;
- site B (`$vals[] = 2.0`) widens the element type and emits the `array_to_mixed` promotion.

The loop back edge brings the promoted `array<mixed>` around, so from iteration 2 the stale-typed site A writes an **unboxed i64 into boxed-cell storage**. Any value-read of that element dereferences the raw scalar as a Mixed cell pointer → SIGSEGV (a raw string payload happens to survive reads, which is why the string flavour corrupted silently instead). Diagnosed via a deterministic bisection: the minimal repro is 2 iterations + `echo $vals[2]` — no `foreach` needed.

## Fix

A shared prescan (`src/types/checker/loop_widening.rs`) walks a loop body's statements recursively, collects its `$name[] = value` push sites, and reports the locals whose indexed-array element type **joins to `mixed`** across all sites (`never` adopts the other side, equal types stay, anything else widens). Then:

- the **checker** widens those locals to `array<mixed>` in the `TypeEnv` before checking each loop body (`while`/`do-while`/`for`/`foreach`; for `foreach` after the value/key bindings so a push of the loop variable joins with its real element type);
- **EIR lowering** widens its `local_types` the same way and **materializes the promotion once before the loop** (in-place `ArrayToMixed` + storeback), so every push site inside the body boxes its value.

Same-type pushes and `never → T` growth are untouched: on those paths the generated IR is **byte-identical to `main`** (verified) — the widening only fires on the representation-changing join to `mixed`. As a bonus, loops that previously re-promoted per iteration now hoist a single conversion out of the loop.

## Verification

- 6 regression tests in `tests/codegen/array_basics.rs`: element read, foreach sum, int+string flavour, float-first ordering, push under `if`, `while`-loop variant (the walker's recursion branches).
- Adversarially reviewed: repro family (2–5 iterations, `while`/`do-while`/nested loops, push under `if`/`try`, foreach value-var push) all match PHP with a clean `--heap-debug`; iterate-and-push snapshot semantics, conditional `break` before the promoting site, COW copies between iterations, and `continue` skipping a site all verified against PHP.
- No regressions: arrays + runtime_gc + optimizer (1231 tests), control/types/loops (502), error_tests (993), lib — all green; build warning-free.

## Found during review (pre-existing, not this change)

- The parser rejects an array push (and comma-separated expressions) in the `for` update clause — filed separately with a repro.


mirchaemanuel:
CI fix pushed (`c1d71ae3`) — the failing `test_multiple_iterator_contains_detach_and_assoc_null_info_error` was a real defect in this PR, thanks for the run.

**What was wrong:** the widening prescan treated any pushed variable *unknown at loop entry* as `mixed` evidence. The synthesized `MultipleIterator::detachIterator` body rebuilds an `array<Iterator>` through exactly such a variable (`$candidate = $this->iterators[$i]; $newIterators[] = $candidate;`), so the array was spuriously widened to `array<mixed>` and its typed-property storeback stopped compiling — for every program touching `detachIterator`.

**The fix:** unknown never defaults to `mixed`. A pushed variable joins its loop-entry type with self-evident **literal** assignments found inside the body (so `$x = 1; $vals[] = $x; $y = 2.0; $vals[] = $y;` still widens); non-literal in-loop sources yield no evidence and are excluded from the join. This needed two layers — the shared prescan, and ir_lower's lookup closure, which was silently turning "unknown name" into `Some(Mixed)` via the `local_type` fallback while the checker's `env.get` correctly said `None`.

Regression tests added for both directions (the detach shape compiles and runs; the literal-through-variable heterogeneous push still widens). Full sweep green **including the spl suite** this time (1803 codegen + 993 error tests).

nahime0:
Follow-up hardening after reviewing the version currently published on this PR (`c1d71ae3`).

The published CI fix correctly stopped treating every loop-local, non-literal value as automatic `mixed` evidence, which fixed the false widening in `MultipleIterator::detachIterator`. However, that version still had a few correctness gaps:

- two differently typed, non-literal values were both considered opaque, so calls such as `getInt(): int` followed by `getFloat(): float` could still leave the first append lowered as a raw write and crash on the second loop iteration;
- a local with a type at loop entry could be reassigned inside the loop from a typed non-literal expression, but the stale entry type could still win;
- the prescan covered `$a[] = ...`, but not the equivalent growth/write surfaces `$a[$i] = ...` and `array_push($a, ...)`.

The local follow-up commits (`e41cb5624` and `e8bf7513d`, to be included in the next branch push) address those gaps:

- the shared scan now covers `$a[] =`, indexed assignments, and `array_push()`, including nested expression sites;
- assignment RHS expressions are preserved and resolved through caller-provided semantic type inference instead of being limited to literals;
- the checker infers values against the loop-entry environment snapshot, while EIR pre-lowering resolves the types it already knows for direct functions, instance/static methods, properties, array-element reads, locals, objects, and ternaries;
- an opaque in-loop reassignment now poisons stale loop-entry evidence instead of silently reusing it;
- opaque + concrete sibling evidence still widens conservatively, while an unknown value alone still does not force `mixed`.

This keeps the original `MultipleIterator` protection intact: its property/array read is now resolved as `Iterator`, so the homogeneous `array<Iterator>` rebuild remains typed instead of being widened spuriously.

Focused verification on the final local state:

- warning-free `cargo build`;
- 13 `loop_grown_mixed_array` regressions passing;
- the same 13 cases passing with `ELEPHC_PHP_CHECK=1`;
- `test_loop_rebuild_of_typed_array_not_spuriously_widened` passing;
- `git diff --check` clean.

The complete macOS ARM64 / Linux ARM64 / Linux x86_64 matrix remains for CI after the branch is pushed.

Issue #488, opened 2026-07-07
match with heterogeneous mixed arms (object/array/string) dies at runtime coercing the object arm to string


mirchaemanuel:
## Summary

A `match` whose arms return heterogeneous `mixed` values (object / array / string) compiles but dies at runtime — the match-arm type unification appears to coerce the object arm to string.

## Verified on

`main` (`7396e65b`), macOS ARM64.

## Minimal reproduction

```php
<?php
function pick(int $n): mixed {
    return match($n % 3) { 0 => new stdClass(), 1 => [1, 2], default => "s" };
}
$c = 0;
for ($n = 0; $n < 6; $n++) {
    $v = pick($n);
    $c += is_object($v) ? 10 : (is_array($v) ? 7 : 3);
}
echo $c;
```

## Actual

Runtime fatal (`Object of class stdClass could not be converted to string`), no output.

## Expected (PHP)

```
40
```

## Notes

Found while stress-testing mixed-value shapes for the #484 review. The arms individually work; it is the heterogeneous unification of the match result that mis-coerces.

mirchaemanuel:
Fix up in #492.

Root cause was exactly the suspected unification: `lower_match` typed its hidden merge temp with the syntactic join (`wider_type_syntactic`), which is scalar-biased — any `Str` arm collapses the whole join to `Str`, and it can never produce `Mixed`. Every arm was then coerced into that one type on store: the object arm through a fatal string cast (this report), the array arm through a *silent* one. The pairwise matrix was consistent: `Str` absorbs everything, `null` is absorbed, otherwise the last-folded arm's concrete type wins (`object+int` even failed to *compile*). The checker had an independent bug on the same construct — `ExprKind::Match` inference returned the **first arm's type** and ignored the rest, which is why the inferred-return variant died with `runtime_call from PHP type Str to PHP type Object`.

The fix mirrors the ternary lowering, which already gets this right (`branch_merge_result_type` → merge to `Mixed`, box each branch as-is): match arms now fold through the same merge machinery, and the checker joins all arms. Full before/after matrix in the PR.

Characterizing this also surfaced a pre-existing crash the fix would otherwise have newly routed match values into: `gettype()` on a nullable-int merge temp (`$v = ($argc == 1) ? 1 : null; echo gettype($v);` — segfaults on current `main` via ternary). Fixed in the same PR, with the wider emitter-dispatch pattern filed as #493 and the inferred-return nullability erasure as #494.

Issue #487, opened 2026-07-07
Compound assignment with a closure/indirect call RHS accumulates only the first iteration's value


mirchaemanuel:
## Summary

`Closure::bind()` with a fresh receiver inside a loop miscompiles: only the first iteration's call returns the receiver's value; the accumulated result is wrong.

## Verified on

`main` (`7396e65b`), macOS ARM64. No heap crash — a pure semantic divergence.

## Minimal reproduction

```php
<?php
class P { public int $v = 4; }
$acc = 0;
for ($n = 0; $n < 50; $n++) {
    $f = Closure::bind(function (): int { return $this->v; }, new P());
    $acc += $f();
}
echo $acc;
```

## Actual

```
4
```

## Expected (PHP)

```
200
```

The single-shot shape (no loop) returns 4 correctly, so the first iteration works and subsequent rebinds return 0 (or the loop's later calls miss the bound receiver).

## Notes

Found while extending the #484 regression coverage to the closure-capture boxing site. Unrelated to the leak itself — the miscompile reproduces identically before and after the #484 fix.

mirchaemanuel:
Sharper diagnosis while extending regression coverage: the root cause is **not** `Closure::bind` — it's compound assignment with a closure/indirect call on the RHS.

```php
$f = function (): int { return 4; };
$acc = 0;
for ($n = 0; $n < 20; $n++) { $acc += $f(); }
echo $acc;   // elephc: 4 — PHP: 80
```

Controls (both correct): `$v = $f(); $acc += $v;` via intermediate variable → 80; `$acc += g();` with an ordinary named function → 80. The original `Closure::bind` repro in the description is a symptom of this same defect (the adversarial matrix also showed `$acc += $f()` failing for `use`-capturing closures and callable-string invokers: `4 4` vs `80 4`, `4` vs `150`). Title updated accordingly.

Issue #486, opened 2026-07-07
Owned temporaries passed as mixed arguments are never released by the caller


mirchaemanuel:
## Summary

An owned temporary passed as a `mixed` argument to an ordinary function is never released by the caller: `probe(new P())` / `probe([$n, $n])` in a loop leaks one payload per call (2 blocks/iteration in the repro). The value is boxed for the `mixed` parameter, but the producer's reference survives the call.

## Verified on

`main` (`7396e65b`), macOS ARM64, `--heap-debug`. Output matches PHP (150); heap grows linearly.

## Minimal reproduction

```php
<?php
class P { public int $v = 3; }
function probe(mixed $m): int { return is_object($m) ? 1 : 2; }
$acc = 0;
for ($n = 0; $n < 50; $n++) {
    $acc += probe(new P());
    $acc += probe([$n, $n]);
}
echo $acc;   // 150, correct
```

```
HEAP DEBUG: leak summary: live_blocks=100 live_bytes=5600
```

## Notes

Same ownership family as #483 (expression-position temporaries) and #484 (mixed_box producer reference, fixed for return/coercion sites) but a distinct lowering path: the plain-function `mixed`-parameter argument. Found while extending #484's regression coverage — this shape does not flow through the converted boxing sites.

mirchaemanuel:
Mechanism detail from the #484 review: this shape never emits `Op::MixedBox` in EIR at all — the caller passes the raw `Heap(Object)` and the boxing happens in **codegen call materialization** (`__rt_mixed_from_value` emitted around `src/codegen_ir/lower_inst.rs:~3209`). The assembly confirms the sequence: box → call → `__rt_decref_mixed` on the **cell** only; the payload's producer reference (the `object_new` result held in its frame slot) is never released — one payload per call.

That's why the #484 fix (which releases the producer at the EIR `MixedBox` sites) doesn't cover it: same bug mechanism, different layer. A fix here likely belongs in the call-argument materialization path (release the owned argument temp after the call, mirroring `box_value_as_mixed`'s rule), or by lowering the arg coercion through EIR `MixedBox` so the shared helper applies.

mirchaemanuel:
## Resolved on current `main`

Re-verified today on `main` `ba836179c`, macOS ARM64, debug build, compiled with `--heap-debug` and run under both `--ir-opt=on` and `--ir-opt=off`. Reference: PHP 8.4.23.

[...]

Issue #485, opened 2026-07-07
?string return of a runtime-built string leaks the box's persisted copy once per call


mirchaemanuel:
## Summary

A `?string`-returning function whose value is built at runtime leaks one ~48-byte block per call: boxing the string into the Mixed cell persists a copy for the box (`__rt_mixed_from_value_string` → `__rt_str_persist`), and that persisted copy is never freed when the box dies.

## Verified on

`main` (`7396e65b`), macOS ARM64, `--heap-debug`. Reproduces identically with and without the #484 producer-release fix (the producer here is a concat-scratch string, correctly exempt from heap release), so this is a distinct mechanism from #484.

## Minimal reproduction

```php
<?php
function s(int $n): ?string { return str_repeat("x", 3 + ($n % 2)); }
$len = 0;
for ($n = 0; $n < 50; $n++) { $v = s($n); $len += strlen($v); }
echo $len;   // 175, correct
```

```
HEAP DEBUG: leak summary: live_blocks=50 live_bytes=2400
```

Exactly one 48-byte block per iteration. The caller side is balanced (store/rebind releases the box each iteration), and the callee's IR is balanced for its scratch temporaries — the unfreed block is the box's own persisted string payload.

## Expected

`__rt_decref_mixed` → `__rt_mixed_free_deep` on a tag-1 (string) cell should free the persisted payload copy along with the cell. Either that path is not releasing tag-1 payloads, or the persist stores a pointer the free path refuses; worth tracing `mixed_free_deep`'s tag-1 arm against `__rt_str_persist`'s block layout.

## Notes

Found while fixing #484 (owned-object boxing). The object and array flavours are covered there; strings go through the persist path instead of the retain path, which is where this leak lives.

mirchaemanuel:
Taking this one. Last re-verified during the 16 Jul triage on main `84095e0ee` (one persisted copy leaked per call); will re-confirm on current main and follow up with the diagnosis and a PR, same workflow as #405/#529.

mirchaemanuel:
Diagnosed — and the original hypothesis in this issue (mine) was wrong, so correcting it for future readers: the box's free path is actually fine. `__rt_mixed_free_deep`'s tag-1 arm frees the persisted copy correctly, and the same `?string` return consumed by discard/`echo`/`strtoupper`/concat is fully heap-clean.

[...]

PR #489, opened 2026-07-07
fix(ir): release the producer's reference when boxing an owned value as Mixed (#484)

### LINKED ISSUE #484: mixed_box of an owned object leaks the producer's reference (?Class returns leak one object per call)
## Summary

Boxing an **owned** object into a Mixed cell leaks the producer's reference: `__rt_mixed_from_value` retains object payloads (tag 6 → `__rt_incref`), and `__rt_mixed_free_deep` correctly releases the child when the cell dies — but EIR lowering treats `Op::MixedBox` as *consuming* its operand and never releases the producer's own reference. Net +1 on the payload per boxing of an owning temporary.

The everyday shape is a nullable-object return:

```php
class P { public int $v = 7; }
function g(): ?P { return new P(); }   // return coerces to Mixed: object_new (rc=1) → mixed_box (payload rc→2)

$acc = 0;
for ($n = 0; $n < 50; $n++) { $c = g(); $acc += $c->v; }
echo $acc;   // 350, correct
```

## Verified on

`main` (`7396e65b`), macOS ARM64, `--heap-debug`: **50 leaked blocks in 50 iterations — exactly the `P` objects.** The boxed cells themselves are freed correctly on rebind (`__rt_decref_mixed` → `__rt_mixed_free_deep` → `decref_any(child)` drops the child from 2 to 1 — never to 0).

IR of the return path:

```
v5: Heap(Object) php=P own=maybe_owned = object_new …   ; producer reference, rc=1
v6: Heap(Mixed) = mixed_box v5                           ; runtime retains payload, rc→2
return v6                                                ; v5's reference never released
```

## Expected

One reference per owner: after boxing, the cell owns the payload; an owning-temporary operand's producer reference must be released (a borrowed operand — e.g. boxing a loaded local — is correct as-is: the retain gives the box its own reference alongside the local's).

## Suggested direction

In the `Op::MixedBox` lowering/emission path(s) in `ir_lower`, release the operand after boxing when `value_is_owning_temporary(operand)` — same ownership pattern as the recent throw-of-borrowed fix in #477. Needs an audit of all `MixedBox` emission sites (return coercion, argument coercion, storage coercion) so the rule is applied uniformly.

## Notes

Surfaced while verifying #447 (whose original stored-`tryFrom` shape is fixed on current `main`). Enum `tryFrom` itself no longer exhibits this class (its payload is a persistent singleton); any `?Class`/`mixed`-returning function that produces fresh objects does.

mirchaemanuel:
## Summary

Fixes #484. `__rt_mixed_from_value` retains refcounted payloads (objects, arrays, hashes, callables, nested cells) and persists strings, so a boxed Mixed cell always carries its own reference or copy — but EIR lowering treated `Op::MixedBox` as consuming its operand and never released the producer's reference. Boxing an **owning temporary** therefore left the payload one reference high forever:

```php
function g(): ?P { return new P(); }   // leaked one P per call
```

The boxed cell itself was freed correctly on rebind; only the payload leaked (fresh arrays boxed the same way leaked identically).

## Fix

A new `LoweringContext::box_value_as_mixed` helper emits the `MixedBox` and then releases the operand when it is an owning temporary — the same ownership rule as the throw-of-borrowed fix in #477. The 11 value-carrying emission sites now route through it; the two remaining direct sites box freshly-created nulls (non-refcounted, commented accordingly).

Deliberately unchanged:
- **Borrowed operands** (boxing a loaded local): no release — the box's retain is their +1. Verified use-after-box stays valid.
- **Concat-scratch strings**: the added release is a verified no-op on scratch strings, so string boxing behaviour is byte-identical; the separate persisted-copy leak is #485.

## Verification

- Regression tests (object + array flavours) in `tests/codegen/runtime_gc/regressions.rs`; broad suites green (1872 codegen across callables/closures/runtime_gc/control/types/exceptions/arrays/oop/generators + 993 error tests).
- Adversarially reviewed across ~28 fixtures: no double free, no use-after-release (all converted sites' callers shadow the operand — audited one by one), IR shows exactly one release immediately after the box for owned producers and none for borrows. Closure-capture and invoker-hash shapes improve as a side effect.

## Found during review (pre-existing, filed separately)

- #486 — owned temporaries passed as `mixed` **arguments** still leak: that boxing happens in codegen call materialization, not EIR, so this fix doesn't reach it (mechanism detailed on the issue).
- #487 — compound assignment with a closure/indirect call RHS accumulates only the first iteration's value (retitled from the `Closure::bind` symptom after isolating the root).
- #488 — `match` with heterogeneous `mixed` arms dies at runtime coercing the object arm to string.
- #483 gained the prop-get-through-Mixed shape (5 blocks/iteration) as further evidence.


mirchaemanuel:
Rebased on current main as requested — ready to go. ✅

Conflict resolution notes:
- `src/ir_lower/stmt/mod.rs`: the foreach initializer now combines both sides — boxing goes through the new `box_value_as_mixed` helper (this PR) while the store uses `store_foreach_initializer_local_only` introduced by the #534 fix on main.
- `tests/codegen/runtime_gc/regressions.rs` and `CHANGELOG.md`: kept both sides' additions.

Verified locally: full `runtime_gc` suite green (155/155), including the two #484 regression tests.

Issue #378, opened 2026-06-16
x86_64 runtime routines clobber callee-saved r14/r15 without ABI-compliant save/restore


nahime0:
## Description

On x86_64, several hand-written runtime routines (`__rt_*`) and the shared
heap-marker / array-store codegen use the SysV **callee-saved** registers `r14`
and `r15` as scratch **without** saving and restoring them. Per the SysV ABI a
callee must preserve `rbx`, `rbp`, `r12`–`r15`, so this is an ABI-compliance
defect.

It was latent and harmless until now because the previous codegen never kept a
live compiled value in `r14`/`r15` across a call. The linear-scan register
allocator (added in the EIR backend) exposed it: when it placed a live value in
`r14`/`r15` across a call into one of these routines, the value was corrupted —
observed as `Fatal error: heap memory exhausted` in spread/variadic tests
(`test_spread_mixed_with_elements`) because a corrupted length/count drove a
runaway allocation.

## Current status (worked around)

To ship the register allocator safely, its x86_64 integer pool was restricted to
**`rbx` only**; `r14`/`r15` are excluded. This is documented in
`src/ir_passes/regalloc.rs` and `docs/internals/the-ir.md`. The full codegen
suite passes on Linux x86_64, Linux aarch64, and macOS aarch64.

`rbx` is reliably preserved by the runtime, so the workaround is correct — but it
leaves register-allocation performance on the table for x86_64 (only one
allocatable integer register vs. eight on aarch64).

## Proposed fix

Audit the x86_64 runtime routines and shared heap-marker/array-store emitters
that use `r14`/`r15`, and make them ABI-compliant (push/pop the callee-saved
registers they clobber, or switch to caller-saved scratch such as `r10`/`r11`).
Once `r14`/`r15` are guaranteed preserved across all `__rt_*` calls, add them
back to the allocator pool in `int_pool()` (`src/ir_passes/regalloc.rs`).

Suspected starting points (callee-saved scratch without visible save/restore):
- `src/codegen/stmt/arrays/assign/indexed/normalize.rs` and
  `src/codegen/stmt/assignments/properties/arrays/indexed.rs` (`mov r14, r12`
  heap-marker manipulation)
- various `src/codegen/runtime/**` routines using `r14`/`r15` as working registers

## Impact

- No user-visible incorrect behavior today (worked around).
- Recovering `r14`/`r15` would improve x86_64 throughput of the register
[...]

PR #490, opened 2026-07-07
fix(codegen): callable-invoker trampoline preserves callee-saved registers (#487)

### LINKED ISSUE #487: Compound assignment with a closure/indirect call RHS accumulates only the first iteration's value

mirchaemanuel:
## Summary

Fixes #487. Compound assignment with an indirect callable RHS silently miscompiled: `$acc += $f()` in a loop accumulated only the last call's value (`4` instead of `80`), with the closure demonstrably running every iteration. Every indirect form was affected — closures, arrow functions, callable strings, first-class callables, method FCCs — with any operator, compound or not, including property targets.

## Root cause

An ABI violation in the generated **descriptor-invoker trampoline** (`src/codegen/runtime_callable_invoker.rs`): it saved only fp/lr and then used callee-saved registers as raw scratch (AArch64 `x19–x26` via the nested-call register and the argument loaders; x86_64 `r12/rbx/r13–r15`). The linear register allocator relies on the platform ABI and parks values that live across a `callable_descriptor_invoke` in exactly those registers — so the loaded LHS of `$acc += $f()` was clobbered by the invoke and read back as the trampoline's leftover scratch (the empty args-array length, `0`). Each iteration computed `0 + f()`.

Diagnosed via IR (byte-identical with the optimizer off — lowering and effects are correct) and assembly (the allocator parks `v6` in `x21`; the trampoline's last write to `x21` is the args length). `--regalloc=stack` masked the bug; direct calls were safe because compiled functions save allocator-used callee-saved registers in their own prologues — the invoker trampoline was the one generated code path missing that treatment. Same defect class as #378 on the runtime-routine side.

## Fix

The invoker frame grows to hold a save area, and the prologue/epilogue now store/reload the callee-saved set the body scratches — mirroring compiled-function prologues on both targets.

## Verification

- 3 regression tests in `tests/codegen/callables/expr_calls.rs` (closure compound assign; callable-string/FCC with `-=`/`*=`; non-compound + property target).
- Broad suites green: 2231 codegen (callables/closures/runtime_gc/control/types/optimizer/arrays/oop/spl) + 993 error tests.
- Adversarially reviewed: 30/30 fixtures match PHP (10-arg spill, nested indirect calls, re-entrant trampoline, spread, register pressure, floats — the allocator's float pool d8–d14 is verified never scratched by the invoker); **92/92 emitted invokers audited in asm** (8 saves + 8 reloads + single `ret`, no slot collisions, return registers untouched).
- **Exception path is safe by design**: unwinding is `setjmp`/`longjmp`, which restores callee-saved registers from the try-entry register file, so the skipped trampoline epilogue is irrelevant — verified behaviorally across throw-through-two-trampolines, multiple live values, and partial-loop throw/catch shapes.
- x86_64: static audit complete (the saved set is the full SysV callee-saved set besides rbp; 16-byte alignment preserved); execution coverage delegated to CI's linux-x86_64 matrix.


nahime0:
## Follow-up fixes after review

Compared with the currently published PR head (`060ff694`), the original fix that saves/restores the GPR callee-saved registers in the callable-invoker trampoline remains unchanged.

The review found one additional ABI gap: AArch64 overflow-float staging used `d15`, which is callee-saved under AAPCS64. This was not limited to the invoker's outgoing call path; the same scratch choice was duplicated in shared incoming/outgoing ABI handling and in the AOT/eval descriptor bridges. A caller keeping a live value in `d15` could therefore still be corrupted when more than eight float arguments were involved, even though the current register allocator only assigns `d8-d14`.

The following follow-up changes have been prepared on top of the published PR:

- Added one shared `float_spill_scratch_reg()` ABI helper: `d31` on AArch64 (caller-saved and outside the allocator pool), `xmm15` on x86_64.
- Replaced the direct `d15` use in all six float-spill staging paths: shared incoming parameters, shared outgoing arguments, AOT descriptor entry store/load, and eval descriptor entry store/load.
- Added two structural ABI regression tests covering incoming and outgoing float overflow.
- Added an end-to-end indirect-closure regression with nine float arguments. It verifies `d31` is emitted and `d15` is absent on AArch64, verifies `xmm15` on x86_64, assembles/runs the program, and expects `45`.
- Aligned the six inherited assembly comments in `runtime_callable_invoker.rs` that were failing the repository checker.
- Integrated the branch with the current `origin/main` before applying the follow-up.

Local commits pending push:

- `db2174f4f fix(abi): preserve callee-saved float registers`
- `c96292157 style(codegen): align callable invoker comments`

Focused verification completed successfully:

- `cargo build`
- 2 ABI unit regressions
- the 3 compound-assignment matches plus the live-across-closure regression from this PR
- the new end-to-end float-overflow regression on macOS AArch64
- the same regression through `./scripts/test-linux-x86_64.sh`
- `cargo fmt --all -- --check`, `git diff --check`, and the assembly-comment checker

Only the two pre-existing `libc::time_t` deprecation warnings appeared in the isolated Linux build. Linux AArch64 and the complete supported-target matrix are left to CI after the branch is pushed.

PR #491, opened 2026-07-07
fix(codegen): stamp the canonical x86_64 heap magic through one shared helper (#482)

### LINKED ISSUE #482: x86_64 emitters stamp a transposed heap magic (0x4548504c vs 0x454C5048): refcount ops silently no-op
## Summary

Several x86_64 emitters stamp heap allocations with a **transposed** magic word, `0x4548504c` (`E H P L`), while the canonical constant checked by the refcount helpers is `X86_64_HEAP_MAGIC_HI32 = 0x454C5048` (`E L P H`, see `src/codegen/runtime/arrays/incref.rs:14`, `decref_object.rs:15`). Any object stamped with the transposed word fails the magic check in `__rt_incref` / `__rt_decref_object` / deep-free, so **every refcount operation on it is a silent no-op on linux-x86_64** — the object can never be freed, and acquire/release bookkeeping is inert for it.

## Affected stamp sites (grep `0x4548504c`)

```
src/codegen_ir/lower_inst/objects.rs            (2: LogicException + Error throwable allocation)
src/codegen_ir/lower_inst/enums.rs              (enum ValueError/TypeError throwers)
src/codegen_ir/lower_inst/static_properties.rs
src/codegen_ir/lower_inst/builtins/math.rs
src/codegen/runtime/arrays/value_error.rs
src/codegen/runtime/system/json_throw_error.rs
src/codegen/runtime/spl/doubly_linked_list.rs
src/codegen/runtime/spl/fixed_array.rs
src/codegen/builtins/math/clamp.rs
src/codegen/builtins/strings/hash_crypto.rs
src/codegen/expr/objects/allocation.rs          (frozen legacy backend)
src/codegen/expr/objects/dispatch/enums.rs      (frozen legacy backend)
```

By contrast, the main throwable allocation used by user-level `new` (`src/codegen_ir/lower_inst/objects.rs`, `emit_throwable_allocation`) and `__rt_heap_alloc` itself stamp the canonical `0x454C5048…`.

## Impact

On linux-x86_64 only: throwables and objects created through the affected paths (runtime-raised `ValueError`/`TypeError`/`LogicException`/`Error`, JSON throw errors, some SPL/math/static-property paths) leak once released paths exist for them, and any ownership balancing around them silently does nothing. This was latent while caught exceptions were never released at all; PR #477 (issue #448) makes exception releases real, so these objects now diverge from the correctly-stamped ones (they leak instead of freeing — no crash).

Full disclosure: the `enums.rs` occurrences include code I added in #450/#453, faithfully copying the transposed literal from the pre-existing `emit_throw_value_error_from_string_result_x86_64` — the constant deserves a single named `const` shared by every emitter so this class of typo cannot recur.

## Suggested direction

Replace every literal `0x4548504c…` stamp with the canonical constant (ideally exporting one `X86_64_HEAP_MAGIC_HI32`-derived kind-word helper for emitters), and add a debug assertion or test that greps/validates stamped magic words against the checker constant. Legacy-backend files can be skipped per the backend freeze policy.

mirchaemanuel:
## Summary

Fixes #482. Twelve emitters hand-typed a **transposed** x86_64 heap magic (`0x4548504c` instead of the canonical `0x454C5048`, ASCII `ELPH`) in their kind-6 throwable stamps: the runtime-raised `ValueError`/`TypeError`/`LogicException`/`Error` paths, JSON throw errors, SPL fixed-array / doubly-linked-list exceptions, the hash builtin error path, and static-property/math error objects. The x86_64 refcount and free helpers gate on the canonical magic, so **every refcount operation on those objects was a silent no-op on linux-x86_64**.

## Fix

- One shared, documented helper next to the sentinel constants: `codegen_support::sentinels::x86_64_heap_kind_word(kind)` — all 12 stamps now go through it (the emitted assembly changes only in the magic value).
- The constant's layout is **unit-tested**, and a **repository lint test** scans `src/` so a hand-typed transposed literal can never come back — the class of typo is closed, not just the instances (full disclosure: two of the twelve were mine, faithfully copied from the pre-existing ValueError emitter, which is exactly why a shared constant was the right fix).

## Behavioral note

Neutral **today**: caught-exception releases land with the #448 fix in PR #477, and until then both increfs and decrefs on these objects were no-ops alike. Once #477 merges, throwables from these paths become genuinely freeable on x86_64, matching ARM64 (whose release path has no magic/kind gate). The two changes are independent and safe to merge in either order.

## Verification

Suites green: 1614 codegen (enums/exceptions/spl/json/oop/types/static — every touched emitter's surface) + 993 error tests; sentinels unit + lint tests included; asm comment alignment verified on all rewritten lines.


nahime0:
Follow-up after rebasing/auditing the currently published PR head (`d5a5f6c1`).

Compared with the single commit currently visible on this PR, the follow-up prepared locally for the next push adds the following:

- Synced the branch with current `main` and resolved the merge conflicts.
- Found and fixed **two additional transposed kind-6 heap stamps** in the current exception lowering, bringing the corrected total from 12 to 14.
- Broadened the original targeted helper usage into a repository-wide invariant: **every x86_64 heap-header stamp and every heap-magic check** now goes through `codegen_support::sentinels`, rather than only the originally affected throwable emitters.
- Generalized `x86_64_heap_kind_word(low_bits)` to preserve the complete low 32-bit packed field, including plain kinds, COW/value-type encodings, and magic-only initialization.
- Removed the stale duplicate heap-marker materialization in the sockaddr formatting path while converting it to the shared helper.
- Strengthened the source lint so it catches underscored/case-varied magic literals, rejects local copies of the canonical constant, and rejects hand-typed canonical immediates outside the sentinel module.
- Re-audited and aligned every changed `emitter.instruction(...)` inline comment.

Focused verification completed locally:

- `cargo build`
- sentinel unit/lint tests: **5/5 passed**
- `git diff --check`
- diff-aware assembly-comment audit
- source scan confirms no transposed heap-magic literal remains under `src/`

These follow-up commits are not on GitHub yet; the PR head will include them after the next push. The full supported-target matrix is intentionally left to CI.

PR #492, opened 2026-07-07
fix: match with heterogeneous arm types no longer coerces every arm to one unified type

### LINKED ISSUE #488: match with heterogeneous mixed arms (object/array/string) dies at runtime coercing the object arm to string

mirchaemanuel:
Fixes #488.

## Root cause

`lower_match` typed its hidden merge temp with the **syntactic** expression-type join (`fallback_expr_type` → `infer_expr_type_syntactic`), whose `wider_type_syntactic` fold is scalar-biased: any `Str` operand collapses the whole join to `Str` (and it can never produce `Mixed`). With arms `{object, array, string}` the temp became `Str`, so `store_expr_into_temp` → `coerce_value_for_temp` emitted an `Op::Cast` to string for the object arm — the runtime fatal from the issue — and a *silent* string coercion for the array arm. The full pairwise matrix from the issue reproduced exactly (Str absorbs everything; otherwise the last-folded arm's concrete type wins).

The type checker's `ExprKind::Match` inference had an independent bug in the same construct: it returned the **first arm's type** and ignored every other arm, which is why the heterogeneous match with an *inferred* return type failed to compile (`runtime_call from PHP type Str to PHP type Object`) — two different wrong unifications colliding.

The ternary lowering already solves this correctly via `branch_merge_result_type` → `wider_type_for_merge`, which merges heterogeneous heap types to `Mixed` so each branch value is boxed as-is (`Op::MixedBox`) instead of cast. This PR brings `match` to that exact model.

## Changes

- **`src/ir_lower/expr/mod.rs`** — new `match_merge_result_type()` sitting next to `branch_merge_result_type()`: folds `materialized_expr_type_for_merge` over all arms + default with `nullable_aware_branch_merge_type`, keeps nullable merges, and reconciles non-nullable merges with the syntactic fallback via `wider_type_for_merge` (which can only keep or widen — the Str-absorbing fallback can no longer narrow the temp). `lower_match` now uses it for the hidden temp. `materialized_expr_type_for_merge` also gains an explicit `ExprKind::Match` arm (mirroring the existing `Ternary` arm) so a match nested inside a ternary/coalesce merge uses the same rule instead of falling into the syntactic catch-all.
- **`src/types/checker/inference/expr/mod.rs`** — `ExprKind::Match` inference now folds **all** arm types with `merge_match_arm_result_type()`: identical arms keep their type, `Never`/`Void` arms (`throw` / `null`) defer to the other arms, any other heterogeneous pair widens to `Mixed`. This keeps the checker and EIR lowering in agreement for inferred return types and assignments.
- **`src/codegen/lower_inst/builtins.rs`** — `lower_gettype` now dispatches on `ty.codegen_repr()` instead of the raw PHP type. A nullable-int union (`null|int`) stores an **inline tagged scalar**, but the raw-type dispatch matched `Union(_)` and routed it to the boxed-Mixed path, calling `__rt_mixed_unbox` on a non-pointer payload → segfault. This was a pre-existing hole (the ternary twin `$v = ($argc == 1) ? 1 : null; echo gettype($v);` segfaults on current `main`), but the match fix newly routed top-level `match` int/null results into it, so it is fixed here with regression tests for both constructs. Verified under both `--null-repr=sentinel` and `--null-repr=tagged`.
- **`tests/codegen/control_flow/match_expressions.rs`** (new) — 13 tests: the issue repro, every pairwise arm mix from the characterization matrix (object/string both orders, object/array, object/int, array/int, string/int, object/null), top-level and inferred-return-type shapes, homogeneous-arm and throw-default no-regression guards, a heap-debug balance test, and the gettype tagged-scalar regression. Plus `test_ternary_int_null_gettype` in `ternary.rs` for the pre-existing ternary crash.
- **`docs/php/control-structures.md`** — documents heterogeneous arm behavior (boxed `mixed` result, per-arm runtime types, same-representation merges).

## Behavior after the fix

| Arms | before | after | PHP |
|---|---|---|---|
| object / string | **runtime fatal** | `object\|string` | `object\|string` |
| object / array | `array\|array` (silent) | `object\|array` | `object\|array` |
| object / int | **compile error** | `object\|integer` | `object\|integer` |
| array / int | `integer\|integer` (silent) | `array\|integer` | `array\|integer` |
| string / int | `string\|string` (silent) | `string\|integer` | `string\|integer` |
| object / null | `object\|object` (silent) | `object\|NULL` | `object\|NULL` |
| heterogeneous, inferred return | **compile error** | works | works |
| int / null + `gettype` | `integer` / **segfault** (ternary) | `integer\|NULL` | `integer\|NULL` |

## Validation

- 13 new match tests + 1 ternary regression: green on macOS ARM64 and Linux x86_64 (Docker, `control_flow` suite).
- Focused regression filters all green: `match` (114), `ternary` (36), `enum` (46), `coalesce` (52), `gettype` (16), `error_tests match` (21).
- An adversarial review pass probed nesting (match-in-ternary, ternary-in-match, match-in-match), constant-folded subjects, `--ir-opt=on|off` (byte-identical outputs), declared `?stdClass` returns, throw arms inside try/catch, and heap balance — no new leaks and no checker/lowering disagreement found.

## Adjacent pre-existing findings (out of scope, kept out of this PR)

- Boxed **object/array arm** results leak the producer reference — that is exactly #484 (`mixed_box` producer reference), pending in #489; the behavior here is byte-for-byte identical to the ternary twin.
- The raw-type `Mixed | Union(_)` dispatch pattern that caused the gettype crash appears in a dozen other builtin emitters in `src/codegen/lower_inst/builtins.rs`; only the gettype instance (the one this fix routes values into) is fixed here. Filed separately with the site list.
- `null`-typed arms at *inferred-return* boundaries still erase nullability at the checker level (`object|NULL` prints `object|object` with no declared return type) — identical for ternary and unchanged by this PR; filed separately.


mirchaemanuel:
Pushed three follow-up commits after a multi-agent review of this branch:

- **57f6667bd docs: regenerate builtins catalog** — fixes the failing `Builtins docs in sync` gate: the 3-line comment added in `lower_gettype` shifted every downstream lowering function's line anchor by +3, so the 24 generated pages needed regeneration (pure line-number changes, no semantic drift).
- **a32c2b640 fix: match null arms keep the checker's inferred merge nullable** — `match() { 0 => null, default => "s" }` returned through an *inferred* return type printed `string|string` (PHP: `NULL|string`): the checker's arm merge typed both `null` and `throw` as `Void` and dropped them. Throw arms now normalize to `Never` at the merge call site; genuine null arms widen the merge to a nullable union, mirroring the lowered temp's nullable-aware merge. Adds tests for that plus value-consumption, float/int arms, and nested heterogeneous match.
- **ae0cac4d9 docs: document match arm representation merges as known incompatibilities** — int/bool arms and indexed-array arms with different element types still merge to one representation (PHP keeps per-arm types); moved that out of the "matching PHP" sentence into the known-incompatibilities list, with a pinning test.

Two confirmed pre-existing items intentionally **not** addressed here (both reproduce at merge-base too, via the shared boxing/merge paths):
1. object-into-`Mixed` boxing leaks the payload + box on every evaluation (`--heap-debug` shows `live_blocks` growing per iteration for object arms; same leak shape for ternary object branches and `: mixed` object returns);
2. two indexed-array arms with different element types miscompile the read (`match(){1 => [1,2], default => ["a","b"]}` reads the int arm through `Array<string>`) — now documented as a known incompatibility; merging those arms to `Mixed` would fix it.

Happy to open follow-up issues for either if useful.

nahime0:
Follow-up to the update above, relative to the currently published head (`ae0cac4d`). These changes are ready locally and will appear with the next push:

- Merged the current `main` and reconciled its new assignment-effects inference path. Heterogeneous `match` and full ternary expressions now retain the same Mixed-aware merge type when assigned to a local and returned through an inferred return type.
- Completed the abbreviated ternary (`?:`) path. The checker already inferred a heterogeneous result as `Mixed`, but EIR lowering still used the old scalar-biased syntactic fallback, so a truthy integer such as `7 ?: "fallback"` could be coerced to a string. Lowering now uses the same materialized merge rule and preserves the PHP runtime type.
- Fixed the inferred-return nullability gap tracked by #494 and explicitly called out as out of scope in the previous comment. Object/null results now preserve `NULL` for both `match` and ternary expressions, in direct-return and assign-then-return shapes.
- Added focused regressions for all of the above and updated the changelog for #488 and #494.
- Regenerated the builtin registry plus 23 internal builtin pages after the `gettype()` comment shifted their codegen line anchors. A second regeneration produces no diff.

Local follow-up commits:

- `9c61c6251` — `merge: sync main and fix match/ternary assignment-effects merge`
- `d1a0bc5a4` — `docs(changelog): note heterogeneous match/ternary Mixed merge (#488)`
- `3fc6ab7a6` — `fix: align heterogeneous ternary merge lowering`
- `59fdb4ad0` — `docs: regenerate builtin lowering references`

Local validation:

- `cargo build`
- 24 focused match tests and 42 focused ternary tests
- PHP cross-checks for the new regressions
- object/null cases with sentinel null representation
- new regressions with IR optimization disabled
- builtin docs audit: 0 errors; 913 pages site-compatible
- generated-docs idempotence check
- `git diff --check`

The worktree is clean and ready for the upcoming push and CI run.

PR #518, opened 2026-07-09
feat(types): fluent wither-return covariance through descendant interfaces

chadmandoo:
## What

A PSR-FIG immutable "wither" method (`withHeader(): MessageInterface`) declares an **ancestor** interface as its return type but carries `@return static`. Called through a **descendant**-interface receiver, the result must keep the receiver's type so a subsequent descendant-only call type-checks — matching PHP's runtime `clone $this` and PHPStan's `@return static`.

Previously rejected:

```php
interface Message { public function withHeader(string $v): Message; }
interface Request extends Message {
    public function withMethod(string $m): Request;
    public function method(): string;
}
function chain(Request $r): string {
    // withMethod() was unresolved: withHeader() typed as the ancestor Message
    return $r->withHeader('x')->withMethod('POST')->method();
}
```

## How

Both inference passes resolve a `with*` method whose declared return is a **strict ancestor** of the receiver interface back to the receiver:

- checker — `infer_method_call_on_interface_type`
- EIR lowering — `method_call_result_type`

Both are required: the checker fix alone leaves the EIR backend re-resolving the chained call to the ancestor and failing native codegen (`interface method call to unknown method Message::withMethod`).

Also accepts `Array(Mixed)` where an `AssocArray` parameter is declared (an untyped list literal passed to an `array<int, mixed>` param).

## Test

`test_interface_wither_ancestor_return_stays_receiver` — a `Message` / `Request extends Message` wither chain, byte-parity vs PHP 8.5 (`chain(new Req())` prints `POST`).

### DIFF
--- a/src/ir_lower/expr/mod.rs
+++ b/src/ir_lower/expr/mod.rs
@@ -9340,6 +9340,31 @@ fn method_call_result_type(
         }
         return fallback_expr_type(expr);
     };
+    // Fluent `with*` wither (@return static): a method declared to return a strict
+    // ANCESTOR interface of the receiver interface actually returns `clone $this`
+    // (the receiver's concrete type). Resolve the IR value to the receiver so a
+    // descendant-only chained call dispatches — e.g. `withHeader(): Message` on a
+    // `Request` receiver must stay `Request` for the following `->withMethod()`.
+    // This MUST agree with the checker's infer_method_call_on_interface_type (#547):
+    // if the IR value kept the ancestor type, the EIR backend rejects the chained
+    // call as "interface method call to unknown method Message::withMethod".
+    let fluent_receiver = if let (Some((recv_name, _)), PhpType::Object(return_name)) =
+        (singular_object_class(&object_ty), &return_ty)
+    {
+        if php_symbol_key(method).starts_with("with")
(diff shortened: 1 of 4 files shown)


chadmandoo:
Closing as superseded. #473 now implements sound covariant self-return validation for interface methods and removed the `with*`-prefixed return-type heuristic this PR added — reworking the same `inference/objects/methods.rs` and `type_compat/unions.rs` paths (e183b1b14, 4f666151f). Assuming the receiver interface for any `with*` method is unsound in the general case, so the sound self-return path in #473 is the right home for this. Reproducer/credit can carry over there. Thanks!

nahime0:
Superseded by #578.

#518 identified the real PSR-style fluent-chain failure and provided the original reproducer. The replacement PR carries that reproducer and contributor credit forward, but removes the method-name heuristic and addresses the underlying type-system issue instead:

- native `: static` is preserved as a late-bound method return;
- checker and EIR bind it consistently at the call-site receiver;
- class hierarchies, interfaces, static factories, nullable/union returns, override conformance, and Reflection metadata are covered;
- `withdraw()`-style false positives remain rejected;
- the unrelated array-compatibility change is not included.

Keeping #518 closed in favor of #578 so review and CI continue on the complete replacement. Thanks @chadmandoo for surfacing the original issue and reproducer.

PR #520, opened 2026-07-09
feat(types): array element covariance into a declared array<Object> parameter

chadmandoo:
## What

A PHP `array` parameter is a bare `array` — PHP does not enforce the `@param array<T>` element type at runtime, so the declared element type is a hint, not a hard bound. Two cases were rejected:

```php
final class Builder {
    /** @param list<CondI> $conditions */
    public function __construct(private array $conditions) {}
    private function mkGroup(string $op): Group { return new Group($op); }
    public function addLeaf(Leaf $c): self {
        return $this->withConditions([...$this->conditions, $c]); // spread -> Array(Mixed)
    }
    public function addGroup(string $op): self {
        return $this->withConditions([$this->mkGroup($op)]);       // method-call element
    }
}
```

## How

- `types_compatible` accepts `Array(Mixed)` (a spread literal or an adaptive local) and an element-compatible array in either covariance direction for a declared `Array(T)` param — mirroring the whole-value `(_, Mixed) => true` trust posture, one container level down.
- `array_literal_element_type_for_ir` types a method-call element as `Mixed` instead of the syntactic `Int` fallback. The old fallback stored an object-returning method's result in an `Array(Int)` and int-cast it at the element store (a hard backend error).

## Test

`test_array_element_covariance_spread_and_method_call` — a fluent builder using both a spread element and a method-call element into an object-element array, byte-parity vs PHP 8.5.

### DIFF
--- a/src/ir_lower/expr/mod.rs
+++ b/src/ir_lower/expr/mod.rs
@@ -6907,6 +6907,16 @@ fn array_literal_element_type_for_ir(
             property,
         )
         .unwrap_or_else(|| ir_array_storage_type(infer_expr_type_syntactic(item))),
+        // A method-call element has no resolved storage type here (unlike a free
+        // `FunctionCall`, whose signature is looked up above) — the syntactic
+        // fallback reduces it to `Int`, so an object-returning method (`[$this->mk()]`)
+        // would be stored in an `Array(Int)` and int-cast at the element store (a
+        // hard backend error). Store method-call elements as `Mixed`: an
+        // adaptive-boxed array is run-clean for objects and every other return type,
+        // and matches the `Array(Mixed)` covariance the call boundary now accepts.
+        ExprKind::MethodCall { .. }
+        | ExprKind::NullsafeMethodCall { .. }
+        | ExprKind::StaticMethodCall { .. } => PhpType::Mixed,
         _ => ir_array_storage_type(infer_expr_type_syntactic(item)),
     }
 }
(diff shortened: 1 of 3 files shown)


nahime0:
## Closing as superseded

This PR is no longer needed on current `main`. The motivating fluent-builder scenario already compiles and runs correctly there (output `G:AND|`, matching PHP).

### Why

Two later PRs landed the complete fix for the same surface:

1. **#519** — bare `array` parameters with object elements stay `Array(Mixed)` via `specialize_generic_array_param_hint`, so sibling/spread call sites are not checked against an Elephc-invented `array<ConcreteObject>` bound.
2. **#561** — `array_literal_element_type_for_ir` types method-call elements from the callee’s declared return type, instead of the syntactic `Int` fallback (and instead of stamping `Mixed`).

### What this PR was

A two-part **workaround**, not the final model:

- **Checker:** bidirectional `Array(T)` / `Array(U)` acceptance in `types_compatible` — a global call-boundary relaxation, broader than the bare-`array` param case that actually needed fixing.
- **IR:** `MethodCall` / `NullsafeMethodCall` / `StaticMethodCall` → `Mixed` in array literals — avoids the int-cast backend error, but is coarser than typing by return type.

Cherry-picking this onto current `main` also leaves a dead duplicate `MethodCall` arm after #561 (`unreachable_patterns` warning), which conflicts with the zero-warnings policy.

### Decision

Closing without merge. No rebase needed. If we still want the fluent-builder codegen test from this PR, it can be reintroduced on `main` as a pure test add — without the production hunks.

PR #521, opened 2026-07-09
feat(errors): register UnhandledMatchError as a builtin Error subclass

chadmandoo:
## What

Register `UnhandledMatchError` as a builtin `Error` subclass so explicit references resolve correctly, including `new`, `throw`, `catch`, `instanceof`, and match arms that explicitly throw it.

```php
function classify(int $n): string {
    return match (true) {
        $n < 0 => "negative",
        $n === 0 => "zero",
        default => throw new UnhandledMatchError("no arm for " . $n),
    };
}

try {
    classify(5);
} catch (\UnhandledMatchError $e) {
    echo $e->getMessage();
}
```

This PR does not change implicit unmatched-match lowering. When no arm matches and no `default` exists, elephc still follows its existing fatal runtime path rather than constructing a catchable `UnhandledMatchError`.

## How

- Seeds `UnhandledMatchError` in the builtin autoload class-name set.
- Declares it in checker metadata as `extends Error`.
- Applies the standard builtin Error-family constructor and Throwable signature fixups.
- Documents the explicit class surface and the existing implicit no-match limitation.
- Extends the exceptions example with an explicit match-arm throw.

## Test

- `test_unhandled_match_error_throw_and_catch`: constructs the class in a match default arm, catches it by FQN, and verifies the inherited `getMessage()` API.
- `test_match_without_default_is_fatal`: preserves coverage for the separate implicit no-match fatal path.
- The explicit regression fixture prints `no arm for 5` in both elephc and PHP.


nahime0:
Maintainer update after refreshing this PR against the current `main`:

- Merged `origin/main` into the contributor branch (`dd284a503`).
- Kept the structural fix: `UnhandledMatchError` is registered as a builtin `Error` subclass and seeded in the builtin autoload class list.
- Updated the declaration to the current `FlattenedClass` shape after the EIR migration and added it to the standard Error-family signature fixup.
- Removed the old `test_match_without_default_compiles` smoke test and its incorrect diagnosis. A default-less match already compiled before this PR; the actual regression was explicit class resolution.
- Added `test_unhandled_match_error_throw_and_catch`, which constructs the error in a match default arm, throws it, catches it by FQN, and verifies the inherited `getMessage()` API.
- Documented the supported explicit `new` / `throw` / `catch` / `instanceof` surface and clarified that the implicit unmatched-match path still uses elephc's existing fatal terminator.
- Updated the exceptions example and aligned the PR description with the implemented behavior.

Validation completed locally:

- `cargo build`
- focused explicit `UnhandledMatchError` regression test
- existing implicit no-match fatal regression test
- PHP output cross-check for the explicit throw/catch fixture
- exceptions example compile/run
- `git diff --check`

Fresh CI is running on head `5f19a66a3`. The codegen shard containing the new regression (`8/16`) is green on macOS ARM64, Linux ARM64, and Linux x86_64; the full workflow has no failures at the time of this update.

PR #522, opened 2026-07-09
fix(types): list destructuring accepts a union/assoc RHS carrying an array member

chadmandoo:
## What

`list()` / `[$a, $b] = ...` destructuring only accepted a plain `Array(T)` right-hand side and errored otherwise. A common pattern was rejected:

```php
private function mk(int $n): ?array { /* ... */ }

$entry = $this->mk($n);
if ($entry === null) { continue; }
[$key, $value] = $entry;   // $entry types as Union[Array, Void] here
```

## How

In `check_list_unpack`, accept an `AssocArray` RHS and a `Union` that carries an array-like member (`Array` / `AssocArray` / `Mixed`) — both are arrays at runtime in well-typed code. Their elements bind as `Mixed` (adaptive access), the trust posture PHP applies at runtime.

## Test

`test_list_unpack_null_narrowed_array_return` — destructuring a null-narrowed `?array` return in a loop, byte-parity vs PHP 8.5.


nahime0:
Maintainer follow-up:

The original `Union(...).any(array-like)` workaround has been replaced locally with a fix at the narrowing boundary.

Changes prepared on the maintainer worktree:

- merged the current `main` into the PR branch (`7dc840c89`);
- changed post-guard analysis so direct `continue` and `break` bodies are treated as non-fallthrough, preserving the complement after `if ($entry === null) { ... }`;
- removed the permissive union branch from `check_list_unpack`, so an unguarded `?array` is rejected again instead of being accepted because one union member is array-like;
- kept `AssocArray` as an independently valid list-unpack RHS, with conservative `Mixed` element typing;
- added focused regressions for `continue`, `break`, associative storage, and rejection of nullable arrays without a guard.

Local validation is green:

- `cargo build`
- focused list-unpack error tests: 2 passed
- focused list-unpack codegen tests: 3 passed
- existing `throw`-guard narrowing regression: passed
- `cargo fmt --check`
- `git diff --check`

The broader pre-existing limitation in the block fallthrough analysis is intentionally kept out of this PR and tracked separately in #590.

Issue #527, opened 2026-07-10
foreach by-value over an array of arrays leaks two blocks per iteration (element copy never released)


mirchaemanuel:
## Summary

A by-value `foreach` over an array of arrays leaks exactly **2 heap blocks per iteration**: the owned per-element copy bound to the loop variable is never released, neither on iteration advance nor at loop exit.

## Environment

- elephc main `e63d5d337` (v0.26.0-158), macOS arm64 (Apple Silicon); PHP 8.4.23 as behavioral reference.

## Repro

```php
<?php
$a = [];
for ($i = 0; $i < 4; $i++) { $a[] = ['kw', 'word' . $i]; }
foreach ($a as $row) { echo $row[1] . "\n"; }
```

```
elephc --heap-debug foreach.php && ./foreach
```

## Observed

```
word0
word1
word2
word3
HEAP DEBUG: allocs=35 frees=27 live_blocks=8 live_bytes=352 peak_live_bytes=1352
HEAP DEBUG: leak summary: live_blocks=8 live_bytes=352
```

Output matches PHP exactly; exit 0. **Scaling confirmed**: the same program with 20 rows leaks `live_blocks=40 live_bytes=1760` — exactly 2 blocks per iteration.

## Expected

Leak-free: each iteration's element copy should be released when the loop variable is rebound to the next element and at loop exit.

## Notes

- Suspected mechanism: by-value `foreach` binds each element via `IterCurrentValue`, which produces an owned fresh copy — it is even listed among owning temporaries in `src/ir_lower/context.rs::value_is_owning_temporary` — but no release is emitted on advance/exit for the copy bound to `$row` (a boxed cell + payload, consistent with the 2-blocks-per-iteration signature).
- Under `--web` this compounds per request (see #516 for the mechanics).
- Reproduces byte-identically on pristine main and on the #524 branch (that fix targets chained reads; the foreach copy path is untouched).
- Found while regression-testing #524.

mirchaemanuel:
[...]

PR #531, opened 2026-07-11
fix(ir_lower): release the nullable-access hidden temp after its consuming chained read

### LINKED ISSUE #525: Chained subscript read on a ?array receiver leaks the returned containers (nullable-access hidden temp is never released)
## Summary

When the receiver of a chained subscript read is nullable (`?array`), the intermediate container goes through the nullable-access path's hidden owned temp, and that temp's reference is never dropped: every non-null call leaks the returned outer array, inner array, and string payloads (~7 blocks per call in the repro below).

This is a **different path** from the plain chained-subscript leak fixed by #524: that fix releases receivers that are direct `array_get`/`hash_get` results, while here the consuming read's receiver is a `LoadLocal` of a hidden owned temp materialized by `lower_nullable_array_access` (`src/ir_lower/expr/mod.rs`, via `declare_owned_hidden_temp`/`take_owned_temp`). Verified unaffected by #524 — byte-identical leak report with and without that fix.

## Environment

## Repro

```php
<?php
function f(int $i): ?array {
    if ($i > 0) { return [['a', 'b' . $i]]; }
    return null;
}
for ($k = 0; $k < 30; $k++) {
    $r = f($k);
    $x = (string) $r[0][1];
    echo $x . ';';
}
echo "\n";
```

```
elephc --heap-debug nullable_chain.php && ./nullable_chain
```

## Observed

```
;b1;b2;…;b29;
HEAP DEBUG: allocs=325 frees=120 live_blocks=205 live_bytes=11680 peak_live_bytes=11856
HEAP DEBUG: leak summary: live_blocks=205 live_bytes=11680
```

205 blocks / 11,680 bytes leaked over 30 calls (~7 blocks per non-null call). Printed values are correct; exit 0. (PHP additionally warns `Trying to access array offset on null` for the `$k = 0` call, which elephc omits — minor divergence, secondary to the leak.)

## Expected

Leak-free: the hidden owned temp holding the intermediate container should be released after the consuming read, exactly as the non-nullable chained read is after #524.

## Notes

- Same "owned expression temporary never released" family as #483; filed separately because the leaking path (the nullable-access hidden temp) is specific, with a precise repro.
- Under `--web` this class of leak compounds per request until heap exhaustion (see #516 for the mechanics: latency ramp + `Fatal error: heap memory exhausted`).
- Found while regression-testing #524.

mirchaemanuel:
Fixes #525: a chained subscript read on a `?array` receiver leaks the intermediate container on every evaluation.

## Root cause (confirmed at source level)

`lower_nullable_array_access` (`src/ir_lower/expr/mod.rs`) materializes the intermediate through a hidden owned temp (`declare_owned_hidden_temp` → `store_value_into_temp` acquires +1) and returns it via `take_owned_temp`, which `LoadLocal`s the value and **clears the slot without releasing**. Because `value_is_nullable` is false for `Mixed`, the consuming outer read does not re-enter the nullable path — it goes to `lower_array_access_from_value` with receiver = the owned-temp load, taking the `Op::RuntimeCall` arm (`__rt_mixed_array_get`). `store_local` knows how to release owned-temp loads; the read path had no analogous release, so the temp's reference leaked (~7 blocks per non-null call in the issue repro: outer array + inner array + string payloads).

## The fix (IR-level, all targets)

In `lower_array_access_from_value`, after emitting the read: when the op is the boxed-Mixed runtime read **and** the receiver is an owned-temp load, emit a release through the existing `release_if_owned` primitive. Exactly once (the temp slot is already cleared, so this read is the last consuming use), covering both merge results — the populated container from the non-null path and the boxed Mixed null from the null path (`__rt_decref_mixed` is null-tolerant).

Safety: `__rt_mixed_array_get`'s documented contract (`src/codegen_support/runtime/objects/mixed_array_get.rs`) is that *every successful return is an owned Mixed* — string payloads persisted, child heap pointers retained — so releasing the receiver cannot invalidate the just-extracted result. Typed reads (`ArrayGet`, `StrCharAt`, …) can return borrowing payloads and are deliberately untouched.

## Evidence (macOS arm64, base v0.26.1 `93f576168`; all outputs byte-identical pre/post and match PHP 8.4.23)

| Program | before (live blocks) | after |
|---|---:|---:|
| issue repro (30 calls) | 205 | 175 — the −30 is exactly the hidden temp/iteration |
| chained read isolated from unrelated leaks | 146 | **116 = the store-only floor: #525 construct fully clean** |
| deeper `$r[0][0][1]` | 322 | 292 (−30, its one hidden temp) |
| null-receiver case | 2 | 1 (boxed-null temp now freed; no new warnings) |
| `$r[0][0] . $r[0][1]` (temp used twice per iter) | 294 | 234 (−60, no double-release) |
| plain non-nullable chained read | 90 | 90 (unchanged, #524's territory) |

The post-fix residual in the raw repro decomposes exactly (arithmetic checked) into two **pre-existing, independent** leaks that reproduce on clean main with zero nullable chained reads: a `?array`-return/store leak (~4 blocks/call) and the Mixed→string cast-source leak (fixed separately in the sibling PR for #527). Both are follow-up candidates.

## Tests

2 regression tests in `tests/codegen/runtime_gc/regressions.rs` (using a `?array` parameter so the unrelated pre-existing leaks cannot mask the assertion; both exercise the null and non-null path every iteration). Proven fail-pre-fix (`live_blocks=250`/`400` via stashed-src build), pass post-fix. Focused suites all green: nullsafe 21, nullable 34, nested 119, runtime_gc 124, arrays 337, array_access 24, isset 20, coalesce 52, ternary 36, mixed 221. `cargo build`/`--release` zero warnings; debug-profile IR validator clean.

## Notes for the maintainer

- **Textual overlap with #524**: both PRs append a release block to the same tail of `lower_array_access_from_value`. The union is trivial and semantically safe — the two conditions are disjoint (a receiver's defining op is either an owned-temp load or a direct index read, never both). I merged #524+#525+#527 locally to verify: union resolution compiles clean and all three issue repros stay fixed/leak-free.
- Known not-covered (documented, pre-existing): the *middle* intermediate of 3-level chains on the Mixed runtime path (`$r[0][1][0]` — receiver is a direct RuntimeCall result, not a hidden temp); `($r ?? [])[0][1]` is a pre-existing compile error on main (`runtime_call with receiver PHP type Void`); the `?array`-return/store leak. Follow-up issue candidates.


nahime0:
Maintainer follow-up after integrating the current `main` locally (including #524):

- Resolve the overlapping `lower_array_access_from_value` and runtime-GC test hunks by keeping both disjoint release paths and both regression suites. The #524 receiver is defined by an index-read op; the #531 receiver is a `LoadLocal` from a one-shot owned temp, so the two releases cannot target the same SSA value.
- Tighten the #531 predicate so this release is limited to a `PhpType::Mixed` receiver handled by `__rt_mixed_array_get`. `Op::RuntimeCall` also covers `ArrayAccess` object dispatch, so checking the opcode alone is broader than the ownership proof in this PR.
- Clarify the safety explanation: ownership is established structurally by the `LocalKind::OwnedTemp` load after `take_owned_temp` clears its slot. `release_if_owned` itself is type-gated; it does not filter `Borrowed`/`Moved` ownership states.
- Re-run the #525 and #516 regression tests, the nullable-focused codegen tests, `cargo build`, and `git diff --check` on the merged tree.

This keeps #531 surgical while preserving the semantics and coverage added by #524.

Issue #529, opened 2026-07-10
Nested write into an Array(Mixed) element ($a[$i][$j] = ...) is silently lost: the parent keeps the stale inner array


mirchaemanuel:
## Summary

A nested write into an `Array(Mixed)` element — `$a[$i][$j] = ...` — is **silently lost**: the program compiles and runs, but the parent array keeps the stale inner array (a correctness miscompile), and the mutated copy leaks.

This is the only nested-write shape that compiles at all: on a homogeneous typed nested array the same statement is a compile-time error (`error[4:1]: Nested array assignment requires a Mixed or ArrayAccess target`), so the Mixed path below is the supported target shape — and it miscompiles.

## Environment

- elephc main `e63d5d337` (v0.26.0-158), macOS arm64 (Apple Silicon); PHP 8.4.23 as behavioral reference.

## Repro

```php
<?php
// Heterogeneous elements force Array(Mixed), the supported nested-write target shape.
$a = [['x', 'y0'], ['x', 'y1'], ['x', 'y2'], 7];
$a[2][1] = 'patched';
echo $a[2][1] . "\n";
echo $a[1][1] . "\n";
```

```
elephc --heap-debug write_read.php && ./write_read
```

## Observed

```
y2
y1
HEAP DEBUG: allocs=24 frees=10 live_blocks=14 live_bytes=744 peak_live_bytes=1144
HEAP DEBUG: leak summary: live_blocks=14 live_bytes=744
```

The write is silently lost (`y2` instead of `patched`), plus 14 leaked blocks.

## Expected (PHP 8.4.23, verbatim)

```
patched
y1
```

## Notes

- Suspected mechanism: the nested-lvalue write path on a Mixed element mutates a materialized **copy** of the inner array (the boxed Mixed cell's payload is unboxed/copied rather than written back through the parent slot), so the parent keeps the stale inner array and the mutated copy leaks.
- With the #524 fix applied, the wrong output is byte-identical (`y2` / `y1`); only the leak count changes (14 → 6, from the released chained-read intermediates in the two `echo` lines) — the miscompile itself is unaffected.
- Severity: silent data loss on a construct that compiles without warnings; arguably the most severe of the batch filed from this test run.
- Found while regression-testing #524.

mirchaemanuel:
[...]

Issue #345, opened 2026-06-08
`foreach` by reference followed by value iteration segfaults on indexed arrays


nahime0:
## Description

A by-reference `foreach` over an indexed array can leave the array in a bad runtime state for a later by-value `foreach` over the same array.

PHP 8.4.19 prints the expected mutated values. The current Elephc tree compiles successfully, but the generated binary exits with `SIGSEGV`.

## Minimal repro

```php
<?php
$a = [1, 2, 3];
foreach ($a as &$v) { $v *= 2; }
unset($v);
foreach ($a as $v) { echo $v; }
```

## Steps to reproduce

1. Save the repro as `main.php`.
2. Run `php -d display_errors=1 -d error_reporting=E_ALL main.php`.
3. Compile with Elephc and run the produced binary.

## Expected behavior

PHP output:

```text
246
```

The Elephc-compiled binary should print the same output and exit `0`.

## Actual behavior

- PHP 8.4.19: stdout `246`, exit `0`
- Elephc current tree: compile succeeds, binary exits `SIGSEGV` (`-11`) with no stdout/stderr

Direct indexed reads after the by-ref loop still work:

```php
<?php
$a = [1, 2, 3];
foreach ($a as &$v) { $v *= 2; }
unset($v);
echo $a[0], $a[1], $a[2];
```

That prints `246` under Elephc, so the crash seems specific to the later by-value `foreach` path.

## Environment

- Elephc crate version during build: `0.23.7`
- Git commit: `09cb0523`
- OS: macOS arm64
- PHP baseline: `PHP 8.4.19 (cli)`

PR #533, opened 2026-07-11
fix(codegen): guard container consumers against the null-container sentinel (chained-read miss segfault)

### LINKED ISSUE #526: Chained subscript read where the first index misses segfaults (null-fallback sentinel dereferenced as an array pointer)
## Summary

In a chained subscript read, a miss on the **first** index (so the miss fallback becomes the receiver of the second read) segfaults: for a refcounted element type the null fallback materializes the scalar sentinel `0x7fff_ffff_ffff_fffe` instead of a null container, and the consuming outer read then loads the array length header from that sentinel address and faults.

A miss on the **second** index alone behaves correctly (warns, yields empty string, exits 0, heap clean) — only a first-index miss whose fallback is consumed as a container crashes.

## Environment

## Repro

```php
<?php
$a = [];
for ($i = 0; $i < 3; $i++) { $a[] = ['kw', 'word' . $i]; }
// Miss on the first index: the second read consumes the miss fallback.
$x = (string) $a[7][1];
echo 'x=' . $x . "\n";
echo 'done' . "\n";
```

```
elephc miss_inner.php && ./miss_inner    # also reproduces with --heap-debug
```

## Observed

```
```
then **SIGSEGV** (exit 139). The two `echo` lines never run.

## Expected (PHP 8.4.23)

```
Warning: Undefined array key 7 …
Warning: Trying to access array offset on null …
x=
done
```
exit 0.

## Notes

- Suspected mechanism: `emit_array_get_null_fallback` (`src/codegen/lower_inst/arrays.rs`, the `_ =>` arm) materializes the scalar sentinel for refcounted element types; the consuming `array_get` then executes `emit_load_from_address(len, array_reg, 0)` against the sentinel value as if it were an array pointer.
- The working variant for contrast: `$a[1][7]` (first read hits, second misses) warns and continues correctly.
- Reproduces identically on pristine main and on the #524 branch (that fix only releases valid intermediates; the miss fallback path is untouched).
- Found while regression-testing #524. Severity: crash on legitimate PHP (an out-of-range read that PHP handles with warnings).

mirchaemanuel:
Fixes #526 — and the crash class turned out wider than the report: on main, **all** of these segfault (exit 139), not just the chained read: `isset($a[7][1])`, `$a[7][1] ?? …`, `$m['nope']['x']`, `$a[9][0][0]`, `$a[0][9][0]`, `foreach ($a[7] ?? [] as $v)`, `foreach ($a[7] as $v)`.

## Root cause

`emit_array_get_null_fallback` (`src/codegen/lower_inst/arrays.rs`) materializes the in-band `NULL_SENTINEL` (`0x7fff_ffff_ffff_fffe`) for missed reads of refcounted element types, and every downstream container consumer dereferenced it unguarded: `lower_array_get_{aarch64,x86_64}` loaded the length header straight from it; the same held for `__rt_hash_get`, `__rt_array_get_mixed_key`, `__rt_hash_iter_next` (null-guarded `0` but not the sentinel), the isset array probe, and the foreach length snapshot. Compounding it, `IsNull` on statically-container-typed slots was **hard-coded false**, so `$a[7] ?? []` evaluated the miss as non-null and propagated the sentinel into foreach.

## Design

Consumer-side guards, keeping the sentinel as the established null-container representation — it is already what `??`/isset compare against for scalars, what `Throwable::getPrevious()` uses for null objects, and what incref/decref already skip via heap-range checks, so it is GC-safe by construction. (Changing the fallback to a different representation was considered and rejected: the sentinel-as-null convention is load-bearing across comparisons, so a representation change would ripple much wider.)

- New shared emitter helper `emit_branch_if_null_container` (`src/codegen_support/sentinels.rs`): branch on `reg == 0 || reg == NULL_SENTINEL`, symmetric ARM64/x86_64 arms.
- Guarded: `lower_array_get_*` (jumps to the fallback **past** the warning, so the Silent variants stay silent automatically), the isset array probes, `IsNull` for `Array`/`AssocArray`/`Iterable`/`Object`, the foreach indexed-length snapshot; plus 3-instruction sentinel extensions of the existing null guards in `__rt_hash_get`, `__rt_array_get_mixed_key`, `__rt_hash_iter_next` (both arches each).
- One ir_lower change so isset/`??` silence propagates through the whole subscript chain (`isset($a[7][1])` must not warn at any level — PHP semantics).

## Before → after (base v0.26.1 `93f576168`, macOS arm64; all post-fix runs exit 0 and are PHP-compared)

| Case | before | after |
|---|---|---|
| `$a[7][1]` (issue repro) | warn + **SIGSEGV** | warn, `x=`, `done`, heap clean |
| `$a[1][7]` (contrast) | OK | byte-identical to baseline |
| `isset($a[7][1])` | **SIGSEGV** | `false`, no warning |
| `$a[7][1] ?? 'dflt'` | **SIGSEGV** | silent, no crash (yields `''` — pre-existing gap, see notes) |
| `$m['nope']['x']` string-keyed | **SIGSEGV** | `NULL`, mirrors the working direction |
| 3-level miss at each level | 2 of 3 **SIGSEGV** | `NULL` ×3, one warning per real miss |
| `foreach ($a[7] ?? [] as $v)` / `foreach ($a[7] as $v)` | **SIGSEGV** | iterates empty / warns and skips |
| Mixed-element miss | OK | unchanged |

`--heap-debug`: the repro and every previously-crashing variant end clean; the leaks remaining on some second-index-miss paths are byte-identical on unmodified main (pre-existing; they belong to the leak family being fixed in #524/#531/#532 — verified composed: on a local merge of those PRs the same programs end clean).

## Cross-arch verification

Both arch arms edited symmetrically in every touched emitter (diff-reviewed). Beyond that, I captured the actual **x86_64 assembly** on this host (user asm via a runtime-cache pre-seed, runtime asm via an `as` shim — the toolchain otherwise insists on assembling with the host `as`) and verified the emitted guards verbatim (`test/jz` + `movabs` sentinel + `cmp/je`) in `array_get`, the isset probe, `is_null_container`, the iterator length snapshot, `__rt_hash_get`, `__rt_array_get_mixed_key`, `__rt_hash_iter_next`. Executing x86_64 locally is not possible on this host — relying on CI for the run matrix.

## Tests

10 new regression tests in `tests/codegen/regressions/arrays.rs` (crash repro asserting stdout+warning+exit, second-index guard, heap-clean assertion, isset/`??` silence, string-key chain, 3-level chain, foreach direct + coalesce, Mixed-element guard); the exact fixtures exit 139 on main. Focused suites, 0 failures: arrays **347** (incl. the 10 new), array_basics 81, isset 21, nested 119, coalesce 54, foreach 105, is_null 18, previous 3, runtime_gc 122. Zero build warnings; `scripts/check_asm_comments.py` clean on all touched codegen files.

## Notes for the maintainer

- Pre-existing gap, deliberately not addressed here: `$a[7][1] ?? 'dflt'` yields `''` instead of `'dflt'` because `IsNull` on a statically-`Str` slot is constant-false — identical behavior for the already-working `$a[1][7] ?? 'dflt'` on main. Fixing it means giving `Str` a null representation, which needs an audit of every empty-string producer; follow-up issue candidate.
- PHP's *second* diagnostic (`Trying to access array offset on null` / `foreach() argument … null given`) is not emitted — no such warning infrastructure exists, and the already-working miss path omits it too.
- By-ref foreach over a missed element and chained *writes* to missed elements are out of scope (read path only).
- Touches `src/ir_lower/expr/mod.rs` in different hunks than #531/#532 (and #524) — expected to auto-merge.


nahime0:
Opened #554 to track the pre-existing statically-`Str` null-coalescing gap documented in this PR.

It covers both `$a[7][1] ?? 'dflt'` after this PR's sentinel guard and the already non-crashing `$a[1][7] ?? 'dflt'` behavior on `main`, with explicit controls ensuring that a legitimate empty string still does not select the default. The issue also captures the required audit of string null representation, storage classes, ABI paths, and ownership.

Keeping this separate lets #533 remain focused on preventing null-container sentinel dereferences while #554 owns the PHP-semantic follow-up.

nahime0:
I opened separate follow-ups for the two write/by-reference cases called out as out of scope here:

- #556 — by-reference `foreach` over a missing element dereferences the null-container sentinel and crashes.
- #555 — chained writes through a missing intermediate element fail to autovivify the parent, lose the write, and leak.

This keeps #533 focused on read/by-value consumer safety. #556 also distinguishes the new crash from the separate by-ref loop issue in #345, while #555 links the existing-parent work in #529/#553.

PR #546, opened 2026-07-16
fix: close residual ownership leaks from issue 540

### LINKED ISSUE #516: --web: linear per-worker latency ramp is a per-request refcount leak (chained subscript reads on nested arrays); --max-requests recycles miscounted by the crash-loop guard

### LINKED ISSUE #540: Residual EIR ownership leaks keep persistent-worker latency growing after #516/#538
## Summary

The dominant leaks tracked in #516 and #538 were fixed, but current `main` still has five independent EIR ownership gaps that accumulate in persistent `--web` workers.

On `main` at `8eb03688b` (macOS ARM64), the Ivory PHP-highlight route leaks exactly **31 heap blocks per request** (about **21 KiB/request**). Every orphaned block observed by a block-level heap dump had refcount `1`, so these are missing owner releases rather than cycles.

The remaining leaks also explain the residual latency ramp. They are grouped here because they share one reproducible workload and one acceptance benchmark, but each item should be fixed in a separate PR.

## Exact attribution

| Residual ownership gap | Blocks/request | Evidence |
|---|---:|---|
| Constructor argument temporaries are not released | **14** | `+2` from `new App(new Db(...))`; `+12` from constructing `Paste` with four owned Mixed property-read temporaries |
| `file_get_contents()` is not classified as an owning builtin result | **4** | Two successful reads; each leaves its boxed Mixed cell and owned string alive |
| Returning `new Paste(...)` through `?Paste` retains the original object after `MixedBox` | **5** | The `Paste` object plus its four string properties remain alive |
| An object receiver unboxed/retained from a Mixed local is not released after `PropGet` | **5** | One retained `Paste` graph: object plus four string properties |
| The call-result alias guard keeps an inline array passed to an Array-returning method | **3** | The inline `['//', '#']` argument plus its two string elements remain alive |
| **Total on the PHP route** | **31** | Exact match with the measured per-request slope |

The route-level accounting is also exact:

- `/about`: `+2`
- `/raw/{id}`: `2 + 4 + 5 = +11`
- highlighted plain paste: `+28`
- highlighted PHP paste: `28 + 3 = +31`

### 1. `ObjectNew` does not clean up owned argument temporaries

`src/ir_lower/expr/mod.rs::lower_new_object()` lowers constructor operands and emits `ObjectNew`, but unlike ordinary calls it never invokes owned-argument cleanup after the constructor has retained or copied its arguments.

This affects both nested object construction and constructor arguments produced by property reads/casts. Binding the nested object or converted property values to typed locals removes these leaks through normal local cleanup.

### 2. `file_get_contents()` ownership metadata is incomplete

`src/codegen/lower_inst/builtins/io.rs::lower_file_get_contents()` returns a freshly allocated boxed string-or-false value. However, `builtin_call_result_owns_storage_as_temporary()` in `src/ir_lower/context.rs` does not include `file_get_contents`.

Consequently, a string cast or retaining consumer does not recognize the boxed result as an owning temporary and never releases the original box/string pair.

### 3. Mixed return boxing retains but does not release its source

`src/ir_lower/stmt/mod.rs::coerce_to_return_type()` emits `MixedBox` when a concrete object is returned through a nullable/Mixed storage type. The runtime `__rt_mixed_from_value` correctly retains the object for the new box, but lowering does not release the original owned `ObjectNew` result afterward.

The caller releases the returned Mixed box correctly; the original source reference still leaves the object graph at refcount `1`.

### 4. Property reads do not clean up an owning receiver

When a local has Mixed frame storage but is known as an object at a later use, loading it unboxes and retains the object. `lower_property_get()` / `lower_property_get_from_value()` emits `PropGet` without releasing that owning receiver.

A fix must stabilize/acquire the property result before releasing a sole-owner temporary receiver. Blindly releasing the receiver after `PropGet` can turn expressions such as `make()->stringProperty` into use-after-free bugs.

### 5. Call-result alias detection is too coarse

`call_result_may_alias_arg()` treats every `Array -> Array` call as possibly returning the same payload. This is necessary for real identity-style functions, but it also prevents cleanup when a method such as the tokenizer's `scan()` returns a newly allocated array.

Passing `['//', '#']` inline therefore leaks three blocks. Binding that array to a local makes the route clean for this component.

The fix should use a precise return-alias/ownership contract or summary. Simply deleting the `Array/Array` guard would risk freeing a real returned alias.

## Latency impact

The renderer and the highlighter result itself add no residual blocks. The PHP tokenizer does, however, execute about **33,843 allocator operations per request**, compared with about **999** for the plain path. The leaked blocks pin and fragment the allocator's first-fit free lists, so the high-churn PHP path amplifies a relatively small live-block leak into a large latency ramp.

Fresh-worker, single-worker, 500-request benchmark:

| Variant | First 10 avg | Last 10 avg | Request 500 |
|---|---:|---:|---:|
| Current source | 3.43 ms | **45.06 ms** | **47.66 ms** |
| Diagnostic source workarounds removing 27/31 blocks | startup-noisy | **3.58 ms** | **3.47 ms** |

The workaround version still leaked exactly the four blocks from the two `file_get_contents()` calls, but its latency stayed flat. This is diagnostic evidence, not a proposed application-level fix.

## Work items

- [ ] Classify `file_get_contents()` as returning fresh owned storage, preferably through builtin ownership metadata rather than another long-lived manual exception.
- [ ] After retaining `MixedBox` return coercion, release an owned concrete source without changing borrowed/aliased returns.
- [ ] Release owned constructor argument temporaries after `ObjectNew`, including nested objects and Mixed-to-typed argument conversions; preserve by-reference constructor semantics.
- [ ] Stabilize property-read results and release owning temporary receivers for named, dynamic, and nullsafe property paths without introducing use-after-free.
- [ ] Replace the type-only call-result alias guard with precise alias/return-ownership information so fresh Array-returning calls release inline Array arguments while true aliases remain valid.

## Acceptance criteria

- [ ] Each item has a focused heap-debug/GC regression that fails before the fix and reports zero residual ownership afterward.
- [ ] Cover `--ir-opt=on` and `--ir-opt=off` where relevant.
- [ ] Cover macOS ARM64, Linux ARM64, and Linux x86_64 through the shared EIR/runtime path.
- [ ] Add explicit aliasing/COW and borrowed-result controls for the delicate property-receiver and call-alias fixes.
- [ ] On the Ivory benchmark route, a single worker shows no per-request live-block growth after warmup.
- [ ] Early and late latency remain flat across 500 identical sequential requests.
- [ ] Keep this umbrella issue open until all five work items and the combined benchmark are complete.

## Related

- #516 — original persistent-worker leak and latency-ramp investigation.
- #538 / #539 — dominant indexed-to-hash COW-generation leak and its fix.
- #512 — related returned-object ownership/corruption family.
- #527 / #532 — Mixed stringification ownership fixes.

nahime0:
## Summary

- make Mixed-to-string call conversions explicit EIR-owned values so normal return/argument alias analysis can transfer or release them safely
- add single-source builtin metadata for result storage that cannot alias arguments, covering `rawurldecode()`, `implode()`, `htmlspecialchars()`, and `htmlentities()`
- release owning scalar-cast inputs and constructor-created temporary Mixed ABI boxes
- add heap-debug regressions for independent results, passthrough aliases, named arguments, later by-reference receiver mutation, constructor conversions, and optimizer on/off behavior

## Why

The fixes merged for #540 removed the five known root causes behind #516, but the original Ivory persistent-worker workload still retained 10 heap blocks per rendered request. The remaining growth came from three gaps:

1. stabilized inline reads consumed by scalar casts or independent builtins were not released;
2. Mixed-to-string argument conversion ownership was hidden in ABI materialization;
3. wrapper return-alias analysis treated independent builtin results as potentially reusing argument storage.

The conversion now lives in EIR, where the existing alias summary can distinguish independent results from true passthroughs. This deliberately preserves the stabilizing owner across later by-reference argument mutations and avoids relying on pointer equality for string slices.

## Verification

- `cargo check`
- `cargo build --release`
- 16 focused `test_regression_540_*` heap-debug tests
- focused #516, object-new, Mixed-boxing, and `file_get_contents()` ownership regressions
- optimizer coverage with `--ir-opt=on` and `--ir-opt=off`
- builtin registry/parity tests
- generated builtin docs audit and site compatibility validation

On the original Ivory source, without application-level workarounds:

| Route | Requests | Live blocks |
|---|---:|---:|
| `/about` | 100 | 160 → 160 |
| `/raw/b00080` | 100 | 160 → 160 |
| `/p/t00080` | 100 | 160 → 160 |
| `/p/b00080` | 500 | 160 → 160 |
| `/p/b00080` with `--no-ir-opt` | 100 | 160 → 160 |

Across three fresh 500-request latency runs, the average first-50 latency was 3.167 ms and the average last-50 latency was 2.991 ms, so the previous per-worker latency ramp is gone.

Closes #540.
Closes #516.


nahime0:
Checked against #516 . All criteria are satisfied. This should close also #516

Issue #369, opened 2026-06-15
Integer overflow wraps instead of promoting to float


nahime0:
## Description

Elephc currently wraps signed integer overflow for `int + int` at the 64-bit boundary, while PHP promotes the result to `float`.

## Minimal reproduction

```php
<?php
$x = 9223372036854775807;
var_dump($x + 1);
```

## Steps to reproduce

1. Save the program above as `main.php`.
2. Run `php main.php`.
3. Compile and run with Elephc:
   - `target/debug/elephc main.php`
   - `./main`

## Expected behavior

PHP 8.4.19 prints:

```text
float(9.223372036854776E+18)
```

Elephc should match that promotion behavior.

## Actual behavior

The current Elephc `main` build prints:

```text
int(-9223372036854775808)
```

So the result is wrapping as a two's-complement 64-bit integer instead of promoting to float.

## Environment

- Elephc crate version: 0.23.11
- Git commit: `120d6704a8665b05b6591e2a7aee31cc7056a350`
- PHP baseline: PHP 8.4.19 (cli)
- OS: macOS 26.5 / Darwin 25.5.0 arm64
- Reproduced on: 2026-06-15

Issue #397, opened 2026-06-22
Loose equality / switch truncates a Mixed (untyped) float operand to int


nahime0:
## Summary

Loose equality (`==`) and `switch` give wrong results when one operand is an **untyped / `Mixed`** value that holds a `float` at runtime. The `Mixed` operand is unboxed via `__rt_mixed_cast_int`, which **truncates** the float to an int before comparing, so `1.5` compares as `1`.

This is the `Mixed` counterpart of the statically-typed float loose-equality fix (float-vs-int `==` and float `switch` for *typed* operands). The typed case is fixed; this untyped case is not, because there is no tag-aware runtime loose-comparison helper for `Mixed` operands.

## Reproduction

```php
<?php
function classify($x) {            // untyped param -> Mixed
    switch ($x) {
        case 1:   return "int-one";
        case 1.5: return "onefive";
        default:  return "other";
    }
}
echo classify(1.5), "\n";
```

Also reproduces with a bare comparison through an untyped variable, e.g. `$m = some_mixed(); var_dump($m == 1);` where `$m` holds `1.5`.

## Expected (PHP)

```
onefive
```

## Actual (elephc)

```
int-one
```

The `Mixed` subject `1.5` is cast to int `1` and matches `case 1`.

## Root cause

In `src/codegen_ir/lower_inst/comparisons.rs`, `lower_loose_eq` routes a `Mixed` operand through the integer path (`loose_intish_comparable` → `emit_intish_compare` → `load_intish_value`), which unboxes a `Mixed` with `__rt_mixed_cast_int`. A `Mixed` holding a float is therefore truncated before the comparison, instead of being compared with PHP `==` semantics on its actual runtime type tag (int / float / string / bool / null).

The switch dispatch (`src/ir_lower/stmt/mod.rs`, `lower_dynamic_switch_dispatch` / `float_loose_eq_pair`) can only divert *statically-typed* floats to `Op::LooseEq`; a `Mixed` subject keeps the integer fast path.

## Suggested fix

Add a tag-aware runtime loose-comparison helper (e.g. `__rt_mixed_loose_eq`) that inspects the `Mixed` tag and applies PHP `==` rules (numeric compare for int/float, numeric-string parsing for strings, truthiness for bool, null rules), then route `LooseEq`/`LooseNotEq` involving a `Mixed` operand through it. The switch dispatch should then send `Mixed` subjects down the loose-equality path as well.

## Scope / notes

- Pre-existing limitation; not a regression.
[...]

PR #547, opened 2026-07-16
fix(ir): release checked-arithmetic Mixed boxes consumed by parent operations (#500)

### LINKED ISSUE #500: Regression since v0.26.0: parenthesized integer operands of binary ops leak one temporary per evaluation (pure int math)
### Summary

Regression between v0.26.0 and current main: integer expressions whose binary operator has a **parenthesized compound operand** leak one heap temporary per evaluation. On the default 8 MB heap a tight loop dies with `Fatal error: heap memory exhausted` within ~1.8M evaluations. The same code split into single-operation statements does not leak. **v0.26.0 is clean** on all of these repros.

Possibly related to the recent memory-model/optimizer work (and to the "owned temporaries never released" family, e.g. #483 / #486), but note this triggers with **pure int arithmetic** — no calls, no objects, no Mixed in sight — and `--ir-opt=off` does not help, so it looks like base codegen.

### Environment

- **Leaks:** main @ `e63d5d337` (local build), macOS ARM64
- **Clean:** v0.26.0 official release, same machine (peak RSS stays ~1 MB)
- Reference: PHP 8.4.23 runs everything correctly

### Repro

```php
<?php
// leaks on main, clean on v0.26.0 (default 8MB heap)
function comboParen(): int {
    $acc = 0;
    for ($i = 0; $i < 3000; $i++) { $acc = ($i * 7 + 1) & 0xFFFF; }   // LEAKS
    return $acc;
}
function comboVar(): int {
    $acc = 0;
    for ($i = 0; $i < 3000; $i++) { $v = $i * 7 + 1; $acc = $v & 0xFFFF; }   // clean
    return $acc;
}
for ($k = 0; $k < 600; $k++) { comboVar(); }
echo "comboVar ok\n";
for ($k = 0; $k < 600; $k++) { comboParen(); }
echo "comboParen ok\n";
```

Observed on main: `comboVar ok` prints, then `Fatal error: heap memory exhausted`.

More shapes that leak on main (each verified in isolation, all clean on v0.26.0):

```php
$acc = ($i * 7 + 1) % 65536;          // parenthesized operand of %
$acc = $SIN[($i * 7 + 5) & 1023];     // parenthesized operand inside an array INDEX
$v = $a + $b + $c;                    // 3+ term chained sum of array-read values
```

while these are clean on main too:

```php
$acc = $i & 0xFFFF;                   // plain var & literal
$acc = $i >> 2;   $acc = $i | 16;     // shifts / or
$acc = $i * 7 + 1;                    // mul+add without a masking parent op
```

### Expected

Integer arithmetic should not allocate (or should release its temporaries). v0.26.0 behavior restored.

mirchaemanuel:
## Summary

The #369 overflow-to-float work (PR #441) lowers runtime int `+`/`-`/`*` (and unary `-`) through checked-arithmetic ops that each return a freshly allocated **owned** boxed `int|float` Mixed cell (40 bytes). The assignment and call-argument paths release that box correctly, but every *inline* consumption site — coercing it as the operand of `%`, a bitwise/shift op, `**`, a comparison, or an array index — never did, leaking one box per evaluation until the heap dies (`Fatal error: heap memory exhausted` in ~1.8M evaluations on the default heap).

The fix generalizes the proven #527 pattern: `release_stringified_source_if_owned` becomes a shared `release_coerced_source_if_owned` helper (same `value_is_owning_temporary` gate plus a repr-releasability check), applied at the int/float coercers and the array-index key path. Borrowed locals are untouched — the split `$v = $i * 7 + 1; $acc = $v & 0xFFFF;` shape stays exactly as before.

Per your analysis on the issue: the `**` exponentiation path is covered (base and exponent positions, int and float exponents — each leaked 1 block/eval on post-#546 main, verified differentially), and the chained array-read sum is indeed already clean on current main; it is kept as a control test. The branch is rebased on post-#546 main as requested, and the `lower_cast` conflict was resolved by keeping both sides: #546's explicit `(int)`/`(float)`/`(bool)` cast release arm plus this PR's generalized helper on the string arm (disjoint sites, no double release).

## Tests

- 19 `test_regression_500_*` heap-debug regressions: every leaking shape from the issue (`&`, `%`, shifts, `**`, comparisons, unary minus, int-coerced and mixed-key array indexes, nested masking) plus the clean control shapes (split statement, plain scalar ops, chained mixed adds, call arguments, string coercion, loop-carried accumulators) and an overflow-parity control (`gettype` of a runtime-overflowing multiply still prints `double`).
- 11 of the original 18 tests failed on the unfixed baseline (the canonical repro died of heap exhaustion); the `**` test leaked 1 block/eval on post-#546 main before the fix.
- Full `runtime_gc` suite green (196/196, including the new #546 tests), `--ir-opt` on and off.

While auditing, three pre-existing sibling leaks were confirmed on unfixed main and left out of scope (happy to file/fix separately): the mixed-key array-*write* path leaks its boxed key (plus a read miscompile after mixed-key int writes), the mixed-key `isset` probe leaks key and probe boxes, and a `match`/`switch` subject built from checked arithmetic leaks its subject box.

Fixes #500


mirchaemanuel:
Root-caused the linux-x86_64 CI failure — it is not the fix; it was the `gt ok` assertion in the overflow-parity control test, and the divergence is pre-existing. Fixed in 9c8e3a76e.

The relational consumer `(($a * 2) > 0)` of an overflow-promoted box truncates the float payload through the pre-existing `__rt_mixed_cast_int` before comparing, and the float→int conversion is architecture-divergent on overflow: ARM64 `fcvtzs` saturates to `INT64_MAX` (so `> 0` accidentally holds), while x86_64 `cvttsd2si` yields the integer-indefinite `INT64_MIN` (so it doesn't). Verified differentially under Docker linux/amd64: **unfixed main behaves identically**, both inline and through a stored variable — a pre-existing relational-comparison parity gap in the #397 family (loose equality and `switch` dispatch tag-aware; relational operators on this path still truncate). PHP compares floats numerically, so strictly speaking both targets are wrong and ARM64 only passes by saturation luck.

The control test now asserts the arch-stable promotion behavior (`gettype` → `double\ndouble`) plus heap cleanliness; release-through-comparison stays covered by the non-overflowing `test_regression_500_comparison_releases_checked_operand_box`. Re-verified after the change: all 19 `regression_500` tests green on macOS ARM64 and on Dockerized linux-x86_64.

Happy to file the relational truncation gap as its own issue with the repro and the asm references.

mirchaemanuel:
Filed the relational truncation gap as #548 with the repro, the per-target results table, and the asm references.

PR #560, opened 2026-07-18
fix(ir-lower): lower an object-returning function's unreachable fall-through to a null sentinel

chadmandoo:
### What

An object-returning function whose only implicit fall-through is **unreachable** — e.g. a
`try` whose body `return`s and whose every `catch` `throw`s — still leaves a structural
try/catch join block that needs a value of the object return type for its terminator.

`emit_default_return_value` synthesized that placeholder with a **0-operand `Op::RuntimeCall`**,
which the EIR backend cannot lower for a heap/object return type:

```
unsupported EIR backend feature: runtime_call with 0 operands returning PHP type Object("Conn")
```

### Fix

Add an `IrType::Heap(_) if …Object(_)` arm to `emit_default_return_value` that emits the
null-object sentinel via `Op::ConstNull` (the representation `const_null` already produces for a
non-tagged result) instead of the un-lowerable 0-operand `Op::RuntimeCall`. The join is
unreachable, so the placeholder is never materialized at runtime — it only needs a representation
codegen can lower. This mirrors the existing `Mixed` heap arm directly above it.

### Test

`tests/codegen/exceptions.rs::test_object_return_unreachable_fallthrough_placeholder` — a
`create(): Conn` whose `try` returns and whose `catch` throws. Fails before the change with the
error above; passes after. Full `cargo test` suite green (no regressions).


nahime0:
Maintainer update:

- Merged the current `main` into this branch and resolved the `tests/codegen/exceptions.rs` conflict while preserving both main's newer exception coverage and this PR's regression.
- Replaced the Object-specific `ConstNull` fallback with the CFG-level fix. `lower_try_catch` and `lower_try_catch_finally` now track whether the try body or any catch can actually reach `try.after`; joins with no incoming fallthrough are terminated with `Terminator::Unreachable`.
- Removed the Object-only arm from `emit_default_return_value`, so dead joins no longer synthesize fake typed returns.
- Added a structural EIR regression covering Object and Array heap returns, with and without `finally`, plus end-to-end exception regressions.
- Added the corresponding `[Unreleased]` changelog entry.

Local validation:

- `cargo build`
- `cargo test --lib ir_lower::tests::dead_try_after_joins_are_unreachable -- --exact`
- `cargo test --test codegen_tests codegen::exceptions::` — 50 passed
- Object + Array fixture compiled and ran with `--no-ir-opt` — output `pg|1,2`
- `git diff --check origin/main...HEAD`

Current head: `a53a935be`. The complete supported-target matrix is left to CI.

PR #561, opened 2026-07-18
fix(ir-lower): type an indexed array-literal method-call element by the callee return

chadmandoo:
### What

An indexed array-literal element that is a **method call** — `[$this->factory->link('x')]` — was
typed by the syntactic `Int` default rather than the callee's declared return type. The literal
was stamped `array<int>`, so the returned object got int-cast into it:

```
unsupported EIR backend feature: int cast for PHP type Object("Link")
```

### Fix (two related arms)

1. `array_literal_element_type_for_ir` — add `MethodCall` / `NullsafeMethodCall` /
   `StaticMethodCall` arms that resolve the element type from the callee's declared return type
   (return type is argument-independent for user methods), guarded by
   `materializable_array_element_type` so a `Void`/`Never` return keeps the syntactic fallback
   (no `array_push for Void`). The `Spread` arm also widens an empty `array<never>`/`array<void>`
   element to `Mixed` instead of collapsing the outer literal.
2. `instance_callable_object_class` — add a `PropertyAccess` arm so a property-access receiver
   (`$this->factory`, not just `$this`/`$var`/`new`) resolves through the property's declared
   type. Without it the method-call arm above can't find the receiver's class.

### Test

`tests/codegen/arrays/indexed/heterogeneous.rs::test_indexed_array_literal_property_receiver_method_element_typed_by_return`
— a `row(): Row { return new Row([$this->factory->link('View'), $this->factory->link('Edit')]); }`.
Exercises both arms. Fails before with the error above; passes after. Full `cargo test` green.


nahime0:
Maintainer update:

- Merged the current `origin/main` into the PR branch (`5b4bb16a7`).
- Kept the indexed-array root-cause fix, but aligned it with the current EIR return-type helpers from `main`.
- Applied the same declared-return typing to associative array literal values, fixing the mirrored `hash_set value PHP type Object(...)` failure.
- Extended instance receiver resolution through declared properties, nullsafe property access, chained method results, and static method results; nullsafe calls preserve the possible `null` result before selecting array storage.
- Added the missing `[Unreleased]` CHANGELOG entry.
- Added regression coverage for indexed property receivers, chained-method receivers, static calls, genuinely-null nullsafe receivers, empty generic spreads, and the associative mirror case.

Local validation is green: `cargo build`, `cargo fmt --check`, 5 indexed regressions with `ELEPHC_PHP_CHECK=1`, the associative regression with `ELEPHC_PHP_CHECK=1`, and 34 focused associative-array tests.

CI is fully green on the exact new head `7e733fecfdb9a9f586c7185398ae7410350e427b`: 111 successful CI jobs, 1 expected skip, and no failures. The PR is `MERGEABLE` with a `CLEAN` merge state against current `main`.

PR #563, opened 2026-07-18
test(codegen): cover nullable string strlen ownership

### LINKED ISSUE #485: ?string return of a runtime-built string leaks the box's persisted copy once per call

mirchaemanuel:
# test(codegen): cover nullable string strlen ownership

Closes #485.

## Why this PR is now test-only

The original patch correctly diagnosed the legacy leak: `lower_strlen` coerced a boxed string through `__rt_mixed_cast_string`, whose tag-1 arm returns an owned persisted copy, and then consumed only its length. Switching that read-only consumer to borrowed coercion was a real ownership fix on the original base.

After merging current `main` (`40737447b`), the typed-EIR builtin migration has removed that lowering path. Dynamic `strlen()` now emits `Cast(Str) -> StrLen`; the cast result is an owning temporary and EIR releases it after `StrLen`. The exact #485 repro prints `175` with `HEAP DEBUG: leak summary: clean`.

Keeping the legacy backend patch would therefore be stale. This branch now preserves the current EIR implementation and retains only the useful regression coverage.

## Tests

`tests/codegen/runtime_gc/nullable_string_return.rs` covers:

- one `strlen()` call on a runtime-built `?string`;
- the exact 50-iteration #485 repro;
- use of the boxed value after `strlen()`;
- the nullable branch;
- a `?string` parameter boundary.

Each fixture checks output and a clean heap-debug summary.

## Validation

- `cargo test --test codegen_tests nullable_string_return`
- `ELEPHC_IR_OPT=off cargo test --test codegen_tests nullable_string_return`
- `cargo test --lib strlen_uses_backend_neutral_eir_graph`
- `cargo test --lib mixed_string_cast`
- `git diff --check`

No `CHANGELOG.md` entry is added because the post-merge delta adds regression coverage but no new user-facing behavior.


nahime0:
Updated this PR against current `main`.

What changed:

- Merged `origin/main` (`40737447b`) into the contributor branch; the updated head is `e9dd6ae8e`.
- Resolved the stale lowering conflicts by keeping the current typed-EIR `strlen()` implementation. The legacy `lower_strlen` path patched by the original PR no longer exists on `main`.
- Kept the useful regression coverage only: five heap-debug tests for the exact #485 loop repro, a single call, value reuse after `strlen()`, the nullable branch, and a `?string` parameter boundary.
- Updated the PR title and description to reflect that the final delta is test-only and will close #485 when merged.
- Did not add a `CHANGELOG.md` entry because the rebased PR adds regression coverage but no new user-facing behavior.

The original diagnosis and borrowed-coercion fix were correct on the old base. On current `main`, the same leak is already prevented through the more general EIR ownership path: dynamic `strlen()` lowers to `Cast(Str) -> StrLen`, followed by release of the owning cast temporary.

Validation completed locally:

- `cargo test --test codegen_tests nullable_string_return` — 5/5 passed
- `ELEPHC_IR_OPT=off cargo test --test codegen_tests nullable_string_return` — 5/5 passed
- `cargo test --lib strlen_uses_backend_neutral_eir_graph` — passed
- `cargo test --lib mixed_string_cast` — passed
- `git diff --check` — clean

GitHub now reports the PR as mergeable; the full CI matrix is running on exact head `e9dd6ae8e`.

PR #570, opened 2026-07-19
fix(ir,codegen): autovivify missing parents in nested array writes

### LINKED ISSUE #555: Chained writes through a missing array element fail to autovivify the parent
## Summary

A nested assignment through a missing intermediate array element does not autovivify that element. The write is lost, warnings are emitted for a legal PHP write, and the failed path leaks runtime allocations.

This is the chained-write case explicitly left out of scope by #533.

Reproduced on `main` at `0b26415c9b`.

## Reproduction

```php
<?php
$a = [['x', 'y'], 7];

$a[7][1] = 'patched';

echo $a[7][1] . "\n";
```

## Current behavior

elephc exits normally, but:

- prints an empty line instead of `patched`;
- emits `Undefined array key 7` twice;
- under `--heap-debug`, reports `allocs=12 frees=10 live_blocks=2 live_bytes=88`.

PHP creates the missing intermediate array, stores the value, prints `patched`, and emits no warning.

## Why this is separate from #529 / #553

#529 and #553 address nested writes when the parent element already exists: the leaf write is routed through the parent cell instead of mutating a detached read result.

For `$a[7][1] = 'patched'`, however, lowering first obtains `$a[7]` through a read path. The miss produces a detached null/Mixed value. The leaf setter can mutate that detached value, but there is no fetch-for-write/autovivification step that installs the new container back into the outer `$a[7]` slot, so the update is lost.

The fix therefore needs write-context lookup semantics for intermediate elements, not just another leaf setter change.

## Required semantics and regression coverage

Please cover:

- indexed and string-keyed missing intermediate elements;
- multiple missing levels, for example `$a['x']['y']['z'] = 1`;
- no undefined-key warning for legal autovivifying writes;
- correct COW and alias behavior when the outer array is shared;
- heap-debug cleanliness;
- existing-parent cases from #529 / #553 as controls;
- `--ir-opt` on and off;
- all supported targets.

The implementation must preserve PHP's distinct behavior for incompatible scalar intermediates; it should not blindly convert every non-array parent.

#553 also documents a related residual limitation for 3+ level chains whose existing intermediate is a concrete homogeneous array. Both cases appear to need a fetch-for-write helper that can publish/promote intermediate storage in place. If they require materially different semantics, the concrete-intermediate case should be split into its own follow-up rather than being silently left uncovered.

## Related

- #526 / #533 — null-container read safety
- #529 / #553 — nested writes through an existing parent cell

mirchaemanuel:
# fix(ir,codegen): autovivify missing parents in nested array writes

Fixes #555.

## Symptom

```php
$a = ['x'];
$a[7][1] = 'patched';   // PHP: autovivifies $a[7] as an array, no warning
echo $a[7][1];          // elephc: empty line, "Undefined array key 7" twice,
                        // and the dropped value leaked
```

## Root cause

The parent chain of a nested assignment was lowered as a plain read. A missing element produced a detached `Mixed(null)` box (with an undefined-key warning), and `__rt_mixed_array_set` released the value in its incompatible-target path. Nothing installed a container into the missing slot: there were no write-context lookup semantics for intermediate elements. Additionally, a string key on an indexed payload branched to the same drop path, losing legal writes like `$a[0]['new'] = 'v'`.

## Fix

The parent chain now lowers with fetch-for-write semantics:

- New `__rt_mixed_array_get_for_write` autovivifies missing indexed elements (grow, zero-fill, length update), null gap slots, boxed-null slots, and missing hash keys, installing an empty-array cell into the parent storage and returning the STORED cell (retained) so the three-operand set writes through the parent. Indexed payloads are first normalized via `__rt_array_to_mixed` (COW split + boxed slots, republished into the cell) — which also lands writes through concrete intermediates of 3+ level chains, the residual limitation documented in #553.
- Concrete `array<mixed>`/assoc locals go through a new `__rt_array_ensure_elem_for_write` wrapper plus a local storeback; a literal string key on an indexed local promotes it via `Op::ArrayToHash`.
- `__rt_mixed_array_set` now promotes indexed payloads to hash storage for string keys (new `__rt_mixed_cell_promote_to_hash`) instead of dropping the write, matching PHP array-key semantics.
- Write-context reads emit no undefined-key warning, matching PHP.
- Both AArch64 and x86_64 emitters implemented symmetrically.

## Tests

`tests/codegen/arrays/nested_autovivify.rs` (14 tests): the exact issue repro, append position, null gap and boxed-null parents, assoc missing key, multi-level int and string chains from `[]`, Mixed indexed/assoc receivers, string-key child promotion, existing-parent and shared-array COW controls, and function-return visibility; all assert no warning and heap-debug cleanliness. (The json_decode variant asserts stdout/warnings only — Mixed builtin results stored into locals leak one owner reference on current `main` independently of nested writes, reproducible with a bare read.) 13/13 of the original tests failed before the fix.

## Limitations (documented)

- Incompatible scalar intermediates are intentionally not converted (the write is still dropped; PHP raises `Error: Cannot use a scalar value as an array` — no error machinery added here).
- Negative indexes on dense indexed storage stay unsupported (pre-existing storage model).
- Statically-Mixed key expressions on concrete locals and ArrayAccess-object intermediates keep the previous plain-read path.
- Autovivified gaps remain dense (zero-filled) per elephc's array model, so `count()` can differ from PHP's sparse arrays (pre-existing).

## QA

- Issue repro prints `patched`, zero warnings, `leak summary: clean`, under `--ir-opt=on` and `=off`.
- After rebasing onto current `main` (post-#553, including its extended `nested_mixed_write` suite): `nested_autovivify` 14, `nested_mixed_write` 8, `arrays::` 366, `mixed` 262 — all green. Pre-rebase full slices: `nested` 155, `json` 445, `cargo test --lib` 930 (one pre-existing environment-dependent DNS test failure unrelated to this change, fails identically on pristine main). No new clippy warnings.


nahime0:
Maintainer follow-up after bringing this branch current with `main`:

- Merged `origin/main` at `40737447b` into the contributor branch (`1d4073f21`). The only conflict was `src/ir/mod.rs`; the resolution preserves both main's `PhpTypePredicate` export and this PR's fetch-for-write integration.
- Re-audited the implementation against #555. The fix still addresses the root cause: intermediate elements are fetched in write context and installed into parent storage, rather than mutating a detached read result. No P0/P2 correctness issue was found in the covered missing-parent semantics.
- Aligned the PR with the typed EIR runtime boundary introduced by #582: the magic `Immediate::I64(0x4657)` dispatch is gone, replaced by `RuntimeCallTarget::ArrayFetchForWrite` with a typed signature, stable textual EIR identity, and typed backend dispatch.
- Added an EIR regression that pins `runtime.array.fetch_for_write`.
- Added an explicit CLI regression that runs the issue repro with both `--ir-opt=on` and `--ir-opt=off`, checking output, warning suppression, and heap cleanliness. The focused autovivification module is now 15 tests.
- Updated the stale comment on the three-level concrete-intermediate regression: that path is now covered by fetch-for-write normalization.

Local validation on the new head `ce515a915`:

- `cargo build`
- `cargo test --test codegen_tests arrays::nested_autovivify` — 15/15
- `cargo test --test codegen_tests arrays::nested_mixed_write` — 8/8
- typed EIR lowering regression — 1/1
- `git diff --check`
- assembly-comment alignment check for the touched codegen/runtime files

GitHub now reports the PR as mergeable. Build/archive is green on macOS ARM64, Linux ARM64, and Linux x86_64; builtin-doc and image-sync checks are also green. The downstream test matrix is currently queued because repository runners are not taking jobs, with no failures reported yet. The remaining merge gate is therefore a green result from that queued matrix.

The documented scope limits (incompatible scalar intermediates, dynamic Mixed keys on concrete locals/ArrayAccess intermediates, and typed non-boxed associative entries) remain unchanged and are separate from the missing-parent case fixed here.

Issue #508, opened 2026-07-09
Checker narrows ?T after instanceof but not after an === null guard with early exit


mirchaemanuel:
## Summary

The type checker narrows a `?T`-typed value after an `instanceof` guard, but
not after an `=== null` guard with an early return/exit on the null branch —
even though both eliminate the `null` case for any code that follows.

## Environment

- elephc v0.26.0 (release, commit `892ff162`), macOS ARM64
- PHP has no compile-time type checker to compare against here; this is
  purely about elephc's static flow-sensitive narrowing.

## Repro

```php
<?php
// mk(): ?P
$p = mk(1);
if ($p === null) {
    exit(1);
}
want($p);   // want(P $p): void
```

```
error: Method ... parameter expects Object("P"), got Union([Object("P"), Void])
```

The equivalent code narrows correctly with `instanceof`:

```php
$p = mk(1);
if (!($p instanceof P)) {
    exit(1);
}
want($p);   // compiles
```

## Expected

`=== null` (combined with an unconditional exit/return/throw on the null
branch) should narrow `?T` to `T` for the following code, exactly like the
`instanceof` guard already does.

## Possibly related

#494 ("Inferred return types erase nullability for object/null ternary and
match merges") is in the same neighborhood (nullability handling around
`Void`/`null` merges in the checker) but is about a different point in the
pipeline — inferred *return type* construction at a ternary/match merge, not
narrowing a variable's type after a runtime null check. Flagging in case the
two share a root cause in how `Void`/`null` is modeled, but this is a
distinct symptom.

## Workaround

Replaced the `=== null` guard with `instanceof P` (negated) where the
narrowing was required.

## Notes

Found building a pastebin single-binary app (`--web`) for the Discord
contest, guarding a nullable finder-method result before use.

mirchaemanuel:
Update after re-testing on current `main` `84095e0ee` (local debug build, macOS ARM64): the checker side of this is fixed — the `=== null` + `exit` guard now narrows `?P` to `P` and the repro compiles. But the runtime result is wrong: the narrowed value that reaches `want()` is garbage.

Standalone repro:

```php
<?php

class P {
    public int $v;
    public function __construct(int $v) { $this->v = $v; }
}

function mk(int $x): ?P {
    if ($x > 0) {
        return new P($x + 6);
    }
    return null;
}

function want(P $p): void {
    echo $p->v, "\n";
}

$p = mk(1);
if ($p === null) {
    exit(1);
}
want($p);
[...]

PR #583, opened 2026-07-21
fix(ir_lower): widen mismatched array/array branch merges to boxed mixed

### LINKED ISSUE #549: match arms returning arrays with different element types merge to a single element type: reading the other arm's array segfaults
## Summary

When the arms of a `match` produce **indexed arrays with different element
types** (e.g. `array<int>` vs `array<string>`), the merged result type keeps a
single element type instead of widening the elements to boxed `mixed`.
Reading the array produced by the *other* arm through that merged type is a
miscompile. On current `main` it crashes outright: the int-array arm's
elements are read as string pointers → SIGSEGV.

This was originally spotted as a follow-up during the work on #488 / PR #492
(https://github.com/illegalstudio/elephc/pull/492#issuecomment-4919511824):
at the time (v0.26.x) the repro read wrong data through the merged type. On
current `main` the same repro segfaults, so filing it as its own issue.

Heterogeneous *literal* array payloads already widen to boxed `mixed`
(docs/php/types.md) — it's the cross-arm merge that doesn't.

## Environment

- elephc current `main` (`84095e0ee`), debug build, macOS ARM64
- PHP 8.4: unaffected

## Repro

```php
<?php

$r = match($argc) {
    1 => [1, 2],
    default => ["a", "b"],
};
echo $r[0], "\n";
echo $r[1], "\n";
```

| Run | PHP 8.4 | elephc `main` |
|---|---|---|
| no args (`$argc = 1`, int arm) | `1` `2` | **SIGSEGV (exit 139), no output** |
| one arg (`$argc = 2`, string arm) | `a` `b` | `a` `b` (heap clean) |

The string arm works because the merged element type matches it; the int arm
is the one read through the wrong element type.

## Expected

The merge of `array<int>` and `array<string>` (in `match`, and presumably
ternary/if-else value merges as well) should widen the element type to boxed
`mixed`, like heterogeneous array literals already do.

## Notes

Likely fix direction: `wider_type_for_merge` (src/ir_lower/expr/mod.rs)
currently has no array/array widening rule for mismatched element types, so
one arm's element type wins; merging to an array-of-Mixed (with the same
boxing the literal path uses) would make both arms readable.

mirchaemanuel:
## Summary

Fixes #549 — `match` arms (and ternary / `??` / `??=` merges) returning indexed arrays with different element types merged to a single element type; reading the other arm's array through the merged type misinterpreted the slot bytes (SIGSEGV on current `main` when the int arm is read as strings).

## Root cause

Two cooperating gaps:

- `wider_type_for_merge` resolved `(Array, Array)` and `(AssocArray, AssocArray)` as `right.clone()` — the last arm's element type won wholesale.
- `coerce_value_for_temp` performed no container conversion, so the other arm's runtime array was stored raw and merely relabeled; element reads then decoded the wrong slot width/content (int slots read as string `{ptr,len}` pairs → fault in `__rt_str_persist`).

## Fix

- `wider_type_for_merge` now merges array/array and assoc/assoc **elementwise**, reusing the heterogeneous-literal rules (`merge_ir_indexed_element_type` / `merge_ir_assoc_value_type`): equal reprs keep the type, `Never`/`Void` yields the other side, any mismatch widens to boxed `mixed`.
- The merge-temp store path routes containers through a new `widen_container_value_for_temp` helper: typed containers stored into Mixed-payload temps box their slots via the existing `ArrayToMixed` / `HashToMixed` ops. Borrowed and provisionally-owned sources are retained first so the conversion's copy-on-write split rewrites a private copy (per the PR #489 ownership rules); whole-boxed `?array` cells flowing through `??` unbox with the owned-payload coercion before converting when the cell does not own its payload, so a caller's live array is never rewritten in place.
- `lower_array_to_mixed` / `lower_hash_to_mixed` pass null and in-band null-container sentinels through unconverted on both targets (call-site guard, issue #533 convention), so a missed read forwarded by an arm is never dereferenced.

This fixes match, full ternary, short ternary, `??`, and `??=` in one place — they share the same merge/store helpers. Target-neutral: AArch64 and x86_64 emissions for both guards.

## Tests

18 regression tests in `tests/codegen/control_flow/{match_expressions,ternary,nulls}.rs`: both arm orders for match and ternary, assoc value-type mismatch, empty-array arm, nested arrays, variable (borrowed) arms, missed-read sentinel forwarding, and `??` with a `?array` value side — owning and borrowed cells, including a live caller array passed twice and a 20-iteration loop — with heap-clean assertions under `--heap-debug`. Focused neighbor suites are green locally on macOS ARM64 (match 151, ternary 45, coalesce 63, nulls 32, plus array/hash/assoc/spread/optimizer filters, ~1,600 tests). The issue's repro prints PHP-parity output on both arms.

## Notes

- The sentinel null-identity gap for merged missed reads (`$r === null` where PHP sees `null`) is pre-existing (#508 family) and unchanged here — this PR guarantees crash-freedom and `??`-fallback parity for that shape.
- x86_64 emissions mirror in-file idioms and are compile-verified locally; counting on CI for the Linux matrix.


mirchaemanuel:
Follow-ups from the adversarial review of this fix, all verified pre-existing on `main` (none are regressions of this PR): #585 (unguarded null-container sentinel in `__rt_mixed_array_get` on ternary-forwarded missed reads), #586 (merge temp leaks when read through `?? default`), #587 (checker types heterogeneous merges as `mixed`, rejecting valid array uses that now run correctly at runtime).

nahime0:
Maintainer follow-up after the current-main refresh:

- Merged current `origin/main` (`40737447b`) into the branch in `8b81946cd`, bringing the PR across the large EIR migration without conflicts.
- Re-reviewed the implementation against #549. The fix is a root-cause contract correction: merge typing widens indexed/associative array payloads elementwise, merge-temp materialization performs the matching `ArrayToMixed` / `HashToMixed` conversion, and borrowed/provisional owners are retained so COW preserves live source arrays. The shared merge/store path covers `match`, ternary, `?:`, `??`, and `??=`.
- Updated `docs/php/types.md` to remove the now-fixed indexed-array incompatibility and kept only the existing `int`/`bool` divergence.
- Added the #549 entry under `[Unreleased]` and removed the stale array-divergence claim from the earlier #488 changelog entry (`df2608fe6`).
- Confirmed #585, #586, and #587 remain separate pre-existing follow-ups; #587 still represents the checker-side array typing gap and does not invalidate the EIR/layout fix in this PR.

Local validation on the refreshed branch:

- `cargo build` — passed
- 20 focused codegen/ownership regression tests — 20 passed, including both arm orders, associative arrays, COW preservation, heap-debug balance, sentinel passthrough, ternary/short-ternary, and nullable-array `??`
- assembly-comment alignment — passed
- `git diff --check origin/main...HEAD` — clean

GitHub now sees head `df2608fe6`. Fresh CI is currently queued/in progress on that exact SHA; once the required matrix is green, I see no remaining blocker to merge.

PR #591, opened 2026-07-22
fix(runtime): guard boxed-mixed array get/set against the null-container sentinel

### LINKED ISSUE #585: Missed array read forwarded through a ternary merge segfaults in __rt_mixed_array_get (unguarded null-container sentinel)
## Summary

A missed indexed-array read (`$rows[5]`) forwarded as one arm of a ternary produces the in-band null-container sentinel; reading the merged result (`$r[0] ?? "none"`) crashes in `__rt_mixed_array_get`, which dereferences the sentinel. The hash sibling `__rt_mixed_hash_get` is already guarded per the issue #533 convention and prints PHP-parity output for the equivalent shape.

Found during the adversarial review of PR #583 (issue #549) and verified independent of it: the homogeneous-arm variant (`$rows[5]` vs `[3, 4]`) crashes identically, and its emitted IR contains no `array_to_mixed`/`hash_to_mixed` and bypasses `wider_type_for_merge` entirely (the ternary temp is typed through the nullable-union path and whole-boxes with `mixed_box`), so no #583 code path executes.

## Environment

- elephc current `main` (`07391e965`), debug build, macOS ARM64 (also reproduces with PR #583 applied, and in both `--null-repr` modes)
- PHP 8.4: unaffected

## Repro

```php
<?php
$rows = [[1, 2]];
echo $r[0] ?? "none", "\n";
echo "done", "\n";
```

| Run | PHP 8.4 | elephc `main` |
|---|---|---|
| no args (missed-read arm) | `Warning: Undefined array key 5` then `none` `done` | warning, then **SIGSEGV (exit 139)** |
| one arg (literal arm) | `a` `done` | `a` `done` |

lldb: the fault is in `__rt_mixed_array_get` dereferencing the null-container sentinel (`0x7ffffffffffffffe`).

## Expected

The sentinel should be treated as an absent container: `$r[0] ?? "none"` prints `none` and execution continues, as PHP does (and as the guarded `__rt_mixed_hash_get` sibling already behaves).

## Notes

Same family as #508 / #580 / #581. Likely fix direction: guard `__rt_mixed_array_get` against null / `NULL_SENTINEL` receivers per the #533 convention, mirroring `__rt_mixed_hash_get`.

mirchaemanuel:
## Summary

A missed indexed read (`$rows[5]`) forwarded as one arm of a ternary whole-boxes into a
`mixed` cell whose payload pointer is the in-band null-container sentinel
(`NULL_SENTINEL = 0x7ffffffffffffffe`) rather than a real array pointer. `$r[0] ?? "none"`
then crashed in `__rt_mixed_array_get`, which dispatched on the cell tag, loaded the
container pointer from `[cell + 8]`, and guarded it only with a zero check (`cbz`/`test`) —
which the non-zero sentinel slips past. The next load dereferenced the sentinel and
segfaulted (exit 139). The sibling writer `__rt_mixed_array_set` had the identical hole on
`$r[0] = …`.

The ternary's nullable-union path bypasses `wider_type_for_merge` (the `match` merge path
does not), so the sentinel reaches the helpers in-band. PHP 8.4 warns and yields `null` /
drops the write; elephc crashed.

## Root cause

Both helpers load the container payload from `[cell + 8]` in each of their indexed / assoc /
object paths and guarded it against a zero pointer only. A missed container read boxed into a
`mixed` stores the sentinel (not zero, not a valid pointer) in that slot, so the guard is
bypassed and the following dereference faults.

## Fix

`__rt_mixed_array_get` and `__rt_mixed_array_set` now route a null **or** null-container
sentinel payload to their existing absent-container paths via the shared
`emit_branch_if_null_container` helper (issue #533 convention: guards live inside the runtime
helpers), on both aarch64 and x86_64. The read returns `Mixed(null)` with PHP's undefined-key
warning; the write is dropped, matching elephc's existing null-target behavior.

## Test plan

New regressions in `tests/codegen/regressions/arrays.rs`:
- exact issue repro (read yields the default + undefined-key warning)
- present-arm control (valid boxed indexed read unaffected)
- indexed write into the sentinel container (no crash)
- heap-clean checks for both the read and the dropped write

Red→green: the four sentinel-path tests crash before the fix and pass after. Focused slices
(`ternary`, `mixed_array`, `coalesce`, `null_container`, `missed`, `regressions::arrays`,
`null_sentinel`, `json`, `stdclass`, `spl_`, `assoc`) pass: 958 tests, 0 failures.

Fixes #585


mirchaemanuel:
While validating this fix I found an adjacent pre-existing crash on the append path (`$r[] = ...` on the same sentinel-carrying mixed cell) — it does not go through the helpers guarded here and needs auto-vivify semantics, filed separately as #592.

mirchaemanuel:
Picking this up. CI is red on the current head `bd354a5ee`; it was green on `9068d92e6` (2026-07-22), and the failures show on all three architectures (macos-aarch64, linux-x86_64, linux-aarch64).

The failures cluster in the ArrayAccess family: `codegen::spl::interfaces::test_array_access_subscript_read_write_isset_unset`, `codegen::spl::classes::test_phase4_spl_fixed_array_runtime_methods`, `codegen::spl::classes::test_phase4_spl_fixed_array_serialization_and_from_array_helpers`, `codegen::spl::heaps::test_spl_object_storage_bulk_operations`, `codegen::spl::decorators::test_caching_iterator_full_cache_array_access_and_flags`, plus several `codegen::io::streams::test_phar_oop_*` failing with `Fatal error: Uncaught Error: Call to a member function getContent() on null`.

One concrete data point: `test_array_access_subscript_read_write_isset_unset` expects `SGvE1UE` and gets `SSE1UE`.

All of the above is read off the CI logs; we have not reproduced it locally yet and are not claiming a cause.

mirchaemanuel:
Pushed a fix for the CI failures on `bd354a5ee`. There were **two** independent regressions in that commit, not one — the second was not in what I reported earlier tonight, I found it only by running the full codegen suite rather than the failing clusters.

Both share the same cause: the new warn-on-missing operand gave boxed-Mixed subscript reads a third operand, and two consumers still interpreted three operands as a write.

**1. ArrayAccess reads dispatched `offsetSet`.** `try_lower_array_access_runtime_call` classified the method purely by operand count (`3 => "offsetSet"`), so `echo $obj[$k]` called the setter with the warn flag as its `$value`. That function runs at `lower_inst.rs:2050`, before the `result_php_type != Void` disambiguation added at `:2053`, so that guard never covered ArrayAccess receivers. This is what broke SplFixedArray, SplObjectStorage, CachingIterator, Phar OOP and the `test_array_access_subscript_read_write_isset_unset` trace (`SGvE1UE` → `SSE1UE`: the read stores `(string)true` into `$stored`).

Reads are now identified structurally by having a result value — subscript writes lower through `emit_void` and carry none — and the flag operand is stripped before dispatch so `offsetGet` keeps its one-argument signature. I deliberately did not reuse `result_php_type != Void` here: `codegen_repr()` collapses null and void, so a read whose declared `offsetGet` return type has no runtime representation would still have misdispatched silently.

**2. `[$a, $b] = $narrowed` failed to compile.** `lower_mixed_array_runtime_get` now requires the third operand, but only the array-access producer was taught to supply it. `lower_list_unpack` also emits a boxed-Mixed read whenever the source is not an Array/Hash IR type — precisely a narrowed `?array` — and still emitted two operands, so `codegen::arrays::list_unpack::test_null_guard_{break,continue}_narrows_list_unpack_rhs` died with `CodegenIrError: runtime_call missing operand 2`. It now supplies the flag through the same helper. Verified byte-identical to `php` on a short source: `Warning: Undefined array key 1` then NULL.

**One thing worth your judgement.** The dispatch into `lower_mixed_array_runtime_get` is by *type*, not by symbol name — `lower_binary_runtime_call` (`lower_inst.rs:2479`) routes any `RuntimeCall` with a `Mixed | Union` receiver and non-void result there. So the set of producers that must supply the flag is not greppable, and I found two of two by looking. I did not audit all 36 `Op::RuntimeCall` emission sites, and I'm flagging that as unverified rather than claiming it clean. The failure mode is loud (`runtime_call missing operand N` at compile time), which is why I kept the strict `expect_operand` rather than defaulting the flag at the consumer: a tolerant default would be correct for plain reads but wrong for `isset()` / `??`, turning a compile error into a spurious-warning bug in exactly the paths this PR touches. If you'd rather have the tolerant default anyway, it's a two-line change.

**Side effect, deliberate:** the structural discriminator also makes `offsetGet(): void` dispatch correctly ("G") where the last green head hard-errored `unknown method Box::append`. I judged removing a silent-miscompile path worth it, but a checker-level diagnostic may be what you actually want there.

No CHANGELOG entry and no new tests: both regressions were introduced and fixed inside this unmerged PR so nothing user-facing changed, and the tests that caught them are the pre-existing ones that turned red.

**Verification**, all on macOS aarch64 and all against the exact pushed commit:

| suite | result |
|---|---|
| `codegen::spl` | 247 passed, 0 failed |
| `codegen::io::streams` | 347 passed, 0 failed, 8 ignored |
| `codegen::regressions::arrays` | 88 passed, 0 failed |
| `array_access` filter | 26 passed, 0 failed |
| `list_unpack` filter | 21 passed, 0 failed |
| `cargo test -p elephc --lib` | 751 passed, 0 failed |
| `error_tests` | 1109 passed, 0 failed |
| `lexer_tests` / `parser_tests` | 208 / 375 passed, 0 failed |

`codegen::regressions::arrays` covers both the 7 tests `bd354a5ee` added and the 6 issue-#585 regressions, so the original fix is intact.

Two caveats I'd rather state than have you find: the full `cargo test --test codegen_tests` was still running when I pushed — at 3685 passed with a single failure, `test_eval_dispatches_gethostbyname_builtin_call`, which is environmental (this machine's DNS resolves NXDOMAIN to 127.0.0.1) and reproduces identically on an untouched `bd354a5ee` worktree. And Linux x86_64 / aarch64 were not run locally; neither fix is target-specific — both live in target-neutral lowering, not in assembly — so CI is the check for those.

Unrelated pre-existing bug found while validating, reproduces byte-identical on the last green head so it predates this branch: `$obj[$k] ??= v` on an ArrayAccess receiver calls `offsetGet` where PHP calls `offsetExists` (elephc `GSG`, php `ESG`). Happy to open an issue if useful.

PR #593, opened 2026-07-22
fix(ir_lower): release owned condition operands after truthiness coercion

### LINKED ISSUE #586: Reading a match-merge result through `?? default` leaks the merged container on every iteration
## Summary

When the result of a `match` branch merge is read through the null-coalescing form (`$r[0] ?? default`), the merged container leaks on every loop iteration. Direct element reads (`$probe = $r[0];`) of the same merge are heap-clean, so the imbalance is specific to the `??`-read path over a merge temp.

Found during the adversarial review of PR #583 (issue #549) and verified pre-existing: it reproduces with **homogeneous** arms (below), which never execute any code path #583 touches.

## Environment

- elephc current `main` (`07391e965`), debug build with `--heap-debug`, macOS ARM64 (also reproduces with PR #583 applied)
- PHP 8.4: unaffected

## Repro

```php
<?php
$count = 0;
for ($i = 0; $i < 20; $i++) {
    $r = match($i % 2) {
        0 => [1, 2],
        default => [3, 4],
    };
    $count = $count + ($r[0] ?? 0 ? 1 : 0);
}
echo $count;
```

```
$ elephc --heap-debug leak.php && ./leak
20HEAP DEBUG: allocs=82 frees=62 live_blocks=20 live_bytes=992 peak_live_bytes=1216
HEAP DEBUG: leak summary: live_blocks=20 live_bytes=992
```

Stdout is correct; exactly one block per iteration stays live (the merged container).

## Expected

`leak summary: clean` — the merge temp should be released after the `??` read like it is after a direct element read.

## Notes

- With heterogeneous arms (`[1, 2]` vs `["a", "b"]`, post-#583 widening) the same loop leaks 30 blocks instead of 20 — the converted arrays carry heap-boxed cells — so the widening only amplifies the leaked bytes, it does not cause the leak.
- Also reproduces when the arms are heterogeneous **literals** (which widen through the long-standing literal path), so it is not tied to the merge-widening conversion at all.

mirchaemanuel:
## Summary

Reading a value through the null-coalescing form (`$x[0] ?? default`) and using the result directly as a condition leaked the boxed `Mixed` coalesce temporary on every evaluation (issue #586). The issue's loop repro leaks one heap block per iteration; stdout is correct.

Bisection shows the leak is **not** tied to the `match` merge, to ternary merges, or to element-type widening (issue #549): a plain indexed-array local read through `?? default` leaks just the same, and the trigger is that the coalesce result is consumed **as a condition**. Storing it in a local first is already heap-clean, and varying the array size keeps the leaked block at the fixed box size — so the leaked allocation is the coalesce box, not the container.

## Root cause

`LoweringContext::truthy` emits `Op::IsTruthy`, which reads its operand and yields a fresh, non-aliasing boolean, but it never releases the operand. When the operand is an owned temporary — most visibly the boxed `Mixed` produced by `lower_null_coalesce`'s `take_owned_temp` — nothing releases it after the truthiness test, so it leaks. Every condition site that fully consumes its operand was affected: `if`, ternary, `while`/`do-while`/`for`, short-circuit `&&`/`||`, negation `!`, and magic `__isset`.

## Fix

Add `LoweringContext::truthy_consuming`, which computes truthiness and, when this path owns the input (`value_is_owning_temporary`), releases it through the type-gated `release_if_owned` — the same ownership contract the scalar-coercion casts already use for `(bool)`/`(int)`/`(float)`. Route every condition site that discards its operand through it. `lower_short_ternary` keeps using `truthy` because PHP `?:` forwards the original value when truthy, so releasing there would double-free.

## Test plan

New heap-debug regression tests in `tests/codegen/runtime_gc/regressions.rs` (`$i`/`$argc` to defeat constant folding):

- the exact issue repro (`match` arms + ternary), plus ternary-merge and plain-array siblings;
- `if`, `&&`, and `!` condition siblings;
- controls: coalesce result stored in a local then tested, and a direct element read in a ternary condition, guarding against double frees.

6 leaking shapes were red before the fix and the 2 controls were already green; all 8 pass after. Focused slices green: coalesce/ternary/logical/truthy/isset (154), match/heap/leak/mixed/ownership/nulls (546), ir_lower lib (28), loop/control-flow (445), negation/switch/magic_isset/bool (114).

Fixes #586


mirchaemanuel:
While bisecting this I found two adjacent pre-existing bugs (both reverified on current main, independent of this fix — no `??` involved): #594 (self-referential array literal reassignment reads freed storage → garbage values / non-terminating `while`) and #595 (assoc-array local rebound to a fresh literal in a loop leaks the previous hash each iteration; the indexed sibling is clean).

nahime0:
Follow-up changes have been added to complete the ownership fix and refresh the branch:

- extended the fix to non-short-circuiting logical `xor`; `lower_truthy_bool` now releases an owning temporary after canonical truthiness conversion when the result does not alias the input, while preserving PHP integer truthiness semantics;
- added a heap-debug regression with owned `??` temporaries on both XOR operands. Before the fix it left 40 live blocks across 20 iterations (two per iteration); it is now heap-clean;
- merged current `main` (`0fd01c162`) into the branch without conflicts.

Validation on head `683f90afe`:

- `cargo build`
- 7 focused coalesce/condition ownership regressions
- XOR heap-debug regression with `ELEPHC_IR_OPT=off`
- existing logical-XOR semantic and EIR backend smoke tests
- `git diff --check`

All passed.

PR #600, opened 2026-07-23
fix(codegen): auto-vivify appends onto a null-container sentinel mixed cell

### LINKED ISSUE #592: Appending to a boxed-mixed null-container sentinel segfaults in the inline append path instead of auto-vivifying
## Summary

Appending (`$r[] = "z"`) to a `mixed` cell whose payload is the null-container sentinel
(`NULL_SENTINEL = 0x7ffffffffffffffe`, produced by a missed indexed read forwarded through a
ternary merge) segfaults (exit 139). PHP 8.4 auto-vivifies the null into a fresh `["z"]` array
and continues.

Found while fixing #585 (PR #591): the append path does **not** go through the
`__rt_mixed_array_get` / `__rt_mixed_array_set` helpers guarded there, so it still crashes
after that fix.

## Repro

```php
<?php
$rows = [[1, 2]];
$r[] = "z";
echo "done", "\n";
```

Run with no arguments (`$argc == 1`, missed-read arm):

- **elephc** (main `707cda6b5`, with or without PR #591): `Warning: Undefined array key 5`, then SIGSEGV (exit 139).
- **PHP 8.4**: same warning, then `done` (and `$r === ["z"]`).

The literal arm (run with one argument) works.

## Analysis

The crash is in the **inline** append lowering emitted into `eir_main`, not in a runtime
helper: lldb shows the fault at `ldur x1, [x0, #-0x8]` with `x0 = 0x7ffffffffffffffe`,
immediately before a call to `__rt_array_to_mixed` — the append path reads the container
header inline (no null/sentinel guard) before converting. Observed on macos-aarch64; the
x86_64 lowering emits the same unguarded header read.

Per the #533 convention the plain null/sentinel guard belongs where the raw pointer is
dereferenced — but note a guard alone is not enough here: PHP semantics require
**auto-vivification** (missed/null container + append → fresh single-element array stored
back into the cell), similar in spirit to the autovivify work in #555 / PR #570.

## Related

- #585 / PR #591 — sibling get/set helpers, now guarded (this path is distinct).
- #533 — null-container sentinel guard convention.
- PR #583 — pending call-site guards for `__rt_array_to_mixed` on the match/ternary **merge**
  path; the append path reads the header at a different site and needs its own guard plus the
  auto-vivify semantics.

mirchaemanuel:
Fixes #592.

## Root cause

A missed indexed read forwarded through a ternary/branch merge boxes the null-container sentinel (`0x7ffffffffffffffe`) into a `mixed` cell tagged as an indexed array. Appending to that cell (`$r[] = v`, lowered to `Op::MixedArrayAppend`) unboxes the receiver, accepts the tag-4 payload, and reads the array header inline (`ldur x1, [x0, #-8]` on aarch64, `mov rsi, [rdi - 8]` on x86_64) before calling `__rt_array_to_mixed`. The only pre-dereference guard was a plain-zero check (`cbz x1` / `test rdi, rdi`), so the sentinel passed through and the header read faulted at `sentinel - 8` (exit 139). This append path calls `__rt_array_to_mixed` directly rather than through `lower_array_to_mixed`, so the call-site sentinel guards added in #583 did not cover it.

PHP 8.4 instead auto-vivifies the null into a fresh single-element array and continues: for the repro it warns `Undefined array key 5` on the missed read, then prints `done` with `$r === ["z"]` (no warning on the append).

```php
<?php
$rows = [[1, 2]];
$r[] = "z";
echo "done", "\n";
```

Run with no arguments, this segfaulted (139); it now prints `done`.

## Fix

`lower_mixed_array_append` (aarch64 + x86_64) now guards the unboxed payload for both the null pointer and the in-band null-container sentinel before the header dereference, per the issue #533 convention. For a container-shaped tag whose payload is null or the sentinel — indexed (4), associative (5), or null (8) — it auto-vivifies: it allocates a fresh empty indexed array (`__rt_array_new(0, 8)`, matching `__rt_mixed_new_empty_array_cell`), transfers that reference into the Mixed cell, retags the cell as an indexed array, and appends at index 0 through the shared `__rt_mixed_array_set` tail. Scalars whose payload happens to be zero or equal to the sentinel value (`PHP_INT_MAX - 1`) are excluded by the tag check and keep the existing drop behavior.

Ownership is balanced (the cell owns the fresh array; the array owns the appended value; the replaced sentinel/null payload holds nothing to release), so the auto-vivify path stays heap-clean. Both supported architectures are handled in the same change.

## PHP semantics

Matches PHP's `$x[] = v` on `null`: the null container auto-vivifies to a fresh `[v]` array, the missed read still warns, and the append itself does not warn. A real-null Mixed cell (`$r = cond ? null : [1]; $r[] = "z";`) also auto-vivifies correctly through the same tag-8 route.

## Test plan

New `tests/codegen/arrays/mixed_append_autovivify.rs`:
- the exact issue repro (prints `done`);
- the `["z"]` read-back (length 1, index 0);
- a heap-clean sentinel variant asserting the missed-read warning is still emitted;
- an associative missed-read sibling;
- a multi-append case that exercises array growth;
- a real-array Mixed-cell control guarding against a double-conversion regression.

Fixtures make the taken ternary arm `$argc`-dependent to defeat constant folding. Focused slices run green: `arrays::mixed_append_autovivify`, `nested_autovivify`, `array_to_mixed`, `append`, `push`, `sentinel`, `null_container`, `mixed_array`, `regressions::arrays`, `ternary` (~292 tests, 0 failures). x86_64 coverage relies on CI (no local cross-assembler).


mirchaemanuel:
While validating this fix I found two adjacent pre-existing bugs (both reverified on current main + this branch): #601 (`implode()` over a boxed-mixed indexed array leaks one block per appended element — the appends themselves are balanced) and #602 (`count()` on a raw sentinel cell still segfaults where PHP raises a TypeError — this PR's auto-vivify only masks it when an append happens first).

nahime0:
I updated this PR after merging the latest `main` and completed the Mixed/null-container write path beyond the original crash site.

Changes now included:

- canonicalize zero/sentinel array and hash payloads to a real `Mixed(null)` while boxing;
- share existing-cell array autovivification across append, set, and fetch-for-write paths;
- route Mixed append through a single runtime helper for both supported codegen backends;
- support real associative-array append with PHP's next automatic integer key;
- keep defensive zero/sentinel guards for legacy array-shaped Mixed cells;
- avoid the previous duplicate indexed-array conversion before append;
- add regressions for pure null append, direct set after a missed read, nested write from null, and associative Mixed append.

Focused validation completed:

- `cargo build`
- macOS/AArch64: Mixed append autovivify (8/8), nested autovivify (17/17), nested Mixed write (8/8)
- Linux/x86_64: Mixed append autovivify (8/8)
- `git diff --check`
- assembly comment/alignment checker on all touched codegen/runtime files

Published head: `4d61f4cffdd67ee6e2a911f919a874c60db637ce`.

GitHub Actions for that exact head is currently queued.

Issue #528, opened 2026-07-10
Rebinding a loop-grown array local after pushing it into an outer array leaks one block per rebind


mirchaemanuel:
## Summary

Rebinding a loop-grown array local to `[]` after pushing it into an outer array leaks exactly **one 40-byte block per rebind** (N outer iterations → N−1 leaked blocks). The output is correct; only memory is lost.

## Environment

- elephc main `e63d5d337` (v0.26.0-158), macOS arm64 (Apple Silicon); PHP 8.4.23 as behavioral reference.

## Repro

```php
<?php
$a = [];
for ($i = 0; $i < 6; $i++) {
    $mid = [];
    for ($j = 0; $j < 3; $j++) { $mid[] = ['x', 'y']; }
    $a[] = $mid;
}
echo count($a) . "\n";
```

```
elephc --heap-debug bisect_b.php && ./bisect_b
```

## Observed

```
6
HEAP DEBUG: leak summary: live_blocks=5 live_bytes=200
```

**Scaling law (measured)**: N outer iterations → N−1 leaked blocks, 40 bytes each. 3 iters → `live_blocks=2 live_bytes=80`; 6 → `5/200`; 10 → `allocs=154 frees=145 live_blocks=9 live_bytes=360`.

## What is required to trigger it (negative results, all clean on pristine main)

- a single outer iteration (no rebind): clean;
- an inner loop with exactly one push (`$mid = []; $mid[] = ['x','y'];`): clean;
- assigning a literal instead of growing (`$mid = [['x','y']];`): clean.

The leak needs all three of: an inner **grow loop with ≥ ~3 pushes** into `$mid` (enough to force a storage reallocation), then `$a[] = $mid`, then **rebinding `$mid = []` on the next outer iteration** — one stale block survives per rebind event.

## Expected

Leak-free: after `$a[] = $mid` transfers/retains the final storage, the `$mid = []` reassignment should release everything the local still owns.

## Notes

- Suspected mechanism (behavioral evidence only, not traced to an emitter): growing `$mid` via repeated `[]=` reallocates its storage; after the push retains the final pointer, the rebind's release path drops all but one of the blocks associated with the grown array — consistent with a pre-grow allocation or an ownership-transfer bookkeeping miss on the push+reassign sequence.
- Dynamic strings are not required (the repro uses only literals).
- Reproduces byte-identically on pristine main and on the #524 branch (the repro contains no chained reads).
- Found while regression-testing #524.

[...]

PR #603, opened 2026-07-23
fix(codegen): release the boxed value an associative-array local owns when rebound in a loop

### LINKED ISSUE #595: Reassigning an associative-array local to a fresh literal in a loop leaks the previous hash every iteration
## Summary

Reassigning an associative-array local to a fresh literal inside a loop leaks the previous hash
on every iteration. Stdout is correct; the heap grows linearly with the iteration count. The
same pattern with an **indexed** array literal is heap-clean, so the release-on-rebind exists
for indexed arrays but is missing (or mistyped) for hashes.

Found while working on #586 (PR #593); it is independent of that fix — plain direct read, no
`??` involved.

## Repro

```php
<?php
$count = 0;
for ($i = 0; $i < 20; $i++) {
    $m = ["k" => $i + 1, "j" => "x"];
    $count = $count + $m["k"];
}
echo $count, "\n";
```

Compiled with `--heap-debug` (main `707cda6b5`, macos-aarch64):

```
210
HEAP DEBUG: allocs=162 frees=142 live_blocks=20 live_bytes=928 peak_live_bytes=3472
HEAP DEBUG: leak summary: live_blocks=20 live_bytes=928
```

One leaked block per iteration (20/20); PHP prints the same `210`.

## Notes

- Indexed sibling (`$m = [$i + 1, 5];` read via `$m[0]`) in the identical loop shape: heap-clean.
- A single assignment (no loop rebind) is also clean — the leak is specifically the **previous**
  hash not being released when the local is rebound.
- Likely family: #528 (indexed rebind leak after pushing into an outer array), umbrella #551.

mirchaemanuel:
## Summary

Reassigning an associative-array local to a fresh literal inside a loop leaked the previous hash's boxed value on every iteration. Stdout was correct; the heap grew linearly with the iteration count. The indexed-array sibling was already clean.

```php
$count = 0;
for ($i = 0; $i < 20; $i++) {
    $m = ["k" => $i + 1, "j" => "x"];
    $count = $count + $m["k"];
}
echo $count, "\n"; // 210, but --heap-debug: live_blocks=20
```

## Root cause

The rebind release itself was correct — the previous hash was released and `__rt_hash_free_deep` decrefs its boxed values. The leak was one level deeper. `__rt_hash_set` takes ownership of the inserted value (it does not incref), so the hash-value materializer only re-retains values it classifies as *borrowed* through `value_can_own_mixed_box_source`. That allow-list of fresh owned boxed-Mixed producers omitted the checked-integer arithmetic ops (`ichecked_add`/`sub`/`mul`) and `mixed_numeric_binop`. A boxed arithmetic result such as `$i + 1` was therefore treated as borrowed and re-incref'd with no matching release — one leaked block per rebind. The indexed sibling stayed clean because `array_push` retains and the lowering releases the owning temp.

## Fix

Recognize those four ops as owned boxed-Mixed producers in `value_can_own_mixed_box_source`. They always return a freshly allocated, solely-owned boxed Mixed cell (never an operand alias), so a `mixed`-valued container (hash, indexed array, or object property) steals the sole reference on insert instead of re-retaining it. The box is then released exactly once when the container is torn down. The predicate has always been an over-approximation paired with the source type by its callers (it already includes `call`/`runtime_call`/`method_call`, which can produce non-refcounted results), so the addition follows the existing contract.

## Test plan

New heap-debug regression suite `tests/codegen/runtime_gc/assoc_rebind_release.rs` (12 tests, all assert `leak summary: clean` + correct stdout): the exact #595 repro; `ichecked_mul`/`sub` and `mixed_numeric_binop` value variants; empty-literal rebind; a dynamic-string value (guards the already-clean string path); a `mixed` object property (fixed by the same classification); the indexed control (regression guard); ownership guards where the hash escapes into an outer array, is passed to a function, or is returned (double-free guards); and the #528 loop-grown-rebind shape.

Red→green: 8 of the 12 leaked before the fix, all 12 pass after. Focused slices green: `runtime_gc arrays::nested_mixed_write hash assoc` (585) and `objects ownership reassign invoker` (315, covering the predicate's other call sites — object properties and invoker arguments). `cargo build` is warning-free.

## Note on #528

#528 (indexed rebind leak after pushing into an outer array) was re-verified on current `main` and is already heap-clean — fixed structurally by the EIR ownership work that has landed since it was filed. It is a different code path from this fix; this PR adds its shape as a permanent regression guard, and the issue can be closed as already resolved.

Fixes #595


mirchaemanuel:
While bisecting this I found an adjacent pre-existing refcount corruption, independent of this fix (fails identically with the classification change reverted): #604 — a directly-passed owned boxed-mixed call argument whose return value is also consumed trips `heap debug detected bad refcount` in a loop. Routing the value through a local first is clean.

nahime0:
Maintainer integration update on `650e3855e`:

- Merged current `main` (`0fd01c162`) into the existing contributor branch without rewriting its history.
- Replaced the duplicated non-string codegen producer allow-list with the EIR ownership contract: after final local-slot widening and provisional-release pruning, lowering publishes proven owning temporaries as `Ownership::Owned`, updating both `Value` and `Instruction` metadata; Mixed boxing/container stores now consume that metadata.
- Kept strings on the dedicated classifier because `Owned` alone cannot distinguish heap-owned strings from concat scratch storage.
- Added EIR regression coverage for checked arithmetic and `MixedBox` ownership, borrowed user-call results, metadata coherence, and the string-scratch boundary.
- Focused local validation on the published candidate: `cargo build`; 12 ownership-lowering tests; 231 `runtime_gc` tests; all 12 #595 tests with normal optimization and with `ELEPHC_IR_OPT=off`; assembly-comment and `git diff --check` hygiene.

The adjacent #604 call-argument corruption remains intentionally separate and is being handled in #618.

CI is now running on the exact published head `650e3855eada3a7e9b3a34d611388b815db3e3a3`.

nahime0:
CI repair on `06fa48c23`:

- Reproduced the deterministic `linux-x86_64` failure in `test_fiber_invokable_object_literal_uses_descriptor_receiver` locally (both CI retries produced a dangling receiver).
- Root cause: the new final ownership metadata correctly marks the inline `ObjectNew` as `Ownership::Owned`, but descriptor capture treated `Owned` alone as permission to steal the reference. Lowering also emits an explicit `Release` for the Fiber constructor argument after the constructor retains/copies it, so the captured receiver was freed and later overwritten by `Fiber::start()` arguments.
- Fixed the ownership-transfer boundary without restoring the producer allow-list: retaining consumers may steal only when the value is `Owned` **and** EIR has no explicit `Release` use for that value. Descriptor/closure captures and implicit Mixed/global boxing now share this predicate. An owned value with scheduled cleanup is retained; a genuinely transferred value is still stolen.
- Updated the changelog wording to document the distinction between ownership and consumability.

Focused validation:

- exact failing test: pass with `--ir-opt` on and off
- all 31 `codegen::fibers::arguments` tests: pass
- 4 inline instance-method first-class callable tests: pass
- captured-closure descriptor cleanup regression: pass
- all 231 `runtime_gc` tests, including the 12 issue #595 cases: pass
- `cargo build`, assembly-comment alignment, and `git diff --check`: pass

CI should restart on exact head `06fa48c235ca8a62cd527d7aa67330f63fac6ed7`.

PR #605, opened 2026-07-23
test(nulls): lock in ?? string-miss default selection

### LINKED ISSUE #554: Null coalescing misses the default for missing statically-string array elements
## Summary

Null coalescing does not select its default when a missing array access has a statically known `Str` result:

```php
<?php
$a = [
    ['', 'word0'],
    ['', 'word1'],
];

echo '[' . ($a[7][1] ?? 'dflt') . "]\n";
echo '[' . ($a[1][7] ?? 'dflt') . "]\n";
echo '[' . ($a[1][0] ?? 'dflt') . "]\n";
```

PHP distinguishes the two missing reads from the real empty string and prints:

```text
[dflt]
[dflt]
[]
```

On current elephc, the already non-crashing second-index miss `$a[1][7] ?? 'dflt'` yields `''` instead of `'dflt'`. PR #533 makes the first-index miss `$a[7][1]` safe instead of dereferencing the null-container sentinel, but it then exposes the same pre-existing semantic gap: that expression also yields `''`.

The coalescing expressions must remain silent, matching PHP.

## Root cause

`??` lowers its left-hand access silently and then tests the result with `IsNull`. In `src/codegen/lower_inst/predicates.rs::emit_is_null_result()`, statically known `PhpType::Str` values fall through to constant false.

A missed read of a statically-string slot is therefore indistinguishable at the null check from a legitimate empty string. By the time `IsNull` runs, the missing/null information has been erased.

Current `main` at the time of filing: `0b26415c9b`.

## Design scope

A fix must preserve a distinction between:

- a missing/null string-valued read, for which `??` selects the default; and
- a real empty string, for which `??` keeps `''`.

One possible direction is to give `Str` an explicit null representation. Another is to preserve missingness explicitly through silent access/coalesce lowering. Either approach needs a broad audit because strings use multiple storage classes and ownership contracts.

At minimum, audit:

- persistent and literal empty strings;
- concat/scratch-backed strings;
- heap-owned strings, including owned zero-length allocations;
- casts and builtin-produced strings;
- array/hash/property/string-offset miss paths;
- local, argument, return, property, and array-element storage;
- string ABI materialization and release/refcount helpers on every supported target.

The implementation must not reinterpret legitimate empty strings as null, create invalid sentinel pointers for string consumers, or introduce leaks/double releases.

## Regression coverage

- [ ] First-index miss: `$a[7][1] ?? 'dflt'` returns `'dflt'` silently after #533.
- [ ] Second-index miss: `$a[1][7] ?? 'dflt'` returns `'dflt'` silently.
- [ ] A present empty string remains empty and does not select the default.
- [ ] Cover indexed and string-keyed associative arrays with statically-string values.
- [ ] Cover nullable string locals/returns where the same representation is used.
- [ ] Cover `--ir-opt=on` and `--ir-opt=off`.
- [ ] Heap-debug remains clean for missing and present-empty controls.
- [ ] Validate macOS ARM64, Linux ARM64, and Linux x86_64 through the shared EIR/runtime path.
- [ ] Preserve the null-container crash guards and warning behavior fixed by #533.

## Related

- #526 — chained first-index miss currently dereferences the null-container sentinel.
- #533 — guards sentinel-consuming container paths; explicitly leaves this string-null/coalescing gap for follow-up.

mirchaemanuel:
Closes #554.

## Summary

`($a[i][j] ?? 'dflt')` must select the default when the accessed slot is statically typed `Str` and the read misses, while a *present* empty string keeps `''`. This PR adds the regression coverage the issue's checklist calls for. **The underlying behavior is already correct on `main`** — it was fixed as a side effect of #533 — so this is a test-only change that locks the behavior in against silent regression and lets #554 be closed.

## Root cause (resolved by #533)

`??` lowers its left-hand access silently (`ArrayGetSilent`) and then tests the result with `IsNull`. Before #533, `emit_is_null_result` let statically-known `PhpType::Str` values fall through to constant-false, so a missed string slot — whose fallback is `{ptr = NULL_SENTINEL, len = 0}` — was indistinguishable from a real empty string, and every miss yielded `''` instead of `'dflt'`.

#533 closed the gap by routing `PhpType::Str` through `emit_string_null_bool`, which compares the string pointer against `NULL_SENTINEL`. This is sound because legitimate empty strings never carry that pointer: they use a null (`0`) or valid data pointer regardless of storage class (literal, persistent, scratch/concat, heap zero-length). Only the miss fallback stamps the sentinel, and it is consumed by the coalesce branch — never materialized as a usable string.

## Design rationale vs the two sketched directions

The issue sketched (a) giving `Str` a global explicit null representation, or (b) preserving missingness through the silent-access/coalesce lowering. #533 realized the minimal form of (b): the miss information is carried locally in the string pointer (the sentinel already flowed there for the crash guards) and consumed at the `IsNull` decision, with no change to string storage classes or ownership contracts — avoiding the broad audit the issue feared.

## Test plan (maps to the issue's regression checklist)

- First-index miss `$a[7][1] ?? 'dflt'` → `'dflt'` silently — ✅
- Second-index miss `$a[1][7] ?? 'dflt'` → `'dflt'` silently — ✅
- Present empty string stays `''` — ✅
- Indexed **and** string-keyed associative arrays with statically-string values — ✅
- Nullable string locals/returns through `??` — ✅
- `--ir-opt=on` and `--ir-opt=off` — ✅ (both pass)
- Heap-debug clean for missing and present-empty controls — ✅
- `??` path stays silent; #533's null-container crash guards and warning behavior preserved — ✅

Tests fail if #533's `PhpType::Str` arm is reverted to constant-false (reproducing the `''`-instead-of-`'dflt'` symptom) and pass as shipped. Indices are bound to `$argc`-derived locals so the element read stays statically `Str` (an inline arithmetic index is typed `Mixed` and would route through the boxed-Mixed null path instead).

### DIFF
--- a/tests/codegen/control_flow/nulls.rs
+++ b/tests/codegen/control_flow/nulls.rs
@@ -134,3 +134,141 @@ echo $r[0], $r[1], $s[0], $s[1], "\n";
     );
     assert_eq!(out, "1212\n");
 }
+
+/// Regression test for issue #554: `??` must select its default when a missing
+/// indexed-array read has a statically-known `Str` result, while a present empty
+/// string keeps `''`. Indices are bound to `$argc`-derived locals so the element
+/// read stays statically `Str` (an inline arithmetic index like `$a[$argc + 6]`
+/// is typed `Mixed` and would route through the boxed-Mixed null path instead),
+/// forcing the miss through the `IsNull`(`Str`) branch decision: a missed string
+/// slot carries the null-string pointer sentinel, a real empty string does not.
+/// Covers both the first-index miss (missed outer container) and the second-index
+/// miss (missed inner string slot).
+#[test]
+fn test_null_coalesce_string_array_miss_selects_default() {
+    let out = compile_and_run(
+        r#"<?php
+$a = [['', 'word0'], ['', 'word1']];
+$miss_i = $argc + 6;
+$hit_i = $argc;
+$miss_j = $argc + 6;
+$hit_j = $argc - 1;
+echo '[' . ($a[$miss_i][$hit_j] ?? 'dflt') . "]\n";
+echo '[' . ($a[$hit_i][$miss_j] ?? 'dflt') . "]\n";
+echo '[' . ($a[$hit_i][$hit_j] ?? 'dflt') . "]\n";
+"#,
+    );
+    assert_eq!(out, "[dflt]\n[dflt]\n[]\n");
+}
+
+/// Regression test for issue #554 using the issue's verbatim literal-index
+/// reproduction: `$a[7][1] ?? 'dflt'` and `$a[1][7] ?? 'dflt'` select the
+/// default, and the present empty string `$a[1][0]` stays `''`, matching PHP's
+/// `[dflt]` / `[dflt]` / `[]` output exactly.
+#[test]
+fn test_null_coalesce_string_array_miss_literal_repro() {
+    let out = compile_and_run(
+        r#"<?php
+$a = [['', 'word0'], ['', 'word1']];
+echo '[' . ($a[7][1] ?? 'dflt') . "]\n";
+echo '[' . ($a[1][7] ?? 'dflt') . "]\n";
+echo '[' . ($a[1][0] ?? 'dflt') . "]\n";
+"#,
+    );
+    assert_eq!(out, "[dflt]\n[dflt]\n[]\n");
+}
+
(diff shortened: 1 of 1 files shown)


nahime0:
Given that this PR introduces only tests to cover what #533 fixed for #554, I'm gonna merge it directly. Thanks @mirchaemanuel

PR #613, opened 2026-07-23
fix(builtins): release boxed string elements consumed by implode over a mixed array

### LINKED ISSUE #601: implode() over a boxed-mixed indexed array leaks one block per element appended through the Mixed path
## Summary

`implode()` over an indexed array held in a boxed `mixed` cell leaks the boxed element cells:
one heap block per element that entered the array through the Mixed append path. Stdout is
correct; the heap grows with the element count.

Found while working on #592 (PR #600); independent of that fix — no sentinel involved, the
merge arms below are plain literals and the appends land on a real array.

## Repro

```php
<?php
$r = $argc == 1 ? [1] : ["a"];
$r[] = "x"; $r[] = "y"; $r[] = "z";
echo implode(",", $r), "\n";
```

Compiled with `--heap-debug` (main `1da317551` + PR #600, macos-aarch64), run with no args:

```
1,x,y,z
HEAP DEBUG: allocs=11 frees=8 live_blocks=3 live_bytes=144 peak_live_bytes=520
HEAP DEBUG: leak summary: live_blocks=3 live_bytes=144
```

Three appends → three leaked blocks. PHP prints the same `1,x,y,z`.

## Notes

- The same shape WITHOUT the `implode()` call (reading elements by index instead) is
  heap-clean, so the appends themselves are balanced — the leak is in implode's consumption of
  the boxed Mixed elements (each boxed cell read for stringification is retained or copied and
  never released).
- Likely family: #486 (owned temporaries passed as mixed arguments never released), umbrella
  #551.

mirchaemanuel:
## Summary

`implode()` over an indexed array held in a boxed `mixed` cell leaked one heap
block per string element (issue #601). Stdout was correct but the heap grew with
the string-element count.

## Root cause

In `__rt_implode`'s Mixed-element loop each boxed element is stringified through
`__rt_mixed_cast_string`. Its string-tag path calls `__rt_str_persist`, which
allocates a fresh owned heap block (kind 1) detached from the source mixed owner.
implode copied those bytes into the shared concat buffer but never released the
persisted copy — one leaked 48-byte block per string element. Integer, float, and
bool elements stringify into shared scratch storage (`__rt_itoa`/`__rt_ftoa`) with
no allocation, so they never leaked.

## Fix

The `__rt_implode` Mixed-element loop now records the cast string pointer in a
dedicated frame slot and releases it with `__rt_heap_free` once its bytes have been
copied, on both the ARM64 and x86_64 paths:

- Borrowed typed-array string slots clear the slot each iteration, so nothing is
  freed (they are owned by the array).
- Integer/float/bool casts return shared scratch storage; `__rt_heap_free`'s
  foreign-pointer guard (live-heap range check) makes the release a safe no-op for
  them.
- The release runs strictly after the copy, so there is no use-after-free; the
  joined output lives in the concat buffer and is never freed.

The x86_64 frame grows from 64 to 80 bytes for the extra spill slot (balanced in
prologue/epilogue); ARM64 preserves its loop registers around the release call.

## Tests

heap-debug regression tests in `tests/codegen/runtime_gc/regressions.rs`: the exact
repro, a multi-string mixed literal, an int-in-mixed no-op-release case, a
20-iteration loop proving the leak no longer scales, an empty-separator join, and a
typed-array control that must stay untouched. Plus emitter unit tests asserting the
release path and balanced frame on both targets. Focused slices (`runtime_gc` 225,
`implode`/`join` 14, `strings` 212, `mixed` 285, `arrays` 386) pass with no
regressions. Linux x86_64 end-to-end coverage relies on CI (no local
cross-assembler).

Fixes #601


mirchaemanuel:
Adjacent pre-existing bug found while validating this fix (reverified on the pre-fix build — independent of this change): #614, `implode()` over a mixed array corrupts the output when a float element follows a string element (`["x", 1.5]` → `1.1.1` instead of `x,1.5`); suspected `__rt_ftoa` clobbering the shared concat buffer mid-join.

mirchaemanuel:
Rebased onto current `main` (`0fd01c162`) so this merges without a merge commit, per your standing preference on contributor branches that go stale.

The only conflict was `CHANGELOG.md`, and it was purely additive: `main` had added the two CLI entries (`--timings`/`--quiet` spinner work and the `-h`/`--help` overhaul) while this branch added the `implode()` leak entry. All three are kept, in that order, at the end of `[Unreleased]`. No code conflicted — the `implode.rs` and regression-test changes replayed untouched.

Verified on the rebased head `fe1145e5c` (macOS aarch64): `cargo build` clean, and the `implode` filter is 13 passed / 0 failed, which covers the 6 new #601 regressions plus the pre-existing implode coverage including the #540 property-argument release test.

Nothing else changed. The branch is now a single commit on top of `main`.

mirchaemanuel:
Rebased again onto current `main` (`afee4cb47`) — it had gone stale behind today's merges. This time `tests/codegen/runtime_gc/regressions.rs` conflicted too: the #586 tests that landed via #593 sit right where the #601 ones do, and the two blocks end with identical assertion tails, so the merge needed the boundary rebuilt rather than just both sides kept. Verified on `668b4a491`: test target compiles clean, `implode` 13 passed / 0 failed, and the whole `codegen::runtime_gc` module 246 passed / 0 failed, which covers the merged-in #586 tests as well as ours. CHANGELOG conflict was additive as before.

PR #616, opened 2026-07-23
test(count): lock the null-container sentinel path after the #585 normalization

### LINKED ISSUE #602: count() on a null-container sentinel mixed cell segfaults instead of raising PHP's TypeError
## Summary

`count()` on a `mixed` cell whose payload is the null-container sentinel (from a missed array
read forwarded through a ternary merge) dereferences the sentinel and segfaults (exit 139).
PHP 8.4 throws `TypeError: count(): Argument #1 ($value) must be of type Countable|array,
null given`.

Found while working on #592 (PR #600); that fix covers the **append** path (which now
auto-vivifies before any later `count()` sees the cell), but a raw sentinel cell that reaches
`count()` directly still crashes.

## Repro

```php
<?php
$rows = [[1, 2]];
echo count($r), "\n";
```

Run with no arguments (missed-read arm):

- **elephc** (main `1da317551`, also with PR #600 applied): `Warning: Undefined array key 5`,
  then SIGSEGV (exit 139).
- **PHP 8.4**: the same warning, then an uncaught `TypeError` (count on null) — a clean
  diagnostic instead of a crash.

## Analysis

The count lowering for a boxed Mixed receiver (`__rt_mixed_count` or its inline call-site)
loads the container payload and dereferences it with at most a plain-zero guard, so the
non-zero in-band sentinel (`0x7ffffffffffffffe`) slips through — same hole family as #585
(get/set helpers, PR #591) and #592 (append path, PR #600), per the #533 convention the guard
belongs where the pointer is dereferenced.

Expected behavior: match PHP — raise the `TypeError` (or elephc's closest uncaught-exception
equivalent) rather than crashing.

## Related

- #533 — null-container sentinel guard convention.
- #585 / PR #591 — get/set helpers guarded.
- #592 / PR #600 — append path guarded + auto-vivify.
- #581 — var_dump on a sentinel-carrying local, same family.

mirchaemanuel:
## Summary

`count()` on a boxed `Mixed` receiver whose payload was the null-container sentinel used to
segfault (exit 139). A missed array read forwarded through a ternary merge
(`$argc == 1 ? $rows[5] : ["a", "b"]`) was the concrete repro from issue #602.

That crash is **already gone on `main`**: the null-container normalization from issue #585
(PR #591) turns the sentinel into a real null before `count()` ever sees it, and
`__rt_mixed_count` already guards tag-4/5 payloads with `emit_branch_if_null_container`.
There is no dedicated runtime change left for this PR to land — an earlier guard here became
unreachable after rebasing onto that normalization.

This PR is **tests only**. It locks the post-#585 behavior so the #602 crash cannot return
silently, and it records the remaining PHP-parity gap so #617 cannot be changed by accident.

## Observed behavior (locked here)

```php
<?php
$rows = [[1, 2]];
echo count($r), "\n";
```

Run with no arguments (missed-read arm):

- **Before #585:** `Warning: Undefined array key 5`, then SIGSEGV (exit 139).
- **On `main` / this PR:** the same warning, then `0` (exit 0) — no crash.
- **PHP 8.4:** the same warning, then
  `TypeError: count(): Argument #1 ($value) must be of type Countable|array, null given`.

The quiet `0` for null receivers also preserves the off-web BUG-0 convention
(`count($_SERVER) === 0` before assignment). Closing the PHP TypeError gap without breaking
that convention is tracked separately in #617. After normalization, sentinel and real null
are no longer distinguishable at `count()`, so #617 is an all-or-nothing decision.

## Tests

New regression file `tests/codegen/regressions/mixed_count_sentinel.rs`:

- uncaught sentinel `count()` does not crash; prints `0` and the undefined-key warning
- `catch (TypeError)` does **not** fire (explicit #617 divergence lock)
- the same path is heap-clean under heap debug
- a real null cell (`json_decode("null")`, tag 8) counts as `0` the same way
- the same ternary merge delivering a real array counts correctly
- an empty container in a `mixed` cell still counts as `0` without a warning

## Non-goals

- No runtime / codegen changes
- No CHANGELOG entry (nothing user-facing ships beyond regression coverage)
- No PHP TypeError parity for null or scalar Mixed receivers (see #617)

Fixes #602

### DIFF
--- a/tests/codegen/regressions.rs
+++ b/tests/codegen/regressions.rs
@@ -33,3 +33,5 @@ mod mixed_method_dispatch;
 mod switch_and_float_params;
 #[path = "regressions/return_this_ownership.rs"]
 mod return_this_ownership;
+#[path = "regressions/mixed_count_sentinel.rs"]
+mod mixed_count_sentinel;
--- a/tests/codegen/regressions/mixed_count_sentinel.rs
+++ b/tests/codegen/regressions/mixed_count_sentinel.rs
@@ -1,0 +1,141 @@
+//! Purpose:
+//! Regression tests for issue #602: `count()` on a boxed `Mixed` receiver whose payload was the
+//! null-container sentinel (produced by a missed array read forwarded through a ternary merge)
+//! used to dereference the sentinel and segfault (SIGSEGV 139). The null-container
+//! normalization introduced with issue #585 removed the crash by turning the sentinel into a
+//! real null before `count()` ever sees it, so these tests lock the resulting behavior in place
+//! rather than re-implementing a guard.
+//!
+//! Called from:
+//! - `cargo test` through Rust's test harness.
+//!
+//! Key details:
+//! - `$argc` keeps the ternary runtime-unknown so the merge (and the missed-read arm) survive
+//!   AST folding; the tests run with no arguments, so the missed-read arm executes.
+//! - A sentinel-carrying cell, a plain null pointer, and a boxed null cell (tag 8, e.g.
+//!   `json_decode("null")`) all count as `0`, matching the codebase's off-web
+//!   `count($_SERVER) == 0` convention.
+//! - That quiet `0` is deliberately *not* PHP parity: PHP raises
+//!   `count(): Argument #1 ($value) must be of type Countable|array, null given` for every null
+//!   receiver. Closing that gap without breaking the off-web convention is tracked in issue
+//!   #617; these tests document the divergence so a future parity change is a conscious one.
+
+use crate::support::*;
+
+/// Issue #602: the sentinel-carrying Mixed cell must not crash. Before the #585 normalization
+/// this program died with SIGSEGV 139; it now reports the missed read and counts as `0`.
+#[test]
+fn test_mixed_count_null_container_sentinel_does_not_crash() {
+    let out = compile_and_run_capture(
+        r#"<?php
+$rows = [[1, 2]];
+$r = $argc == 1 ? $rows[5] : ["a", "b"];
+echo count($r), "\n";
+"#,
+    );
+    assert!(
+        out.success,
+        "the sentinel receiver must not crash count(), stdout={:?} stderr={:?}",
+        out.stdout, out.stderr
+    );
+    assert_eq!(out.stdout, "0\n");
+    assert!(
+        out.stderr.contains("Warning: Undefined array key 5"),
+        "missing the missed-read warning, stderr={:?}",
+        out.stderr
+    );
+}
+
+/// Issue #602 / #617 divergence: no `TypeError` is raised, so a `catch (TypeError)` around the
+/// call never fires and execution simply continues. PHP would take the catch branch here. This
+/// test exists to make that difference explicit and to fail loudly if the behavior is changed
+/// without updating #617.
+#[test]
+fn test_mixed_count_null_container_sentinel_raises_no_type_error() {
+    let out = compile_and_run_capture(
+        r#"<?php
+$rows = [[1, 2]];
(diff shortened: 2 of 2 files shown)


mirchaemanuel:
The remaining count() parity gap mentioned in the description (non-null scalars in a mixed cell silently counting as 0, where PHP raises TypeError with the scalar type in the message) is now tracked as #617 — the machinery from this PR makes it a small follow-up, but it changes a test-locked behavior so it's kept separate.

mirchaemanuel:
CI caught a real interaction here: extending the TypeError to *plain* null receivers breaks the deliberate BUG-0 CLI convention that off-web superglobals count as 0 (`bug0_cli_read_of_server_superglobal_before_assignment_does_not_crash`). Correcting to raise the TypeError for the **sentinel forms only** (the actual #602 crash) and restoring the legacy 0 for plain null / boxed-null receivers — that residual PHP-parity gap moves to #617's scope. The docs-sync failure is just line-number drift from the builtins.rs edit; regenerating. Fix incoming on this branch.

mirchaemanuel:
Correction pushed (`f16c4b668` + docs sync `ac606f34c`). Final routing: **only** values carrying the in-band `NULL_SENTINEL` (as the cell or as a tag-4/5 payload) raise the `TypeError` — i.e. exactly the forms that segfaulted; a plain null cell (off-web superglobals), a null tag-4/5 payload, and a boxed real null (tag 8) all keep their legacy `0`, so the BUG-0 CLI convention (`count($_SERVER) === 0`) is preserved and `bug0_cli_read_of_server_superglobal_before_assignment_does_not_crash` passes again. PHP parity for real-null/scalar receivers stays tracked in #617. Docs desync was the `lower_empty` embedded line shifting 1077→1099 from the new guard; regenerated with the CI sequence, idempotent.

mirchaemanuel:
**This PR no longer needs to change any runtime code — your #585 normalization already fixed #602.** I've rewritten it as tests only and retitled it accordingly.

Rebasing it onto today's `main` conflicted in `mixed_count.rs` against `bd354a5ee`, and the conflict turned out to be semantic rather than textual: both changes decide how `count()` treats a null container. Resolving it textually produced a build where the sentinel branch was simply unreachable — the PR's own tests failed, returning `0` instead of the `TypeError`.

So I checked whether the original crash still exists. On `afee4cb47` with no patch of mine:

```php
$rows = [[1, 2]];
echo count($r), "\n";
```
```
Warning: Undefined array key 5
0                                  ← exit 0, was SIGSEGV 139
```

Normalization turns the sentinel into a real null before `count()` ever sees it, so there's nothing left to guard. The runtime change here had become dead code, and keeping it would have implied a fix that isn't mine to claim.

What remains is the PHP-parity gap: `php` raises `count(): Argument #1 ($value) must be of type Countable|array, null given` where we return `0`. That was always out of scope here and is tracked in #617 — and it's now the *only* thing left, since sentinel and real null are no longer distinguishable at that point. Which also means the narrow version this PR originally shipped (TypeError for sentinel forms only, legacy `0` everywhere else, preserving `count($_SERVER) == 0` off-web) can't be expressed any more. Worth knowing before #617 gets picked up: it's now an all-or-nothing decision against the off-web convention.

The six tests lock what `main` actually guarantees — no crash, warning plus `0`, heap-clean on that path, real-null cell counting the same way, populated merge and empty container unaffected — and one of them records that no `TypeError` is raised where PHP raises one, so #617's divergence can't be changed silently.

Verified on `6d88381e9` (macOS aarch64): test target compiles clean, `mixed_count_sentinel` 6 passed / 0 failed. Diff is two test files, nothing else — no `CHANGELOG` entry, since nothing user-facing changes.

PR #638, opened 2026-07-29
feat(core): finalize PHP streams implementation

Guikingone:
## Summary

PHP streams compliance audit and implementation. The branch landed in two waves: the
17 audited phases from the streams spec, then a resource-lifetime pass that replaced the
descriptor-as-identity model with a generation-safe registry and made resource identity,
closure and numbering match php-src everywhere — including across the `eval()` boundary.

## Wave 1 — audited phases

| Phase | Description |
|---|---|
| **P0.1** | Per-context-handle options table + `fopen(..., $context)` restoration |
| **P0.2** | `stream_socket_client`/`server` full 7-arg/6-arg signatures with `&$errno`/`&$errstr` by-ref validation |
| **P0.3** | `stream_select` → `poll(2)`, removing the 64-fd ceiling (cap 256) |
| **P0.4** | `stream_get_meta_data` real `wrapper_type`/`uri` via per-handle tables (file, php, data, http, ftp, compress.zlib/bzip2) |
| **P0.5** | HTTP wrapper method/headers/content/redirects — confirmed already implemented |
| **P0.6** | `file_get_contents`/`file_put_contents`/`readfile` full signatures (offset, length, flags, context) |
| **P0.7** | `fgetcsv`/`fputcsv` full PHP 8.4 signatures (enclosure, escape, eol) + 5-state CSV parser |
| **P0.8** | True `stream_filter_prepend` with 2-slot filter chain (shift + insert at slot 0) |
| **P0.9** | `PSFS_FEED_ME`/`PSFS_ERR_FATAL` propagation from user stream filters |
| **P1.9** | User-registered filters listed in `stream_get_filters()` |
| **P1.11** | `getaddrinfo` replaces deprecated `gethostbyname` in DNS resolution |
| **P1.12** | `%c` and `%x` format specifiers in `sscanf`/`fscanf` |
| **P1.16** | `stream_set_chunk_size` effective in `stream_get_contents` |
| **P2.1** | 8 historical aliases (`set_file_buffer`, `socket_set_block`, …) |
| **P2.5** | Residual constants (`SEEK_*`, `STREAM_FROM_*`, `STREAM_OPTION_CHUNK_SIZE`, `STREAM_META_MODIFIED`, `STREAM_PF_INET6`) |
| **P2.7** | `stream_is_local` based on scheme/wrapper_id instead of always `true` |
| **`$http_response_header`** | Auto-populated superglobal via `__rt_get_http_response_headers` |

Audit fixes from a 3-model jury: `fgetcsv` escape default back to `"\\"` (8.4, not 9.0),
`stream_socket_client` flags default `1` (`STREAM_CLIENT_CONNECT`), and `STREAM_CLIENT_*`
values corrected against php-src.

## Wave 2 — resource identity, lifetime, and eval parity

- **Generation-safe resource registry.** Streams, contexts and filters are registry
  slots with a generation, not raw descriptors, so a stale handle cannot resolve to a
  recycled descriptor.
- **PHP's resource numbering, exactly.** 1/2/3 are the standard streams, **4 is the
  request's default stream context** — created lazily at the first stream open of any
  kind, which is why `stream_context_get_default()` reports a lower id than a stream
  opened before it — and user resources start at 5. Ids are never reused.
- **Closed resources report `Unknown`** at every display site (`var_dump()`,
  `get_resource_type()`), for `fclose`/`pclose`/`closedir` alike, while keeping their id.
- **Filters are real resources on a chain**: built-in filter names resolve at run time,
  `php://filter` attaches through the same chain, and filters close with their stream.
- **Leaks closed**: request resources and the registry itself are released at exit —
  previously every compiled program leaked the slot table whether or not it used streams.
- **Network robustness**: `SIGPIPE` is ignored (a closed peer used to kill the process),
  the TLS attach path completes its handshake and honours `ssl.verify_peer`, and `fread`
  no longer truncates a stream at 64 KiB.
- **`eval()` parity**: the eight stream aliases carry eval signature metadata, socket
  by-ref params are declared in both required forms, a host-closed handle reports
  `Unknown` inside `eval()`, and eval-created streams draw from the same id counter as
  compiled code.
- **`fopen('php://stdout')` now duplicates the descriptor**, as php-src does. It used to
  hand back descriptor 1 itself, so `fclose()` on that handle closed the program's real
  standard output and every later `echo` was silently discarded while the process still
  exited 0.
- Also fixed on the way: `foreach` over an array of resources (refused as
  `stream argument PHP type Int`), `implode(',', [false, false])` (refused as
  `array_push for PHP type False`), the `data:` wrapper reported as `data` instead of
  PHP's `RFC2397`, and a transposed heap-magic word on the x86_64 stream TypeError path.

## Verification

Local runs are macOS ARM64; Linux coverage is CI's, and this PR is the first time CI has
ever run on this branch — it was conflicting until now, so no workflow had been triggered.
That first run found two Linux-only defects, both fixed here:

- **`bl _signal` on the whole AArch64 arm.** The leading underscore is a PLATFORM
  property, not an architecture one, so every Linux AArch64 program failed to link with
  `undefined reference to '_signal'`. `Emitter::bl_c` resolves it per platform.
- **`__rt_resource_registry_teardown` passed the slot array in `rdi`** while
  `__rt_heap_free` reads its operand from `rax`. Every CLI program on linux-x86_64
  segfaulted at exit, AFTER printing correct output, which took the whole platform's
  `Codegen`/`Eval Codegen` matrices down. Reproduced in the repo's own x86_64 container
  on `<?php echo 1;` (`run=139`), fixed, and re-verified there (`run=0`). One shard went
  from ~672 failures to 4.

Also fixed after CI exposed them: `stream_socket_client`/`server` accepted one argument
more than PHP (a `peername` parameter PHP does not have), seven stale arity expectations
in `error_tests`, and a `?int` inside an array literal that crashed
`__rt_array_push_refcounted` because a nullable union never reached the tagged-scalar arms.

Green locally: `--lib` 2422/2422 · `-p elephc-magician` 1134/1134 · `error_tests` 362/362 ·
`runtime_gc` 316/322 · `runtime_gc::heap` 16/16 · `ir_backend_smoke_test` + array suites
286/286 · `eval_resource_id_tests` 14/14 · `resource_id_and_hash_context_tests` 25/25 ·
`builtin_parity_tests` 32/32 · `codegen::io::streams` 362/363.

Two further fixes landed after the first CI round: the registry's initial slot array moved
from the heap to static storage — a program compiled with a small `--heap-size` used to die
at startup with `Fatal error: heap memory exhausted`, and every program allocated 528 bytes
it never needed (`allocs=0` now under `--heap-debug`) — and a `?int` inside an array literal
no longer appends its tagged immediate through the refcounted path.

### Known red, characterized

- **TLS session registry** (4 tests). These are the specification tests for migrating
  `_tls_sessions` off the raw-fd side table; the implementation is preserved on
  `wip/tls-session-in-streamstate` and still needs the `ftps://` call sites, SNI string
  ownership, and PHP's refusal to re-enable crypto.
- **User filters are not chain nodes** (2 tests). `FilterState` already carries a
  `php_user_filter` slot and `__rt_filter_create` already accepts one, but
  `lower_user_stream_filter_attach` still routes a genuinely user-registered filter to the
  legacy per-descriptor table, and the chain applier skips nodes whose built-in id is 0
  ("user filters are applied by the bucket path"). So a user filter has no registry slot:
  it stays `is_resource() === true` after its stream closes, and `stream_filter_remove`
  detaches it unconditionally instead of honouring `PSFS_ERR_FATAL` from the closing
  flush. Finishing that migration is the remaining half of the filter work.
- **`fopen(..., $context)` through an untyped parameter.** An untyped parameter is seeded
  `PhpType::Int` by the checker, so `function f($c) { return $c; }` reports `int` for a
  resource. Pre-existing and documented in `fopen_core.rs`; accepting `Int` in the emitters
  was tried before and reverted (it trades a clear diagnostic for a runtime throw), so the
  fix belongs in the checker.
- **Four Linux-only stream behaviours**: `stream_select` wrapper-stream cast readiness, a
  dynamic `ftp://` URL through `file_get_contents`, `fpassthru` capture under output
  buffering, and `PSFS_PASS_ON` control from a user filter. All four pass on macOS, so they
  are only observable through this PR's CI.

## Docs

`CHANGELOG.md` gains the user-facing entries for both waves. `docs/php/streams.md`
documents the id model, the duplicated `php://` descriptor, and the remaining reported-name
differences (`stream_type` for `php://memory`/`temp`/`output` and `data:`, the empty `uri`
for `tmpfile()`, literal-only `php://` resolution). `docs/php/eval.md` documents the shared
numbering space. The 962 generated builtin pages were regenerated from the registries.

## Merge

`origin/main` is merged in (15 commits: `parse_url` #667, by-ref foreach #648). Every
conflict was a generated builtin page — 152 of them, no source conflict — resolved by
regeneration rather than by picking a side.


Guikingone:
## Merged `origin/main` (70 commits) — branch is mergeable again

619 conflicts: 600 generated builtin docs, 19 code. Generated docs were resolved by taking main's tree and re-running `scripts/docs/extract_builtins.py --render --force`, never by picking a side — the regenerator then restored this branch's signatures (`file()` with `$context`, `file_put_contents()` with `$flags`/`$context`, `fputcsv()` with `$escape`/`$eol`) on top of main's.

### Where both sides had built the same thing, main wins

- **`file()` `$flags`**: main's `emit_file_line_flags_*` replaces this branch's implementation. The branch keeps only what it adds — `__rt_file_get_contents_maybe_url` and the `$context` parameter.
- **`file_get_contents()`**: main's `$offset`/`$length` window is kept. Main's `require_absent_stream_context` guard is **dropped**: the read path honours a context now, so refusing one would be a regression. Its test flipped from "rejects a non-null context" to "accepts one".

### Where the two changes are orthogonal, both land

Main hardened every unbounded read against running past the 64 KiB `_concat_buf` into the adjacent BSS globals (`__rt_concat_reserve` / `__rt_concat_grow` / `__rt_concat_publish`). This branch moved the same helpers off raw descriptors onto opaque stream handles and StreamState. `__rt_fread`, `__rt_fgets`, `__rt_stream_get_line` and both `__rt_stream_get_contents` variants now carry both.

The x86_64 `stream_get_contents` helpers were still on the old `_concat_buf` compaction on this branch, so **main's reservation logic was ported into them** rather than left behind — taking this branch's file wholesale would have silently reverted main's fix on one target.

### Three ordering/liveness details this required

- `finish_fopen_context_scope` reads the boxed result out of `x0`/`rax` and calls `__rt_resource_release`, which clobbers the string result pair. It now closes **after** the range trim and the boxing. Closing it early segfaulted **every** `file_get_contents()` — including the 1-argument form.
- `__rt_stream_fd` clobbers `x1`/`rsi`, so the requested read length is reloaded from the frame before the reservation instead of being read out of the argument register.
- `fgets` needed one more frame slot for the resolved descriptor than main's reservation left free, so both arches grow their frame and the descriptor moves to the tail slot.

`_eof_flags`, `_popen_files`, `_dir_handles`, `_glob_handles`, `_tls_sessions`, `_stream_chunk_size` and `_stream_connect_host` stay deleted — the resource registry owns that state. Everything this branch declares now goes through main's `comm_directive` helper, including the eight raw `.comm` emissions it had added, so the Windows PE backend gets them too.

### Verification

| | |
|---|---|
| `stream\|file\|filter\|wrapper\|context\|fread\|fgets` | **1291/1291** |
| `web_tests` | **52/52** |
| broad shard (1/8 of everything) | 1856/1859 — the one failure passes in isolation (port contention) |

### Also in this push: `PSFS_FEED_ME` needs three changes, not two

The known `PSFS_FEED_ME` gap is still open, and probing php 8.5.6 corrected its scope. Two behaviours the earlier note had left implicit:

- a read filter that triples `"ab"` answers three `fread($f, 2)` calls with `ab|ab|ab` → PHP **caps the filtered result at `$length`** and keeps the remainder in a per-stream buffer;
- a filter still holding bytes when the stream ends gets a `$closing` dispatch **whose output reaches the reader**.

elephc has neither. So landing only "FEED_ME returns nothing" + "the read path re-reads" makes the fixture print `<ABCDEF>` — the leak becomes a **loss**. The ignored target test and `docs/php/streams.md` now say so, and the third part (a filtered-read buffer on the StreamState plus a closing flush) is scoped rather than half-built.

nahime0:
Closing this draft.

Wave 1 of streams is already on `main`. This 1,636-file / 305-commit branch cannot be the vehicle for Wave 2: it is ~276 behind, still ABI v1, has no iOS target, and collides with `#734` / `#663` / `#801`.

Please do not rebase `4c3f18fc`. Remaining unique work (generation-safe resource registry, honour `$context`, `$http_response_header`, `php://stdout` dup) belongs in a new, small series from current `main`.

Guikingone:
I'm currently working on the final implementation and my local branch is rebased 😭

nahime0:
Sorry, closed this too fast. Draft is back; keep going.

PR #639, opened 2026-07-29
test(implode): cover numeric stringification inside the mixed join (#614) — runtime fix superseded by a5a6ac6e4

### LINKED ISSUE #614: implode() over a mixed array corrupts output when a float element follows a string element
## Summary

`implode()` over a `mixed` array containing a **float** element produces corrupted output when
the float is not in first position. No crash, no leak — silently wrong result.

Found while fixing #601 (PR #613); verified identical on the pre-fix build, so it is
independent of that change.

## Repro

```php
<?php
$r = ["x", 1.5];
echo implode(",", $r), "\n";
```

- **PHP 8.4**: `x,1.5`
- **elephc** (main `1da317551`, with or without PR #613): `1.1.1`

The reversed order `[1.5, "x"]` prints correctly (`1.5,x`), and integer elements are immune in
any position.

## Suspected root cause

In `__rt_implode`'s Mixed-element loop, float elements stringify through `__rt_ftoa`, which
appears to write into the same shared `_concat_buf` used to accumulate the joined result (and
advances `_concat_off`), clobbering the bytes already written for earlier elements. Integer
elements go through `__rt_itoa`, which uses an uncommitted scratch region, and are unaffected —
consistent with the position-dependence.

A fix likely means giving the float stringification in this loop a scratch destination outside
the accumulating concat region (as `__rt_itoa` effectively has), or copying the float digits
before appending.

## Related

- #601 / PR #613 — the string-element leak in the same loop (fixed; this corruption is
  pre-existing and orthogonal).

mirchaemanuel:
Fixes #614.

## Root cause

`__rt_implode` reads `_concat_off` once on entry, keeps its write cursor in a register, and only publishes the advanced offset when the join is finished. A `mixed` element routes through `__rt_mixed_cast_string`, which stringifies floats with `__rt_ftoa` and ints with `__rt_itoa` — and **both write into `_concat_buf` at the still-unpublished offset**, i.e. directly on top of the bytes the join has already accumulated.

The suspicion in the issue was close but placed the defect on the wrong side: `__rt_ftoa` is behaving exactly as documented (`_concat_buf` at the current `_concat_off`, then advance). The violation is in the caller — `implode` writes into that buffer while telling the world it hasn't. So the fix belongs in `implode`, and `ftoa`/`itoa` keep their contract and their other callers untouched.

## The int case: the same bug, not a separate one

The report says integers are immune. They are not — they only *look* immune. `__rt_itoa` writes right-to-left from `offset+20`, so it corrupts from byte 20 backwards, and a short join never reaches that far. Past the window it corrupts exactly like a float:

```php
$r = ["xxxxxxxxxxxxxxxxxxxxxxxxx", 1]; echo implode(",", $r);
```
```
php   : xxxxxxxxxxxxxxxxxxxxxxxxx,1
elephc: xxxxxxxxxxxxxxxxxxxx1xxxx,1
                            ^ byte 20
```

Floats corrupt from byte 0 because `snprintf` writes left-to-right from the offset. That asymmetry is the whole position-dependence described in the issue, and both directions are covered by the regression tests.

## Fix

Before each `__rt_mixed_cast_string` call, publish the current cursor to `_concat_off` so the numeric cast formats *above* the joined bytes. Derive the final offset from the cursor rather than from `_concat_off`, which the casts have now legitimately moved. Applied to both the ARM64 and x86_64 emitters.

The copies stay safe in both directions: `ftoa` writes at the cursor and the element copy is then a self-copy at the same address; `itoa` returns a pointer at `cursor+20-k`, so the copy runs strictly forward from a higher source to a lower destination.

## Tests

Seven regressions in `tests/codegen/regressions/concat_buffer_args.rs` — the natural home, since the pre-existing tests there cover the sibling `_concat_buf` clobber for call arguments. All seven fail on the pre-fix build:

- float after string (the report's repro), float between strings, consecutive float casts
- leading float — the one order that passed *by accident* before the fix, pinned so it stays correct
- int past the 21-byte `itoa` scratch window
- a long interleaved join mixing both cast kinds
- the repro under `--heap-debug`, whose guard bytes shift the surrounding heap layout

## Verification (macOS ARM64, `e452449ba`... this branch)

| | |
|---|---|
| the 7 new regressions | 7 passed, 0 failed |
| same, with `ELEPHC_IR_OPT=off` | 7 passed, 0 failed |
| `concat_buffer` family | 14 passed, 0 failed |
| `implode` | 14 passed, 0 failed |
| `strings::` | 212 passed, 0 failed |
| `runtime_gc` | 240 passed, 0 failed |
| `regressions::arrays` | 88 passed, 0 failed |
| `cargo build` / `cargo test --no-run` / `git diff --check` | clean, zero warnings |

**Not run locally: the Linux targets.** Docker is unavailable on this machine, so the x86_64 emitter change could not be executed here. I verified it statically instead, by reading the generated `linux-x86_64` runtime assembly — the published-offset sequence and the finalize path emit as intended, and `rax`/`rdx` survive as the string result registers. CI owns the real signal for both Linux targets.

## Two separate defects found while verifying — deliberately not fixed here

Both reproduce identically on pre-fix `main`, so neither belongs in this PR:

1. **`implode()` over a homogeneous float array prints nothing.** `[1.5, 2.5]` → empty. It never reaches the mixed loop: an all-float array keeps a raw layout, and the non-mixed path reads each 8-byte double through the 16-byte `{ptr,len}` string-slot layout. Homogeneous int arrays are fine because they have a dedicated `__rt_implode_int`.
2. **`implode()` over a mixed array leaks one 48-byte block.** Present with an `int` element too (`["x", 1]`), absent without the `implode` call — so it is not the corruption and not element-type specific.

I'll file both as separate issues and reference them here.


mirchaemanuel:
Filed the two separate defects found while verifying this fix, both confirmed pre-existing on unpatched `main`:

- #640 — `implode()` over a homogeneous float array prints nothing (raw doubles read through the string-slot layout; the mixed loop is never reached)
- #641 — `implode()` over a boxed-mixed array leaks one 48-byte block per call, with any element type

Neither is touched by this PR.

mirchaemanuel:
## Do not merge the fix in this PR — it is already in `main`

I have to correct this PR: my local clone was 28 commits behind `illegalstudio/main`, and I diagnosed and fixed #614 against that stale tree. On current `main` (`9f74f5300`) the defect is **already fixed**, by commit `a5a6ac6e4` (@Guikingone's OPcache audit work, 27 Jul), which landed the same change I wrote here — publish the live destination cursor to `_concat_off` before each nested cast, and stamp the absolute end offset instead of adding the length to whatever `_concat_off` holds.

All seven of this PR's cases pass on unpatched `9f74f5300`:

| case | php | elephc on current `main` |
|---|---|---|
| `["x", 1.5]` | `x,1.5` | `x,1.5` |
| `["x", 1.5, "y"]` | `x,1.5,y` | `x,1.5,y` |
| `["a", 1.5, 2.5]` | `a,1.5,2.5` | `a,1.5,2.5` |
| `[1.5, "x"]` | `1.5,x` | `1.5,x` |
| `["xxxxxxxxxxxxxxxxxxxxxxxxx", 1]` | `xxxxxxxxxxxxxxxxxxxxxxxxx,1` | matches |
| 5-element interleaved join | matches | matches |
| `strlen(implode(...))` | `5` | `5` |

So the runtime change here is redundant and should not land.

## What may still be worth keeping: the integer case

One of the seven is not covered anywhere in the tree today. The original report stated that integer elements were immune; **they are not** — they are only immune while the joined prefix is shorter than 20 bytes, because `__rt_itoa` writes right-to-left from `offset+20` and so corrupts from byte 20 backwards, while floats corrupt from byte 0 (`snprintf` writes forward from the offset). That asymmetry is the entire position-dependence the issue described.

```php
$r = ["xxxxxxxxxxxxxxxxxxxxxxxxx", 1]; echo implode(",", $r);
// pre-fix: xxxxxxxxxxxxxxxxxxxx1xxxx,1   <- the digit lands at byte 20
```

`a5a6ac6e4` added unit tests asserting the emitted assembly (`test_implode_{arm64,x86_64}_publishes_live_cursor_across_mixed_cast`) plus an end-to-end float case, but nothing exercises the integer path past `itoa`'s 21-byte scratch window — I could not find any test using that shape.

**Proposal:** I strip the runtime change and rebase this down to the regression tests only, the way #616 and #605 went. Happy to close it instead if you'd rather not carry them. Tell me which and I'll push the reduced branch — I won't force anything onto this branch meanwhile.

The two follow-ups I filed from this work (#640 homogeneous float array joins to nothing, #641 one 48-byte block leaked per mixed-array `implode`) were verified against the stale tree, so I'll re-check both on current `main` and update or close them accordingly.

mirchaemanuel:
## Reduced to tests only, rebased onto current `main`

Pushed `5ff1812d3`. The branch is now based on `9f74f5300` and the runtime change is gone — `a5a6ac6e4` already fixed #614 in `main`, so there is nothing left to fix here.

**What remains:** seven end-to-end regressions in `tests/codegen/regressions/concat_buffer_args.rs`, plus one CHANGELOG line. They pin the observable PHP output of the CURSOR invariant that `implode.rs`'s preamble now documents — complementing the emitter unit tests, which pin the emitted instruction sequence rather than the behaviour.

The one I'd argue is worth carrying is the integer case. #614 reported integers as immune; they are not, they are only immune while the joined prefix is under 20 bytes, because `__rt_itoa` writes right-to-left from `offset+20`:

```php
$r = ["xxxxxxxxxxxxxxxxxxxxxxxxx", 1]; echo implode(",", $r);
// pre-fix: xxxxxxxxxxxxxxxxxxxx1xxxx,1
```

I could not find any existing test using that shape, so the integer direction of the invariant looks uncovered today. The other six cover float-after-string, float-between-strings, consecutive casts, a leading float (the order that passed by accident pre-fix), a long interleaved join, and the repro under `--heap-debug`.

**Verification on `5ff1812d3`, macOS ARM64:** the 7 new tests pass, `cargo build` is clean with zero warnings, `git diff --check` is clean. Linux targets not run locally (no Docker on this machine) — CI owns that signal.

Also re-checked the two follow-ups I filed from the stale tree: #641 does not reproduce on current `main` and I have closed it; #640 still stands but its symptom changed for the better — no longer a silent empty join, now an explicit `unsupported EIR backend feature: implode array element PHP type Float`, so I have updated it to read as a missing feature rather than wrong data.

nahime0:
@Guikingone This PR is tests only. I'm gonna merge it directly.

PR #644, opened 2026-07-29
fix(ir,codegen): release conditionally-aliased call arguments at runtime (#619)

### LINKED ISSUE #619: Alias-suppressed call arguments leak on non-aliasing return paths (needs runtime alias disambiguation)
## Summary

When an owned argument is passed directly to a callee whose return-alias summary says it
*may* return that parameter, the caller suppresses the argument's release. On calls where the
runtime actually takes a **non-aliasing** return path, the suppressed owned box/container
leaks — one block per call. This is the deliberate leak-over-crash trade-off of the alias
suppression (`ReturnArgAlias::Parameters` is a MAY summary, a union over branches), and it
predates PR #618 for array/hash arguments; #618 extends the same behavior to fresh
arithmetic `mixed` boxes when fixing the #604 refcount corruption.

Filing this so the remaining leak is tracked: the real fix is **runtime alias
disambiguation** — compare the returned pointer against the argument pointer after the call
and release the argument when they differ.

## Repro (arithmetic box, with PR #618 applied)

```php
<?php
function maybe($x, $c) { if ($c == 1) { return $x; } return 7; }
$s = 0;
for ($i = 0; $i < 20; $i++) { $r = maybe($i + 1, 0); $s = $s + $r; }
echo $s, "\n";
```

`--heap-debug`: stdout `140` (correct), `leak summary: live_blocks=20 live_bytes=800` — one
leaked box per call on the constant-return path. The aliasing path (`maybe($i + 1, 1)`) is
clean (`210`, no leak).

## Repro (pre-existing sibling: fresh array argument, plain main)

```php
<?php
function maybe($x, $c) { if ($c == 1) { return $x; } return 7; }
$s = 0;
for ($i = 0; $i < 20; $i++) { $r = maybe([$i], 0); $s = $s + 1; }
echo $s, "\n";
```

`--heap-debug`: stdout `20`, `live_blocks=20 live_bytes=1440` — same mechanism through the
long-standing `may_alias` suppression for containers.

## Suggested direction

After a call whose argument release was suppressed by the alias summary, emit a pointer
comparison between the returned value's storage and the suppressed argument's storage: if
they differ, release the argument. This recovers the non-alias paths without reintroducing
the double-release on the alias paths, and would fix both shapes above (and the container
sibling that has leaked since the original may_alias suppression).

## Related

- #604 / PR #618 — the refcount corruption on the proven-alias path (fixed; introduced the
  arithmetic-box variant of this leak as a documented trade-off).
- #486 — the argument-release family this machinery belongs to; umbrella #551.

mirchaemanuel:
Fixes #619 for boxed `mixed` arguments. Follow-up #665 tracks the bare-container path deliberately left out of scope.

Integrated current `main` `e015218f5` in maintainer merge `c1a3fd23e`.

## Root cause

`ReturnArgAlias::Parameters` is a MAY summary — a union over branches — so `proven_aliases_parameter` also holds for a callee that returns the parameter only conditionally (`if ($c) return $x; return 7;`). The suppression site (`src/ir_lower/expr/mod.rs`) therefore skipped the argument release on *every* path. That is required on the branch that hands the box back, or it is freed twice (#604), and it leaks one block per call on the branches that do not.

The lowering has to decide once, statically, for a fact that is only known per call. So the decision moves to runtime.

## Fix

A new EIR instruction, `ReleaseUnlessAliases(arg, result)`: after the call, compare the argument payload against the value the callee returned and release the argument only when they differ. Same pointer means ownership moved into the result and the caller must keep its hands off; different pointer means the callee dropped it and the caller still owns it.

Both supported ABIs lower it, reusing the comparison shape `emit_branch_if_cleanup_temp_aliases_result` already uses for ABI-side cleanup slots. On a callee that genuinely always returns the parameter the comparison always matches, so #604's behaviour is unchanged.

## Deliberate restriction, and why

The comparison is emitted **only when both sides are boxed `mixed`**, i.e. directly comparable as single pointers.

I initially emitted it unconditionally. That regressed four tests with `heap debug detected bad refcount`: a `mixed` result that *wraps* a bare container holds a different pointer than the container itself, so the comparison reads "not aliased" for a value the result does own, and releases it twice. Bare container arguments therefore keep the previous suppression and their pre-existing leak — the one that predates #618 — and `test_conditional_return_callee_container_arg_still_leaks_on_non_alias_path` pins that leak so the restriction stays deliberate and the test turns red the day the container path is covered. Covering it means reaching through the box to its payload field, which I'd rather do as its own change than smuggle in here.

## Tests

In `tests/codegen/runtime_gc/regressions.rs`, next to the #604 family:

- **non-alias path releases the argument** — the issue's arithmetic-box repro; leaked 20 blocks / 800 bytes before
- **mixed paths in one program** (`maybe($i + 1, $i % 2)`) — releasing on the aliasing iterations would double-free, skipping it on the others leaks; leaked exactly 10 blocks before, one per non-aliasing iteration
- **container argument still leaks** — the documented restriction above
- the existing `test_conditional_return_callee_alias_path_stays_balanced` is the #604 guard and still passes

## Verification on `35dc9e18b`, macOS ARM64

| | |
|---|---|
| the 4 conditional-return tests | 4 passed, 0 failed |
| same, `ELEPHC_IR_OPT=off` | 4 passed, 0 failed |
| `runtime_gc` | 256 passed, 0 failed |
| `callables` | 534 passed, 0 failed |
| `error_tests` | 1139 passed, 0 failed |
| `ir_backend_smoke_test` | 255 passed, 0 failed |
| `cargo test -p elephc --lib` | 919 passed, 0 failed |

Not run locally: the Linux targets — no Docker on this machine. The x86_64 lowering is the mirror of the AArch64 one and reuses the existing comparison helper's shape, but CI owns the real signal.

## Maintainer integration (2026-08-03)

- Opened #665 to track representation-aware runtime alias disambiguation for bare container arguments.
- Merged current `main` `e015218f51a5f321a8f8b7ddbced38f177ae1f53` without rewriting contributor commits.
- Resolved the changelog conflict by preserving every `main` entry and linking the residual container limitation to #665.
- Revalidated `cargo build`, the four `conditional_return_callee` regressions with EIR optimization enabled and disabled, assembly-comment alignment, and `git diff --check`.
- CI on merge head `c1a3fd23eca9b97ed23538738b8b770de4569f59` is the authoritative supported-target signal.


nahime0:
@greptileai please review

mirchaemanuel:
**CI note — the two failures are infrastructure, not test failures.**

`Build & Test` died in the checkout step:

```
fatal: unable to access 'https://github.com/illegalstudio/elephc/':
  Failed to connect to github.com port 443 after 75003 ms: Couldn't connect to server
The process '/opt/homebrew/bin/git' failed with exit code 128
```

and the `Codegen Tests (macos-aarch64 9/16)` shard failed unpacking its test archive:

```
failed to unpack `target/debug/deps/error_tests-3b0afadc715685d4` into .../nextest-archive-7W9Vyf/...
```

Neither reached an assertion — no test reported a failure. The other 111 checks are green. A re-run should clear both; the branch is also `CONFLICTING` against current `main` since it is based on `9f74f5300`, so a rebase would re-trigger CI anyway. Leaving that to you.

mirchaemanuel:
Good catch from the bot — both findings were valid and are fixed in `28d6daf9b`.

Inserting `lower_release_unless_aliases` directly above `lower_release` moved that function's docblock onto the new one. The result was exactly as described: `lower_release` left undocumented, and the new function carrying a stale opening line that described the old one. Restored the original line to `lower_release` and dropped it from the new doc.

Verified after the change: `cargo build` clean with zero warnings, `conditional_return_callee` 4/4, `git diff --check` clean.

The CI failures on this PR remain unrelated to the code — the checkout step timed out (`Failed to connect to github.com port 443`) and the codegen shard failed unpacking its nextest archive; no test reported a failure.

nahime0 (dismissed):
Maintainer follow-up: I handled the requested PR maintenance directly.

- I opened #665 to track the bare-container leak that remains outside this boxed-mixed fix.
- I merged current main into the contributor branch without rewriting contributor history; the integration commit is c1a3fd23eca9b97ed23538738b8b770de4569f59.
- I resolved the CHANGELOG conflict while preserving the entries from main and linking the follow-up issue.
- I revalidated the build, the four conditional-return regressions with EIR optimization both on and off, assembly-comment alignment, and git diff --check.
- Exact-head CI restarted on c1a3fd23eca9b97ed23538738b8b770de4569f59 in run 30826743063.

There were no requested code changes within the boxed-mixed scope. Tracking and current-main integration have now been completed by me as maintainer, and CI is running on the exact updated head, so this request-changes review can be dismissed.

nahime0:
@Guikingone if you give a run of review on this PR I'll proceed to merge it.

src/ir_lower/expr/mod.rs
```diff
@@ -13322,6 +13322,33 @@ fn release_owned_call_arg_temporaries_with_signature(
...
+                    if comparable {
+                        ctx.emit_void(
+                            Op::ReleaseUnlessAliases,
+                            vec![lowered.value, result],
+                            None,
+                            Op::ReleaseUnlessAliases.default_effects(),
+                            Some(span),
+                        );
+                    }
+                }
```
Guikingone, line 13351: Is this code executed or not? Looks like the condition is never met 🤔

nahime0: I've checked, the path is traversed. The if let is redundant, but it's just cosmetic.

Guikingone (approved):

PR #646, opened 2026-07-29
fix(codegen): guard var_dump's container walk against the null-container sentinel (#581)

### LINKED ISSUE #581: var_dump() on an Array-typed local carrying the null-container sentinel segfaults instead of printing NULL
## Summary

`var_dump()` on an Array-typed local that carries the in-band null-container sentinel (the value a missed array read materializes) starts printing `array(` and then segfaults: the dump walker dereferences the sentinel as a real array header instead of printing `NULL`.

Reproduced on `main` at `d36bc6b64`. Same family as the #526 read-side hardening — the dump/debug walkers were not part of that audit.

## Reproduction

```php
<?php
$a = [['x', 'y']];
$arr = $a[7];
var_dump($arr);
echo "done\n";
```

## Current behavior

```text
array(
```
then `SIGSEGV` (exit 139). `done` is never reached.

## Expected PHP behavior (verified with PHP 8.4)

```text
Warning: Undefined array key 7 in ...
NULL
done
```

## Notes

- The static type of `$arr` is `Array`, so the dump path takes the array branch unconditionally; the sentinel check needs to happen before the header walk (or the walker should route sentinel-carrying values to the `NULL` branch, matching how guarded read paths treat them since #526).
- The equivalent `echo $arr ?? 'was-null';` correctly prints `was-null`, so the value itself is a well-formed sentinel — only the dump walker mishandles it.

mirchaemanuel:
Fixes #581. Based on current `main` `9f74f5300`.

## Root cause

Not in the `__rt_var_dump_*` runtime walkers — in the lowering. `emit_var_dump_array` (`src/codegen/lower_inst/builtins/debug.rs`) guarded its header walk with a **zero-pointer check only**:

```rust
abi::emit_branch_if_int_result_zero(ctx.emitter, &null_label);
```

That comment described the one case it was written for: an untyped null-defaulted property rebound to array storage. A missed element read produces something else — the in-band null-container sentinel, which is **non-zero**. So the branch never fired, `array(` was written, and the next instruction dereferenced the sentinel as a live array header.

Visible in the emitted assembly for the repro:

```
    ldur x0, [x29, #-96]
    cbz  x0, _eir_main_var_dump_array_null_4   <- zero check only
    str  x0, [sp, #-16]!
    ...                                        <- writes "array("
    ldr  x0, [x0]                              <- SIGSEGV on the sentinel
```

which matches the reported output byte for byte: `array(` with no newline, then the crash.

## Fix

Swap the zero-only branch for the shared `sentinels::emit_branch_if_null_container`, which recognizes **both** null shapes and is already used by the guards from #526, #556 and #585. Scratch register from `abi::secondary_scratch_reg`, the same way `lower_array_len` takes one.

One call site covers the `Array`, `AssocArray` and `Iterable` arms plus both entries from `emit_var_dump_mixed`, because they share this function. The helper is target-aware, so both supported architectures are handled with no ARM64-only path.

Before / after on the report's repro:

```
before: Warning: Undefined array key 7 / array(     [exit 139, "done" never reached]
after:  Warning: Undefined array key 7 / NULL / done [exit 0]
php:    Warning: Undefined array key 7 / NULL / done [exit 0]
```

## Tests

Six regressions in `tests/codegen/regressions/arrays.rs`. Four are behavioural and fail on the pre-fix build with `program crashed`; one is a guard for the shapes that already worked; one asserts the emitted guard instruction directly.

The assembly test was checked with an explicit negative control — reverting the source makes it fail on both architectures with `missing sentinel comparison 'cmp rax, r10'` and `'cmp x0, x10'` respectively, so x86_64 is genuinely verified rather than assumed. It runs through `compile_source_to_asm_with_options` with `ELEPHC_TEST_TARGET`, the same way #556's test does, because the CLI builds a runtime object even under `--emit-asm` and so cannot cross-emit locally.

Two sibling shapes were fixed by the same one-line change and are covered: a **hash** source whose value is an array, and one whose value is a hash (the `AssocArray` arm) — the latter crashed too and is not mentioned in the issue. Also covered: a missed read through `?? null`, and a genuine `null` local as a non-regression guard.

## Verification (macOS ARM64, `beea5b280`)

| filter | result |
|---|---|
| the 6 new tests | 6 passed, 0 failed |
| same, `ELEPHC_IR_OPT=off` | 6 passed, 0 failed |
| asm test, `ELEPHC_TEST_TARGET=linux-x86_64` | 1 passed, 0 failed |
| asm test, `ELEPHC_TEST_TARGET=linux-aarch64` | 1 passed, 0 failed |
| `var_dump` (existing) | 34 passed, 0 failed |
| `sentinel` | 41 passed, 0 failed, 3 ignored |
| `array_miss` (#526 guards) | 20 passed, 0 failed |
| `null_container` (#556/#592) | 2 passed, 0 failed |
| `autovivify` (#600/#592) | 26 passed, 0 failed |

Narrow filters only, per the project's test policy; CI owns the full matrix. Linux targets not executed locally (no Docker), but the x86_64 lowering is verified through the asm test above rather than by assumption.

## Out of scope, reported separately

- **`print_r()` has the same crash and no guard at all** — not even the zero check `var_dump` had. `print_r($a[7])` still segfaults on current `main` and on this branch. Different lowering and different PHP semantics (`print_r(null)` prints nothing, it does not print `NULL`), so it wants its own change. Filed separately.
- `var_dump($a[7][2])` where the element type is a **scalar** prints `string(0) ""` instead of `NULL`. Both warnings are correct; the static type of the expression is `Str`, so the dump takes the string arm and the chained miss materializes an empty string rather than the sentinel. That is a value-materialization defect on the read path, not a walker defect, and predates this change.


nahime0 (approved):
Fine for me. I'll merge as is, without merging main here. I don't see potential semantic conflicts

Guikingone (approved):

Issue #643, opened 2026-07-29
By-ref foreach inside a by-ref array parameter fails to compile (EIR PhpTypeMismatch)


mirchaemanuel:
## Summary

A by-reference `foreach` inside a function whose parameter is itself by-reference fails to compile with an internal EIR validation error.

## Repro

```php
<?php
function f(array &$r) {
    foreach ($r as &$v) { $v = $v * 2; }
}
$a = [[1, 2]];
f($a[0]);
echo implode(",", $a[0]);
```

- **PHP 8.4.23**: `2,4`
- **elephc** (`main` `ba836179c`, macOS ARM64):
  ```
  EIR lowering error: EIR validation failed: PhpTypeMismatch(ValueId(14))
  ```

No source location is reported, and the program cannot be compiled at all.

## Notes

The nesting is what triggers it: a by-ref `foreach` over a plain local compiles and runs correctly, and a by-ref `array` parameter on its own is fine — combining them is what fails.

This is the shape PHP code normally uses to mutate a nested array through a helper, so it is likely to be hit in practice.

Found while diagnosing #580; verified identical on unpatched `main`.

Issue #642, opened 2026-07-29
By-ref foreach over an object property corrupts the property (count() returns a heap address)


mirchaemanuel:
## Summary

A by-reference `foreach` over an **object property** does not merely lose the writes — it leaves the property corrupted. Afterwards `count()` on it returns a heap address.

## Repro

```php
<?php
class C { public array $x = [1, 2]; }
$o = new C();
echo "before: ", implode(",", $o->x), " count=", count($o->x), "\n";
foreach ($o->x as &$v) { $v = $v * 2; }
unset($v);
echo "after:  ", implode(",", $o->x), " count=", count($o->x), "\n";
```

- **PHP 8.4.23**:
  ```
  before: 1,2 count=2
  after:  2,4 count=2
  ```
- **elephc** (`main` `ba836179c`, macOS ARM64, with or without `--ir-opt`):
  ```
  before: 1,2 count=2
  after:   count=4303311136
  ```

`implode()` returns nothing and `count()` returns what looks like a pointer, so the property's container header is no longer valid.

## Relationship to #580

#580 covers array-*element* sources (`$a[0]`, `$h['a']`), where the writes are silently lost but the parent stays intact and `count()` keeps returning 2. The property source is a different receiver kind and a strictly worse outcome — a corrupted container rather than a lost update — so it is filed separately rather than folded in.

The by-ref foreach copy-on-writes the container it iterates (measured in #580: the by-ref form allocates one extra block even with an empty body) and never publishes it back. For the property receiver something further goes wrong on top of that, since the original storage does not survive.

## Notes

Found while diagnosing #580; verified identical on unpatched `main`, so it is independent of any fix for that issue.

PR #648, opened 2026-07-29
fix(ir,codegen): fetch by-ref foreach element sources for writing (#580)

### LINKED ISSUE #580: By-reference foreach writes through an array-element source are silently lost
## Summary

By-reference `foreach` writes through a source that is an **array element** are silently lost. The loop runs, `$v` is assigned, but the parent array never observes the mutation. The same loop over a plain variable works.

Reproduced on `main` at `d36bc6b64` (also present with #579 applied — unrelated to the missing-element path it fixes).

## Reproduction

```php
<?php
$a = [[1, 2]];
foreach ($a[0] as &$v) { $v = $v * 2; }
unset($v);
echo implode(',', $a[0]), "\n";

$h = ['a' => [1, 2]];
foreach ($h['a'] as &$w) { $w = $w * 10; }
unset($w);
echo implode(',', $h['a']), "\n";

$plain = [1, 2];
foreach ($plain as &$p) { $p = $p * 3; }
unset($p);
echo implode(',', $plain), "\n";
```

## Current behavior

```text
1,2
1,2
3,6
```

Both element-source forms (indexed `$a[0]` and hash `$h['a']`) drop the writes; only the plain-variable form mutates.

No crash, no diagnostic — a pure silent-wrong-result divergence, which makes it nasty: any code that mutates a nested array in place via by-ref `foreach` (a very common PHP pattern) compiles and runs but computes the wrong data.

## Expected PHP behavior (verified with PHP 8.4)

```text
2,4
10,20
3,6
```

## Notes

- Likely related to the parent write-back gap family (#555/#345): the element read materializes a temporary for iteration and the mutated container is never published back into the parent slot, whereas the plain-local form iterates the local's own storage.
- Heap-debug stays clean on the repro, so this is a lost publish, not a lifetime issue.

mirchaemanuel:
Fixes #580 for indexed receivers. Based on current `main` `9f74f5300`.

## Root cause

Measured rather than guessed. Allocation counts under `--heap-debug` isolate which step copies:

| program | allocs |
|---|---|
| `$a = [[1,2]];` (baseline) | 2 |
| `$b = $a[0];` | 2 — no copy |
| `foreach ($a[0] as $v) {}` by value | 2 — no copy |
| `foreach ($a[0] as &$v) {}` by ref, **empty body** | **3 — one copy** |

`Op::ArrayGet` does not copy: it returns the parent's own container **with a retain**. That retain leaves the element at refcount 2 (parent + temporary), so the by-ref iteration copy-on-writes it, mutates the private copy, and releases it at the end of the loop. Every write is lost — no crash, no diagnostic.

The working plain-local form differs by exactly one instruction, and the loop bodies are byte-identical:

```
# works:   $plain
v4: array<int> own=maybe_owned = load_local slot[0]   # no retain -> refcount 1 -> no COW
# broken:  $a[0]
v7: array<int> own=owned      = array_get v5 v6       # retain   -> refcount 2 -> COW
```

The report guessed a missing write-back. It is upstream of that: the loop never shares the parent's storage in the first place.

## Fix

`Op::ArrayGetForWrite` performs the copy-on-write split up front and republishes the unique container into the parent's element slot, so the loop iterates storage the parent genuinely shares.

Its effects are deliberately broad — `READS_HEAP | WRITES_HEAP | WRITES_LOCAL | ALLOC_HEAP | REFCOUNT_OP | MAY_WARN`. Despite the name it is not a pure read: the split rewrites the receiver's element slot and the receiver's own local slot, so it must never be reordered or folded against the plain reads around it. `MAY_WARN` matters too — a missing index must still emit `Undefined array key` and yield the null-container sentinel the loop knows how to skip, which is what keeps **#556** intact.

By-value `foreach` keeps the ordinary retaining read: that extra reference is exactly what makes the runtime copy, and copying is precisely its semantics.

PHP's mid-loop visibility is satisfied, which is the discriminator a write-back-only fix would fail:

```php
foreach ($a[0] as &$v) { $v = $v * 2; echo $a[0][0], ';'; }
// php: 2;2;   this branch: 2;2;   (before: 1;1;)
```

## Scope

Indexed receivers, including elements that are themselves hashes (`$a[0]` holding `['x' => 1]`), and nested receivers (`$a[0][0]`).

**A hash receiver is not covered** (`$h['a']`). Reference-binding a hash element is unimplemented in elephc — an explicit `$r = &$h['k'];` is rejected by the checker with *"Reference assignment to an array element requires an indexed array"* — so there is no alias to iterate. Those keep their previous behaviour rather than starting to fail differently.

## Tests

13 regressions in `tests/codegen/types/iterable/foreach.rs`, covering the indexed element source, mid-loop parent visibility, a nested element source, the keyed form, a hash-valued element of an indexed receiver, a source shared with another owner, an aliased parent, a missing index (warns and skips), plus guards that by-value must NOT mutate and that the plain-local form still does, and a heap-debug leak guard.

## Verification (macOS ARM64, `f72d0e3c7`)

| filter | result |
|---|---|
| the 13 new tests | 13 passed, 0 failed |
| same, `ELEPHC_IR_OPT=off` | 13 passed, 0 failed |
| `iterable::foreach` | 41 passed, 0 failed |
| **`byref_foreach` (the #556 guards)** | **9 passed, 0 failed** |

Narrow filters only, per the project's test policy; CI owns the full matrix. Linux targets not run locally (no Docker on this machine).

The `byref_foreach` row is the one worth checking first: an earlier attempt of mine, aliasing the element through a hidden ref-cell, made the indexed cases pass while regressing exactly those three #556 tests — `LoadArrayElemRefCell` yields a null pointer for a missing index and skips the warning, so `foreach ($a[7] as &$v)` crashed instead of warning and skipping. That approach was discarded for this one.

## Related, not fixed here

Two strictly worse defects found while diagnosing this, both pre-existing and filed separately: #642 (by-ref `foreach` over an object property **corrupts** the property — `count()` returns a heap address) and #643 (by-ref `foreach` inside a by-ref `array` parameter fails to compile with `EIR PhpTypeMismatch`).


src/ir_lower/expr/mod.rs
```diff
@@ -8017,6 +8017,129 @@ fn lower_array_access(
...
+    // The separated element belongs to the parent's slot, not to this read — pin that explicitly
+    // instead of leaving it at the `MaybeOwned` default, which `mark_owned_temporaries` is free
+    // to promote to `Owned` later, and which would then have the loop release the parent's
+    // element on the way out.
+    ctx.builder
+        .set_value_ownership(result.value, Ownership::Borrowed);
+    release_coerced_source_if_owned(ctx, index_value, Some(index.span));
+    // An intermediate that fell back to a retaining read owns a reference to the container it
+    // handed us, and holding it for the whole loop would be a leak. Dropping it here is safe
+    // precisely because `element_fetch_for_write_applies` demanded a local-rooted chain: the base
```
nahime0: **Blocking:** the base local does not necessarily keep this borrowed element alive until the loop ends. The loop body may replace or unset the parent after `IterStart`, while the iterator stores only the borrowed element pointer.

Reproducer:

```php
$a = [[1, 2]];
foreach ($a[0] as &$v) {
    echo $v, ',';
    if ($v === 1) $a = [];
    $v *= 2;
}
unset($v);
echo 'done';
```

PHP and base `9f74f5300` produce `1,2,done`; this head produces `1,done` with `ELEPHC_IR_OPT` both on and off. The parent release leaves the iterator without an owner for its source, so this is also a potential use-after-free rather than only an iteration-count mismatch.

Please carry an explicit lifetime owner acquired after the COW separation (so the pin does not trigger another split), release it on normal and early loop exits, and add this case as a heap-debug regression test.

tests/codegen/types/iterable/foreach.rs
```diff
@@ -625,3 +625,309 @@ fn test_iterable_variadic_arg_stays_boxed_in_runtime_array() {
...
+"#,
+    );
+    assert_eq!(out, "2,4");
+}
+
+/// Documents the hash half of issue #580, which this fix deliberately does NOT cover: binding a
+/// reference to a hash element is not implemented at all, so there is no alias to iterate (an
+/// explicit `$r = &$h['a'];` is rejected by the checker with "requires an indexed array").
+///
+/// The assertion pins the *current* wrong result on purpose. It exists so the borrowed element
```
nahime0: Please do not pin a known silent wrong result as the expected contract. The PR description uses `Fixes #580`, but #580 includes the hash-receiver case and no dedicated follow-up issue currently tracks it. Merging would therefore auto-close an issue whose reproducer still miscompiles.

Either diagnose this unsupported reference-binding shape (or fix it), or keep #580 open/create and link a follow-up and adjust the closing keyword. In either case, this test should not enshrine `1,2` as correct behavior.

docs/internals/the-ir.md
```diff
@@ -540,6 +540,7 @@ across a reset point.
 | `HashNew(key_type, value_type, capacity)` | none | `Heap(Hash)` | `alloc_heap` |
 | `ArrayLen`, `HashLen` | container | `I64` | `reads_heap` |
 | `ArrayGet` | array, index | element type | `reads_heap`, `may_warn`, maybe `may_fatal` |
+| `ArrayGetForWrite` | array, index (`I64`) | element type, **borrowed** | `reads_heap`, `may_warn` |
```
nahime0: Please align this table with `Op::default_effects()`. `ArrayGetForWrite` also has `writes_heap`, `writes_local`, `alloc_heap`, and `refcount_op`; omitting them contradicts both the implementation and the PR's stated ordering contract.

nahime0 (changes requested):
Requesting changes for one blocking lifetime regression plus two scope/documentation corrections.

The COW split fixes the covered cases, and the focused regression suites are green, but the borrowed element is not kept alive when the loop body replaces its parent. I reproduced this against the exact head: PHP and base `9f74f5300` print `1,2,done`, while this PR prints `1,done` with IR optimization both on and off. The inline comment includes the reproducer and the required lifetime invariant.

Please also avoid auto-closing #580 while its hash-receiver case remains silently wrong, and align the IR effects table with the implementation.

Operationally, the branch currently conflicts with live `main` in `CHANGELOG.md`; after the code changes and base refresh, CI needs to be terminal on the new exact head.

mirchaemanuel:
Thanks for the review — taking all three points. A short status so this does not sit silent while I work on them.

The branch is rebased onto current `main` (`dd5de0699`). The only conflict was `CHANGELOG.md`; every source file merged clean, and `cargo check --all-targets` is at 0 errors / 0 warnings on the new base — including against the `ReleaseUnlessAliases` op that landed meanwhile. I have not pushed it yet: the rebase will land together with the fixes below rather than as a separate force-push.

- **Blocking lifetime.** Reproducing your case first, then carrying an explicit owner acquired *after* the COW separation and releasing it on normal and early exits, with the heap-debug regression you asked for. I will check the mirror-image failure too — an owner taken and not released is exactly the symmetric bug.
- **Hash receiver.** I am going to diagnose and fix that shape rather than file a follow-up, so the closing keyword can stay honest and #580 closes on a complete fix. If it turns out to need a mechanism this PR should not grow into, I will say so and switch to the follow-up route instead of quietly narrowing the scope. Either way the test will stop pinning `1,2` as correct.
- **IR effects table.** Aligning it with `Op::default_effects()`.

I will come back with the reproducer output before and after, and the test numbers.

mirchaemanuel:
All three points are addressed. The branch is rebased onto current `main` (`e3959fe30`) and the head is `39db03fcc`.

### Blocking lifetime

Your reproducer, measured on the exact head, `--ir-opt` both ways:

| | before | after | `php` 8.4 |
|---|---|---|---|
| `--ir-opt=on` | `1,done` | `1,2,done` | `1,2,done` |
| `--ir-opt=off` | `1,done` | `1,2,done` | `1,2,done` |

The loop now takes a lifetime reference of its own on the source and drops it on every way out: its exit block covers normal termination and a plain `break`, and `emit_innermost_loop_cleanups` covers `break N`, `return`, and `throw`, where the pre-existing source cleanup is already released. A program exercising all four exits reports `allocs=19 frees=19, live_blocks=0, leak summary: clean` under `--heap-debug`.

Two details worth flagging, because both changed the shape of the fix:

**The acquire has to come after `IterStart`, not after the split in general.** `IterStart` is what runs `__rt_array_ensure_unique` for a by-reference source, so a reference held across it makes that split fire and hands the loop exactly the private copy this path exists to avoid.

**The pin does not survive the optimizer unmarked.** The paired acquire/release peephole cancels an `Acquire` whose only use is its `Release`, on the premise that the raised refcount in between is unobservable — the one premise a pin breaks by construction. That is why `--ir-opt=on` kept miscompiling after the pin was first added. `Acquire` now carries a `Bool` immediate marking it as a lifetime pin, and the peephole skips those pairs; there is a peephole unit test for it.

### Hash receiver

Fixed here rather than deferred, so the closing keyword stays honest.

The reason the hash half was excluded does not hold up: the checker's restriction is on `$r = &$h['k']`, which binds a hash slot to a local through `LoadArrayElemRefCell`. Fetch-for-write binds nothing — it separates the element and iterates the storage the parent already owns — so that unimplemented feature was never a blocker.

Measured the same way as the indexed case, on `$h = ['a' => [1, 2]]`: baseline 3 allocations, by-value `foreach` 3, by-reference `foreach` with an **empty body** 4. One copy, no user write involved.

`Op::HashGetForWrite` is the hash counterpart of `ArrayGetForWrite`; the two differ only in how the element slot is addressed. `__rt_hash_get` already builds the matching entry's address while probing and discarded it, so it now returns it (`x4` on AArch64, `r8` on x86_64, null on a miss). Both registers are caller-saved and were already clobbered by the helper's own probing and key comparisons, and no caller reads them after the call, so no existing contract changes. Everything downstream is shared with the indexed path.

End-to-end, `$h = ['a' => [1,2]]; $n = ['k' => ['x'=>1,'y'=>2]]; $i = [7 => [1,2]]` with a by-reference loop over each: before `1,2|1,2|1,2`, after `2,4|3,6|11,12`, which is what `php` prints. Heap clean.

The test that pinned `1,2` as expected no longer does: it asserts the mutation, and the section comment and docstring that declared the shape out of scope are corrected.

**Scope left out:** a hash whose value type is `Mixed` keeps the retaining read, symmetrically with the indexed `Mixed` element — that slot can hold an invoker ref-cell marker whose read materialises a fresh box rather than the slot's own storage.

### IR effects table

`ArrayGetForWrite` and `HashGetForWrite` now list exactly what `Op::default_effects()` assigns them: `reads_heap`, `writes_heap`, `writes_local`, `alloc_heap`, `refcount_op`, `may_warn`. The `Acquire` row documents the pin immediate, and the peephole section documents why a pin is exempt from cancellation.

### Tests

`cargo test --test codegen_tests by_ref_foreach` → **31 passed, 0 failed** on `39db03fcc`. That includes the new parent-replacement regressions (plain and nested), one per early exit, the hash regressions, and the pre-existing #556 guards. CI is the terminal check on this exact head.

Guikingone (approved):

nahime0 (approved):
Approve

PR #649, opened 2026-07-29
fix(codegen): guard print_r's container walk against the null-container sentinel (#647)

### LINKED ISSUE #647: print_r() on a null-container sentinel segfaults (no null guard in its lowering at all)
## Summary

`print_r()` on an Array-typed local carrying the null-container sentinel segfaults. Unlike `var_dump()`, its lowering has **no null guard at all** — not even the zero-pointer check that `var_dump` had — so it would also crash on a plain null pointer.

## Repro

```php
<?php
$a = [['x', 'y']];
$arr = $a[7];
print_r($arr);
echo "done\n";
```

PHP 8.4.23:

```
Warning: Undefined array key 7 in ... on line 3
done
```

(`print_r(null)` prints nothing at all — note this differs from `var_dump(null)`, which prints `NULL`.)

elephc (`main` `9f74f5300`, macOS ARM64):

```
Array
```
```
exit 139 — SIGSEGV, "done" never reached
```

The warning is correct, then `Array` is written and the following header load dereferences the sentinel.

## Cause

`emit_print_r_array` in `src/codegen/lower_inst/builtins/debug.rs` walks the container header with no branch for either null shape. The sibling defect in `emit_var_dump_array` — where a zero-only check let the non-zero sentinel through — is fixed in PR #646 by switching to the shared `sentinels::emit_branch_if_null_container` helper. `print_r` needs the same guard, but it cannot reuse that fix directly: PHP's output for the null case is *empty*, not `NULL`, so it needs its own branch target, its own regression test, and its own changelog entry.

## Expected

`print_r()` of a missed read should emit the undefined-key warning, print nothing, and let execution continue — matching PHP.

## Notes

Found while fixing #581. Verified on unpatched `main` and on the #581 branch: the crash is present in both, so it is independent of that fix.

mirchaemanuel:
Fixes #647. Based on current `main` `9f74f5300`.

Sibling of #646 (same file, same sentinel family) but a **different function and different PHP semantics** — see below. The two are independent; see the conflict note at the end.

## Root cause

`emit_print_r_array` (`src/codegen/lower_inst/builtins/debug.rs`) had **no null branch at all** — not even the zero-pointer check that `emit_var_dump_array` carried. So a missed element read wrote `Array\n` and passed the null-container sentinel straight to the walker as a live container.

Pre-fix assembly for the repro:

```
    ; print_r()
    ldur x0, [x29, #-96]          <- loads the sentinel
    str  x0, [sp, #-16]!
    ...                           <- writes "Array\n"
    bl   __rt_print_r_indexed     <- sentinel handed to the walker
```

No `cbz`, no comparison — so unlike #581 the crash happens inside the walker rather than at an inline header load.

## Fix, and why it is not #646's fix

The guard uses the shared `sentinels::emit_branch_if_null_container` (zero **or** sentinel) and jumps to a label placed **after** the walker call, skipping both the `Array\n` write and the walk.

It deliberately does not reuse the branch target added in #646, because PHP's semantics differ and I checked them before implementing:

```
php -r 'echo "["; print_r(null); echo "]";'                          -> []
php -r '$s = print_r(null, true); echo strlen($s);'                  -> 0
php -r 'var_dump(print_r(null));'                                    -> bool(true)
```

`print_r(null)` prints **nothing**, where `var_dump(null)` prints `NULL`. Skipping every write is also what makes all three call modes correct with one branch: echo mode prints nothing and still returns `true` (the immediate is loaded later, in `lower_print_r`), and the capture modes leave the buffer empty so `__rt_pr_finish` yields `""`.

## Tests

Eight regressions in `tests/codegen/io/printing.rs` — the natural home, next to the 20+ existing `test_print_r_*` cases. Six are behavioural and fail on the pre-fix build with `program crashed`; one guards the shapes that already worked; one asserts the emitted guard directly.

The assembly test was checked with an explicit negative control: reverting the source makes it fail on the host **and** under `ELEPHC_TEST_TARGET=linux-x86_64` **and** `linux-aarch64`, with `missing print_r null-container skip branch`. So x86_64 is genuinely verified, not assumed.

Sibling shapes, all matching PHP after the fix: hash receiver with an array value, hash receiver with a hash value (the `__rt_print_r_hash` walker — it crashed too), a miss through `?? null`, the **return mode** (`print_r($x, true)`), and the runtime-flag mode (`print_r($x, $flag)`). Return mode was the least obvious: it crashed *without* printing `Array`, because the text went into the capture buffer first, so its test asserts `len=0` rather than merely the absence of a crash.

A miss inside a boxed `mixed` was already correct before this change — the boxing-boundary normalization from #585/#591 covers it — and is left alone rather than claimed.

## Verification (macOS ARM64, `78a6a6590`)

| filter | result |
|---|---|
| `print_r` (incl. the 8 new) | 42 passed, 0 failed |
| same, `ELEPHC_IR_OPT=off` | 37 passed, 0 failed |
| asm test, `ELEPHC_TEST_TARGET=linux-x86_64` | 1 passed, 0 failed |
| asm test, `ELEPHC_TEST_TARGET=linux-aarch64` | 1 passed, 0 failed |
| `output_buffering` (print_r writes through ob) | 44 passed, 0 failed |
| `var_dump` | 28 passed, 0 failed |
| `sentinel` | 41 passed, 0 failed, 3 ignored |
| `array_miss` (#526) | 20 passed, 0 failed |
| `null_container` (#556/#592) | 2 passed, 0 failed |
| `autovivify` (#600/#592) | 26 passed, 0 failed |
| all three call modes under `--heap-debug` | `leak summary: clean`, exit 0 |

Narrow filters only, per the project's test policy; CI owns the full matrix. Linux not executed locally (no Docker), but the x86_64 lowering is covered by the assembly test above.

## Conflict note with #646

Checked with `git merge-tree` rather than assumed:

- `debug.rs` merges cleanly — #646 touches `emit_var_dump_array` (~line 535), this touches `emit_print_r_array` (~line 296), roughly 230 lines apart.
- tests do not collide: these live in `tests/codegen/io/printing.rs`, #646's in `tests/codegen/regressions/arrays.rs`.
- `CHANGELOG.md` conflicts, since both add a bullet under `## [Unreleased]`. Routine rebase resolution: keep both lines. Merge order does not matter.


mirchaemanuel:
**CI note — the three `Non-Codegen Tests` failures are not caused by this change.**

All three targets fail on the same unrelated test:

```
test eval_can_name_enum_cases_matches_php_case_numbering ... FAILED
  panicked at tests/eval_object_handle_tests.rs:112:5
  elephc compile failed:
  native project error: regex support requires managed native package pcre2
```

It is an `eval`/enum fixture failing on a missing managed native package — nothing to do with `print_r` or the sentinel guard. This branch is based on `9f74f5300`, which predates the commits that introduced and then fixed that dependency:

```
a69cd5c89 feat: add managed native dependencies
eec7496f7 ci: initialize native cache at step runtime
ac0cbd507 test: seed managed pcre2 for eval CLI fixtures
```

None of those is in `9f74f5300` (checked), and all three are in current `main`. So rebasing onto current `main` should clear it. I'm leaving the rebase to you rather than force-pushing.

For what it is worth locally, on this branch: `print_r_return_mode_heap_tests` 6/6, `builtin_parity_tests` 8/8, `var_export_and_strstr_result_tests` 20/20, plus the filters listed in the PR body.

mirchaemanuel:
Rebased onto current `main` (`61aeafe73`); head is `a1274a4db`. I said earlier that I would leave this rebase to you — it stayed conflicted for a week, so I did it rather than let it keep aging. Nothing but the base changed; the guard and its tests are untouched.

The only conflict was `CHANGELOG.md`, and only over placement: `var_dump`'s entry for the same sentinel guard (issue #581) landed meanwhile through #646, and this branch adds `print_r`'s. Both are kept, the `print_r` one on top. The diff against `main` is that single line.

This also clears the CI failure I reported here on July 29th. The three `Non-Codegen Tests` targets were failing on `eval_can_name_enum_cases_matches_php_case_numbering` for a missing managed `pcre2` package — a consequence of the old base `9f74f5300`, which predates `a69cd5c89`, `eec7496f7` and `ac0cbd507`. The new base contains all three.

Locally on the new head, filtered to what this touches: `print_r` **42 passed / 0 failed**, `var_dump` **34 passed / 0 failed** — the second because both sentinel guards now live in `builtins/debug.rs` and I wanted it on record that they coexist rather than assume it from a clean merge. CI on this exact head is the terminal check.

nahime0:
Issue #647 was closed automatically when #678 merged because that PR declared `Closes #647`, but the closure was erroneous: the merged tree does not contain the `print_r()` null-container guard or its regression tests, and the issue's exact reproducer still crashes on current `main` (`e47a8e9da`) with exit 139.

This PR therefore remains valid and necessary. I have reopened #647. Current `main` also merges cleanly into this branch; the resulting local head passes all 38 `test_print_r_` codegen tests plus the target-emission guard for Linux x86_64 and Linux AArch64. The merge is local and has not been pushed yet.

PR #659, opened 2026-07-31
feat: complete the OPcache plan and fix the PHP-parity defects auditing it uncovered

### LINKED ISSUE #660: Uncaught exceptions report no class, no message, and the wrong exit status
An uncaught `throw` prints one fixed 32-byte string regardless of what was thrown, and exits `1`:

```console
$ cat e.php
<?php
throw new RuntimeException("boom detail");

$ ./e
Fatal error: uncaught exception     # exit 1
```

Reference PHP 8.5.6 (`php -d xdebug.mode=off`):

```text
Fatal error: Uncaught RuntimeException: boom detail in /path/e.php:2
Stack trace:
#0 {main}
  thrown in /path/e.php on line 2     # exit 255
```

The class and the message are both discarded, so a production crash says nothing about what failed. It is not a plumbing problem: `_exc_value` already holds the Throwable at that point — `lower_throw_value` (`src/codegen/lower_term.rs`) publishes it immediately before calling `__rt_throw_current` — and the uncaught arm in `src/codegen_support/runtime/exceptions/throw_current.rs` simply never reads it, writing the constant `_uncaught_exc_msg` from `src/codegen_support/runtime/data/fixed.rs` instead.

The exit status is a separate, invisible half of the bug: a caller branching on `$?` gets `1` where PHP gives `255`.

## Scope note

The `in <file>:<line>` suffix and the stack trace are a **larger, separate gap**. `Throwable::getFile()`, `getLine()` and `getTrace()` are synthetic constants in `src/codegen/lower_inst.rs` — an empty string, `0`, and an empty array — so elephc has no throw-site origin to report at all. Reporting the class and message is achievable today; reporting the origin needs throw-site location tracking first.

## Measured PHP formatting

```text
Fatal error: Uncaught RuntimeException: boom detail in …    # class + ": " + message
Fatal error: Uncaught MyErr: custom text in …               # user subclass named
Fatal error: Uncaught Exception in …                        # EMPTY message: no colon
```

The empty-message case drops the `": "` separator entirely — the case a naive implementation gets wrong while still looking correct everywhere else.

Guikingone:
Closes the last outstanding item of the OPcache implementation plan (`--strict-opcache`) and the PHP-parity debts that auditing it against reference PHP 8.5.6 exposed.

Fixes #660. Every behaviour below was measured against `php -d xdebug.mode=off` and is byte-identical unless stated.

## Resource stringification

Converting a resource to a string **value** yielded an empty string in every form except direct output:

| form | before | PHP 8.5.6 |
|---|---|---|
| `"$r"` / `. $r .` / `(string)$r` / `strval($r)` | `""` | `Resource id #5` |
| `print $r` | `Resource id #5` | `Resource id #5` |

Three pre-existing gaps: `__rt_mixed_cast_string` had never implemented tag 9; `__elephc_eval_value_cast_string` re-implements its own dispatch instead of delegating, so `strval($r)` stayed empty inside a runtime-interpreted `eval()`; and the `fclose`/`pclose`/`closedir` sentinel stamped a bare `-1` into the Mixed payload, erasing the registry key so every display of a closed handle **minted a fresh id and stole the next `fopen()`'s number** — which made `print` and `var_dump` wrong for closed resources too.

Tag 9 returns borrowed `_concat_buf` scratch, matching tags 0/2/3-true. `implode` is the only caller that releases the result and `__rt_heap_free` ignores concat scratch by documented contract, so the arm can neither leak nor wild-free; an owning arm would have leaked at the other ~24 call sites.

## Eval hash contexts

`hash_init()` inside `eval()` consumed a resource id, shifting every later id by +1. In PHP 8 a HashContext is an **object**. The eval bridge forces the high payload word to zero, so every eval resource was born kind 0 and the kind-2 exclusion never applied. A new inert kind fixes it — tagging the context kind 2 instead would have routed `0x4000000000000000+n` into `elephc_crypto_free`, a wild free.

## Eval interpolation

The runtime eval interpreter never interpolated double-quoted strings, for **any** type: `echo "a:$i:"` printed `a:$i:`. The magician has its own lexer; `$` was never special and `next_token()` returns one token, so it could not structurally expand a string. Adds a real scanner covering `$v`, `{$v}`, `$a[0]`, `$a[k]`, `$o->p`, `\$v`, `$$x`, `{ $x }`.

`"$a[]"` stays **refused**, matching PHP's parse error — the main compiler lexer invents an empty key there, so porting it verbatim would have imported a measured over-acceptance. The AOT eval path was already correct and is untouched.

## Closed-resource type name

`var_dump()` of a closed resource reported its original type instead of `Unknown`, while the closed signal sat in a register at all four render sites, already consumed by the closed arm of `__rt_resource_id_of`.

## Fresh-result leak family

`tempnam()`, `getcwd()`, `microtime()` and `json_encode()` each allocate through `__rt_str_persist` and leaked that block on **every** call:

```
for ($i = 0; $i < 5; $i++) { $x = tempnam("/tmp", "p"); }
HEAP DEBUG: allocs=16 frees=11 live_blocks=5      # before
HEAP DEBUG: allocs=16 frees=16 live_blocks=0      # after
```

`value_is_scratch_string` only turns an `Op::Release` on a `Str` into a `__rt_heap_free_safe` when the callee declares `BuiltinResultOwnership::Fresh`; all four sat in the default `MayAliasArguments` bucket. The EIR was already correct (`release v2`, `own=maybe_owned`) and the emitted assembly contained **zero** `__rt_decref_any` — a correct IR paired with a no-op backend is why no suite could see it.

This is a recurring **family**: `result_ownership` already documents the identical mistake being fixed for `array_flip`, `print_r` and `strstr`.

Two corrections to the reported debt, both from measuring it: the leak is **per call**, not a constant 48 bytes, and `sys_get_temp_dir()` — named alongside `tempnam()` in the report — was already clean. It and `tmpfile()` are now pinned as controls so a future reading of that report cannot promote them to `Fresh`.

## `--strict-opcache`

D5 is the only divergence in this model that can silently change what a program **does** rather than what it **reports**: a forced `opcache_invalidate()` cannot reload code frozen into the binary, so a dev-mode cache-buster keeps running stale code with no signal. The new opt-in flag throws for exactly that request:

| `$force` | in manifest | PHP 8.5.6 | default | `--strict-opcache` |
|---|---|---|---|---|
| `false` | yes | `true` | `true` | `true` |
| `true` | no | `true` | `true` | `true` |
| `true` | yes | `true` | `true` | **throws** |

Narrowing it this way is what keeps the flag from being a blunt "reject opcache" switch that over-rejects valid PHP. A paired test compiles the same probe with and without the flag and requires the no-flag run to match reference PHP byte for byte.

## Uncaught exceptions (#660)

An uncaught throwable printed one fixed 32-byte string and exited `1`, discarding the class, the message and the origin — so a production crash told you nothing about what had been thrown:

| | before | after | PHP 8.5.6 |
|---|---|---|---|
| `throw new RuntimeException("boom")` | `Fatal error: uncaught exception` | `Fatal error: Uncaught RuntimeException: boom in /path/app.php:3` | identical |
| empty message | `Fatal error: uncaught exception` | `Fatal error: Uncaught Exception in /path/app.php:2` | identical |
| `$e->getFile()` / `getLine()` | `""` / `0` | the construction site | identical |
| exit status | `1` | `255` | `255` |

**The location is the CONSTRUCTION site, not the throw site.** This is the load-bearing design decision, and it is measurable: reference PHP reports line 2 for an exception built on line 2 and thrown on line 5. Attaching the origin to `Terminator::Throw` would have been wrong *and* would have required an EIR change — the `new` instruction already carries the span, so nothing in the IR moved. `uncaught_reports_the_construction_site_not_the_throw_site` is the test that would fail under the other design.

The line is stamped into offset 32 of the compact Throwable payload — the one word of the existing 56 bytes nothing ever wrote — so the allocation did not grow and no existing reader shifted. `__rt_heap_alloc` recycles blocks without zeroing them, so all 26 sites that allocate that payload now initialise the slot; the ones with no user `new` behind them write `0`, which the readers treat as "origin unknown" and omit rather than printing `:0`.

The file comes from `_script_source_file`, a per-module constant, because EIR spans carry a line and column but no filename. That is the same canonicalized string `__FILE__` already bakes, so it is not new information in the artifact. Code merged in from an `include` reports the including script's path — the documented limit of the approximation.

**A second uncaught path was found by the comparison.** Errors synthesized by codegen guards (`intdiv($n, 0)`) write their own fatal message and exit before the throwable is allocated, so they never reach `__rt_report_uncaught_exception` — and they still exited `1` after the main path moved to `255`. A script branching on `$?` got a different answer depending on which kind of exception escaped. Both now share one `UNCAUGHT_EXIT_STATUS`.

**The comment I had to delete was right.** `codegen/mod.rs` warned that embedding the source path "trips needle-based optimizer asm asserts", and it did: six optimizer tests searched the whole module text for a needle that also appears in the harness's own temp directory name — `!user_asm.contains("pow")` in a fixture compiled under `elephc_constant_folding_pow` matched the *directory*, not a surviving call. A static sweep of every `make_*_test_dir`/negated-`contains` pair in the tree found exactly those six and no others, so the fix is bounded rather than a 231-assertion rewrite: one helper strips the path bytes and nothing else, so the string literals several of those tests exist to prove were eliminated are still checked.

## Verification

`elephc --lib` 1011 · `elephc-magician --lib` 1127 · `error_tests` 1139 · `ir_backend_smoke_test` 255 · `codegen::runtime_gc` 253 · the 7 OPcache suites 67 · new leak suite 6 · interpolation 6 · resource suites 24 + 14 · `builtin_parity_tests` 8. All 0 failed.

Coverage is pinned on **both architectures** wherever assembly is involved. That is deliberate: an upstream leak fix was silently deleted from `implode.rs` on aarch64 *and* x86_64 during this campaign precisely because no test pinned the x86 side. Each change was sentinel-audited — neutralised, rebuilt, and confirmed to fail its own new tests while leaving every pre-existing suite and every control green.

## Not addressed here

- The uncaught-exception **stack trace**. PHP follows its first line with `Stack trace:`, `#0 {main}` and `  thrown in <file> on line <n>`; elephc keeps no call stack, so `getTrace()`/`getTraceAsString()` stay an empty array and an empty string rather than fabricating frames.
- Errors raised by a codegen guard carry no ` in <file>:<line>` suffix: their message is baked at emit time by a caller that passes no span, so there is no origin to report. PHP does print the operation's own line there.
- The Docker fixtures for PHP 8.2–8.5 and the Linux CLI+FPM re-capture the plan calls for are still outstanding — all verification here is macOS-aarch64.


Guikingone:
Pushed a follow-up that closes #660: **uncaught exceptions now report their class and message**, and exit with PHP's status.

Before, every uncaught `throw` printed one fixed 32-byte string and exited `1`, discarding both the class and the message — a production crash said nothing about what failed. `_exc_value` already held the Throwable at that point (`lower_throw_value` publishes it immediately before `__rt_throw_current`); the uncaught arm simply never read it.

Measured against `php -d xdebug.mode=off` (PHP 8.5.6):

| program | elephc | PHP |
|---|---|---|
| `throw new RuntimeException("boom detail")` | `Fatal error: Uncaught RuntimeException: boom detail` | same `+ in <file>:<line>` |
| user subclass `MyErr` | `Fatal error: Uncaught MyErr: custom text` | same |
| empty message | `Fatal error: Uncaught Exception` | same — **no colon** |
| exit status | **255** (was `1`) | 255 |

The class name is resolved through the same `_class_name_entries` table `get_class()` and `var_dump()` read, which is why a user-declared subclass is covered — it is the case that proves the table lookup rather than a compile-time literal.

**Deliberately not emitted**: the ` in <file>:<line>` suffix and the stack trace. `Throwable::getFile()`, `getLine()` and `getTrace()` are synthetic constants (empty string, `0`, empty array), so elephc has no throw-site origin to report; a fabricated one would be worse than none. The tests assert a **prefix** rather than equality for exactly that reason — asserting equality would freeze the gap as intended behaviour.

Both arches tail-call one shared `__rt_report_uncaught_exception`, pinned by unit tests on each, including one asserting that both arms *delegate* rather than re-inline — inlined copies are how the two architectures drifted apart in `implode.rs` earlier in this campaign.

Sentinel-audited: bypassing the report fails 3 of the 4 end-to-end tests and one unit pin, while the "caught exception prints nothing" control correctly keeps passing.

Also addressed both review findings: `docs/compiling/cli-reference.md` now has a `--strict-opcache` row, and `CHANGELOG.md` has entries under `[Unreleased]`. I did not add a `README.md` example — `--ini`, the closest sibling and also opcache-specific, is not listed there either, so I followed the file's actual curation rather than the literal rule.

CI was green (114/114) on the previous head before this was folded in.

nahime0 (approved):
There was just a misaligned comment on `render_scripts_map_literal`

I've fixed it while aligning the PR to main. 

This will be shipped in v0.26.3

PR #667, opened 2026-08-04
feat/parse url

nahime0:
## Summary

Adds PHP-compatible `parse_url()` on both the AOT compiler path and the Magician `eval()` runtime, plus the predefined `PHP_URL_*` constants.

- **AOT**: registry builtin (`src/builtins/string/parse_url.rs`) → `RuntimeFnId::ParseUrl` → Mixed-returning `__rt_parse_url` scanner (AArch64 + x86_64) with associative-array and component-selector results, Fresh ownership, and catchable `ValueError` for invalid selectors.
- **Magician**: `eval_builtin!` home file with a pure-Rust scanner aligned to PHP 8.4 `ext/standard/url.c` quirks (scheme-relative URLs, IPv6 brackets, empty present components, control-byte substitution, port range checks).
- **Constants**: `PHP_URL_SCHEME` … `PHP_URL_FRAGMENT` (0–7) wired through checker, prescan, name resolver, autoload interpreter, Magician predefined constants, README, and docs.
- **Parity**: shared fixture corpus in `tests/fixtures/parse_url_cases.json` exercised by Magician unit tests and AOT codegen tests; coverage also includes component shapes, named/callable call sites, invalid URLs (`false`), selector `ValueError`, and heap-debug ownership.

## Behavior notes

| Call | Success | Missing component | Invalid URL |
| --- | --- | --- | --- |
| `parse_url($url)` / negative component | assoc array of present keys (`port` as int) | key omitted | `false` |
| `PHP_URL_PORT` | `int` | `null` | `false` |
| other `PHP_URL_*` | `string` (may be `""`) | `null` | `false` |

Selectors greater than `PHP_URL_FRAGMENT` raise:

`parse_url(): Argument #2 ($component) must be a valid URL component identifier, N given`

Any **negative** component selects the full-array form (PHP 8.4). Invalid URLs return `false` even when the selector would otherwise be invalid.

## Test plan

- [ ] `cargo build`
- [ ] `cargo test --test codegen_tests parse_url`
- [ ] `cargo test --test error_tests parse_url`
- [ ] `cargo test -p elephc-magician parse_url`
- [ ] `git diff --check`
- [ ] Optional: `ELEPHC_PHP_CHECK=1 cargo test --test codegen_tests parse_url`
- [ ] CI matrix: macos-aarch64, linux-aarch64, linux-x86_64

## Docs / examples

- `docs/php/strings.md`, generated builtin pages, internals runtime note
- `examples/parse-url/main.php`
- `CHANGELOG.md` `[Unreleased]` entry


nahime0:
@greptile-apps review this

docs/php/builtins/string/parse_url.md
```diff
@@ -0,0 +1,37 @@
...
+- **`eval()` (magician interpreter)**: supported — declarative interpreter builtin ([`crates/elephc-magician/src/interpreter/builtins/string/parse_url.rs`](https://github.com/illegalstudio/elephc/blob/main/crates/elephc-magician/src/interpreter/builtins/string/parse_url.rs)).
+
+_No examples yet — check `examples/` and `showcases/` for usage patterns._
+
+
+
+
+
+
+
```
Guikingone, line 34: Extra lines to remove

nahime0: We have this pattern in every builtin reference markdown. It's a "bug" of the generator. Those are placeholders for examples, see also and notes. We should collapse extra empty spaces.

I'll track this behaviour in a specific issue (#681).

PR #680, opened 2026-08-06
fix(types): assign class and interface ids deterministically

Guikingone:
## Problem

Compiling the **same unchanged source twice** produced different assembly.

On a 30-class fixture: **105 490 differing lines out of 371 914 — about 28% of the output.**

Class and interface ids are handed out while walking `class_map` and `interface_map` (`src/types/checker/driver/mod.rs`), both `HashMap`s. `HashMap` iteration order is randomized per process, so every run assigned different ids.

Those ids are not internal bookkeeping. They reach the emitted assembly three ways:

- symbol names — `_class_propinit_<id>`, `_class_vtable_<id>`, `_class_serprop_<id>`, …
- immediates — 14 392 occurrences of `mov x10, #<id>` in that one fixture
- `.quad` values in the runtime metadata tables

## Impact

**Nothing is miscompiled.** Each binary is self-consistent, and the ids only ever have to agree with themselves.

What breaks is *reproducibility*:

- two builds of the same source cannot be compared, so "did this change alter codegen?" is not answerable by diffing
- any content-addressed caching of build artifacts is impossible — a key computed from generated assembly can never hit

The second point is what surfaced this: it blocks incremental/`--watch`-style work on the build pipeline.

## Fix

Sort both walks before assigning ids. That makes the assignment a pure function of the declared names.

## Tests

`tests/compiler_determinism_tests.rs` compiles a class-heavy, an interface-heavy and an enum-heavy fixture three times each, asserting the emitted assembly is byte-identical.

Two details that matter and are commented in the file:

- **Each compilation runs in its own process.** A single process reuses one `HashMap` seed, so an in-process pair would not exercise the difference.
- **All runs compile the same file in the same directory.** The canonicalized source path is baked into the output (`__FILE__`, `Throwable::getFile()`), so compiling sibling copies would report a path difference as a reproducibility failure. This was caught while writing the test — the first version failed on exactly one line for that reason.

Verified the tests actually catch the bug: with the sorts removed they report ~8 900 differing lines; with the fix, all three pass.

## Verification

| suite | result |
|---|---|
| `--test compiler_determinism_tests` | 3 passed |
| `--test codegen_tests instanceof` | 32 passed |
| `--test codegen_tests enum` | 78 passed |
| `--test codegen_tests classes` | 78 passed |
| `--test codegen_tests interface` | 102 passed |

Also checked by hand: three consecutive compilations of `showcases/doom` (32 files, inheritance, interfaces, enums) now emit identical assembly, where before they did not.

## Note

This does not prove the compiler is reproducible overall — it fixes the source measured here. Other unordered walks may remain in codegen or metadata emission. A repo-wide reproducibility gate would be the way to establish that, and is out of scope for this PR.

### DIFF
--- a/src/types/checker/driver/mod.rs
+++ b/src/types/checker/driver/mod.rs
@@ -207,7 +207,12 @@ pub(super) fn check_types_impl(
 
     let mut next_interface_id = 0u64;
     let mut building_interfaces = HashSet::new();
-    let interface_names: Vec<String> = interface_map.keys().cloned().collect();
+    // Sorted: `interface_map` is a HashMap, whose iteration order is randomized per
+    // process. Interface ids are handed out in this order and are baked into the
+    // generated assembly, so an unsorted walk makes two compilations of the SAME
+    // source produce different output — which defeats any content-addressed cache.
+    let mut interface_names: Vec<String> = interface_map.keys().cloned().collect();
+    interface_names.sort();
     for interface_name in interface_names {
         if let Err(error) = build_interface_info_recursive(
             &interface_name,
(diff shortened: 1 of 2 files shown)


nahime0 (approved):

nahime0:
Looks good to me. Deterministic output makes builds easier to compare, cache, and reason about.

PR #691, opened 2026-08-08
Build every stdlib prelude in Rust, and make class registration pay-for-use

Guikingone:
Every injected standard-library prelude now builds as AST in Rust instead of being spliced together as PHP text and reparsed on each compile, plus the pay-for-use work that fell out of measuring what that text costs.

## Benchmarks

Measured against `origin/main` (`a7a0e090c2`) on the same machine, same samples, same binary flags.

**Compile phases**

| | origin/main | this branch | |
|---|---:|---:|---|
| `Configuring PDO support` (a PDO program) | **1.29 s** | **2.09 ms** | the prelude is built, not parsed |
| `Checking types` (trivial program) | 77.26 ms | 14.90 ms | SPL + Reflection registration is pay-for-use |
| `Checking types` (PDO program) | 179.40 ms | 166.90 ms | |
| Total (trivial program) | 592.83 ms | 438.29 ms | |
| Total (PDO program) | 9.01 s | 8.08 s | assembly still dominates |

**Emitted assembly** (deterministic — these are line counts, not timings)

| sample | origin/main | this branch | |
|---|---:|---:|---:|
| `<?php echo 1;` | 9,957 | 6,952 | **−30%** |
| one `imagecreatetruecolor` call | 191,952 | 9,654 | **−95%** |
| `hash('sha256', …)` | 11,966 | 8,739 | −27% |
| `var_export(…)` | 20,529 | 14,788 | −28% |
| `new PDO("sqlite::memory:")` | 951,505 | 909,615 | −4% |

Two honest notes on this table. The wall-clock rows are single runs on a machine that was under memory pressure, so treat the totals as indicative and the line counts as exact. And the PDO row is the one that barely moves: the eleven-driver surface upstream added compiles to ~950k lines of method bodies, 15% of it a single `PDO::sqliteCreateCollation` that no program in the sample ever calls. That is a dispatch-ladder problem, not a prelude problem, and it is untouched here.

## What lands

**The preludes build in Rust.** hash, tz, var_export, version, list_id, PDO, image, OPcache and the `--web` request surface. PDO is the hard one: upstream expanded it to eleven drivers across seven PHP profiles with seven optional drivers gated by cargo features and environment variables, so it is built for **all 63 profile/driver combinations**.

The PHP source stays, under `cfg(test)`, as the oracle each build is compared against. A transcription with nothing to check it is a claim, not a fact — so `built_declarations_match_the_php_for_every_profile` parses the text each profile ships and compares it to the built AST node by node, across the whole matrix.

**Registration becomes pay-for-use.** SPL, Reflection and the builtin throwables are no longer registered for programs that cannot reach them. The authority for what may be dropped is not a reading of the codegen but the `_*_class_id` symbol table: a runtime helper can only stamp a class id it has a symbol for, so a throwable without one has no unreferenced producer.

**Two silent failures made loud.** `_class_parent_ids` used the same `-1` for "true root class" and "class never emitted", so a `catch` could fail to match without a word; an unemitted class now reads `-2` and the walk aborts with a diagnostic. And the tokenizer was revalidating the whole remaining source as UTF-8 on every lookahead — lexing was quadratic in file size.

**Two tools relearn the prelude's notation.** `gen_image_api_stubs.py` splices throwing stubs for the unimplemented Imagick/Gmagick surface into `image_prelude.rs`; `check_extern_exports.py` matches the prelude's bridge declarations against the staticlib's exports. Both read the file as PHP text. The generator now reads `class("Imagick")` / `method("x")` and emits builder calls, reusing every transcription rule that decides a *signature* untouched — deriving them a second time is how the two drift. Its output comes back **byte-identical** to the hand-transcribed stubs, so the only change to the prelude is the sixteen marker comments that delimit its region.

Placing those markers is where this nearly went wrong quietly. Identifying the generated region by the body every generated stub has — a lone throw of `"... is not supported in elephc"` — swallows eleven **hand-written** methods that also throw that sentence, `Imagick::annotateImage` among them (`"requires FreeType text, which is not supported in elephc"`). Inside the region they stop counting as implemented, and the generator re-emits them under their spec spelling, replacing a specific diagnostic with a generic one. The boundaries are instead read from the PHP this branch replaces, which still carries the markers; the recovered counts match it class for class (Imagick 64 implemented / 294 generated, Gmagick 49 / 98, and so on).

## The bug worth reading

Switching PDO's injection to the built prelude broke all 276 PDO tests, **while the AST oracle compared the two programs equal across 63 cells and the printer rendered them as 3,837 identical lines**.

The experiment that settled it: same patch, injection back on the parsed path — compiles. So the difference was the switch, not the content, and therefore lived in what neither comparison carries.

Spans. `CheckResult::builtin_call_types` maps a builtin call's inferred type to the span of the call, and lowering reads it back. Every node the builders produced carried `Span::dummy()`, so every builtin call in a prelude shared one key and lowering fell back to the builtin's **declared** return type.

**And the declared type was wrong — that is the actual defect.** Six builtins have a check hook returning `PhpType::Pointer` or `PhpType::Callable` — a raw descriptor, not a boxed cell — while declaring `mixed`, because `TypeSpec` was `Int|Float|Str|Bool|Mixed|Void` and had no variant for either. A check hook makes the declared type *non-authoritative*, not unused: every path that cannot name a call falls back to it, and this one handed codegen a boxed cell for an address. `TypeSpec::Ptr` and `TypeSpec::Callable` let those six declare what they return.

The claim is measured, not argued: **with `Span::dummy()` put back on every builder, all 276 PDO tests pass.** Distinct spans stay for one narrower reason — a new debug assertion compares a builtin's declared type against the checker's, and can only compare them where the checked one is findable; under `dummy()` the map is skipped for every prelude call, so the next mismatched declaration would go unwitnessed exactly there.

**How the six were found is the part worth keeping**, because four were not enough. The first sweep grepped for `Ok(PhpType::Pointer` and found five; fixing them left 25 tests failing, which read as *"the span buys precision"* and would have shipped that way. The two it missed return a **variable** holding the pointer type, and a *callable* rather than a pointer — neither matches that spelling. Reading every hook in the directory instead of grepping for one is what closed the last 25.

The same lie had a third home: a hand-curated table in the docs generator patched `zval_pack` to `pointer` while `ptr`, `ptr_null` and `ptr_offset` documented `mixed`. All four now derive from the declaration, and the override is gone.

## A span is a key in four places

The declared-type fix removes the miscompile, but it does not make a shared key safe. So every span-keyed lookup in the compiler was **enumerated rather than recalled**: `builtin_call_types`, `throw_access_sites`, `loop_storage_types`, and the call-type memo inside loop-storage stabilisation. Two were already handled; **the other two were broken by this branch**, because the preludes they key used to be PHP text with real line numbers and are now built nodes carrying `Span::dummy()`.

Measured while type-checking one `new PDO("sqlite::memory:")` program:

| collision | count | consequence |
|---|--:|---|
| loops inheriting another loop's storage contracts | **26** | comparing each against what that loop would compute: **all 26 differed** — one got two contracts where it needed none, another one where it needed three |
| loop-storage call memo answering for a different call | **42** | every method call in a prelude loop body shared one key, so the first call's inferred type was returned for all |

Loop statements now get distinct spans and the memo refuses a span that names no node; both counts are **zero** afterwards.

`throw_access_sites` gets a recording guard instead, because it is written only on an access violation: the fix is to refuse a key that cannot name the offending node. A readonly violation inside a compiler-generated prelude is a compiler bug, and filing it under the shared dummy span would make every *other* prelude property write throw `Cannot modify readonly property`. The PDO prelude does declare `readonly string $queryString` and does write it — that is one `internal_pdo_statement_initializer` special case away.

**"Synthetic" also stopped having two spellings.** `line == 0` was the only test for "compiler-generated" before synthetic spans existed, and now answers wrongly for half of them. `Span::is_from_source` and `Span::identifies_a_node` name the two different questions. The statement-boundary concat reset asks the first, and would otherwise have started firing for prelude loops the moment they got distinct spans.

## The docs table had 188 duplicated keys

Chasing the parameter half of the declared-versus-checked gap led into `PARAM_TYPES`, the hand-curated table that refines what the Rust registry declares coarsely. **188 of its 788 keys were declared twice** — a later batch re-listed builtins the section-organised part already covered. Python keeps the last binding and drops the earlier line without a word, so the dead line is not clutter: **editing it is a no-op that reads exactly like a fix**, and the two copies of a key sit hundreds of lines apart in a 1,300-line file.

181 pairs agreed. **Seven disagreed**, stating a value nobody had applied in a long time (`array_push`, `array_unshift`, `call_user_func`, `count`, `is_a`, `is_callable`, `str_ireplace`). The table now holds one entry per builtin — the live value, written at the organised position — and `audit_builtins.py` fails on a duplicate, verified by injecting one and watching it exit 1. Regenerating the docs before and after the dedup touches the same file set, which is the proof that removing 195 lines changed no page.

The sibling table had the milder form of the same problem: **29 of the 59 `RETURN_TYPE_OVERRIDES` entries only restated what the Rust declaration already said.** An override that agrees is not neutral — it pins the old answer silently on the day the declaration changes, which is precisely what `zval_pack` had been doing while the source could not spell `pointer`. Removing them regenerates to a byte-identical set of pages, so all 30 that remain carry information.

Two smaller things came from the same look. The eighteen pointer-area builtins whose check hook requires a pointer now **declare** one, correcting three internals pages that documented `mixed $pointer` and leaving the fifteen public pages byte-identical (a hand-written override had been compensating). Only the pointer position moved: `offset`, `length`, `value` and `string` stay `Mixed`, because a declared scalar parameter reaches `classify_param_binding` and can have a cast inserted, while `Ptr` reaches its `_ => Rejected` arm exactly as `Mixed` does.

**And the thing that did not need fixing.** 44 builtins have a `PARAM_TYPES` entry longer than their declared parameter list, and `array_walk` really is rejected with three arguments while php.net documents `$userdata`. The generated pages are nevertheless correct — the renderer only refines parameters that exist and ignores the surplus, so `array_walk(array $array, callable $callback): void` is what ships. Reporting 44 doc bugs would have been reporting 44 doc bugs that do not exist.

## Verification

Green: PDO codegen **276** in release and again in **debug**, where the new assertions are live and never fire; lib **1301 + 1146**; image **165**. `cargo check --all-targets` clean, zero warnings. The 63-cell oracle passes, and the generated builtins docs, their audit and the site-compatibility validator all pass on the regenerated output.

**Not run locally: `codegen_tests io`, `web_tests`, and the four opcache suites.** Three attempts at `io` were *killed*, not failed — the machine was down to ~60 MB of free RAM with another build running, and halving the test threads did not help. Flagging it rather than leaving it to look covered: earlier in this work a `codegen_tests opcache` filter reported `ok. 0 passed` because it selected no tests at all, which reads exactly like a pass. CI runs all three, and they pass there.

## Notes for review

- `src/pdo_prelude/build.rs` is generated by `synthetic_class::transcribe` from the parse of the PHP it replaces, with conditionals added by hand only where a profile or driver differs — 20 edit points for the version axis, 5 insertion points for the driver axis, out of 188 declarations. The rest is mechanical.
- The parsed-prelude cache is removed along with the parsing. It cached a template that a one-program-per-process compiler reads exactly once.
- `OptionalDrivers` replaces an ambient `cfg!`/environment read inside `prelude_source_for_version`, which had made every driver combination but the build's own untestable.


Guikingone:
Rebased onto `main` (69 commits) and addressed the review.

**Greptile's finding was valid**, and the arm next to it had the same hole.

`FirstClassCallable(StaticMethod)` recorded nothing, so `Imagick::queryFormats(...)` let the pruner delete the class the program was about to call — surfacing as `Undefined class: Imagick`, pointing at user code rather than at the pass that removed the declaration.

The sibling `FirstClassCallable(Method)` arm was wrong too, and hid better: it scanned the *receiver*, so it looked handled. But the receiver's class is unknown before name resolution, which is exactly why the ordinary method call roots on the *method name* — and the deferred form was dropping it. Both arms now root what calling them would root.

Two regression tests, and I checked they go **red** without the fix — a regression test that passes beforehand proves nothing.

**The rebase had one conflict worth describing.** `0a9d03d4d3` added object and enum-case rendering to `var_export` in the PHP source, after this branch had converted that prelude to builders. Either side of that conflict loses something real: ours drops the new feature, theirs drops the conversion. So neither was taken — the PHP `main` now ships was transcribed again, which keeps both. Verified end-to-end against `php -n`: `\P::__set_state(...)`, `\Status::Active`, `enum(Status::Active)` and `print_r` of an object all match line for line.

The rebase also carried a semantic conflict git could not flag: `with_parse_mode` now takes a `SourceProfile` rather than a `SourceMode`. Built preludes are stamped with `SourceProfile::new`, i.e. PHP's default coercive typing, which is right — a built prelude declares no `strict_types=1` and must not inherit whatever the caller was parsing.

lib **1296** green. The three suites I could not run locally (`codegen_tests io`, `web_tests`, opcache) are unchanged from the earlier note — still CI's call.

Guikingone:
Superseded by #694, which now carries this work unchanged as its first commit (`6e67e0fd0d`) plus the symbol-table strip on top.

Nothing is lost: #694 was rebased onto this branch rather than the reverse, so the commit here is byte-identical there, and the two changes compose — a `<?php echo 1;` binary goes from 182 680 bytes on `main` to 93 896 (−49 %), the preludes accounting for the metadata and the strip for the symbol table.

The `feat/rust-synthetic-classes` branch is left in place.

PR #695, opened 2026-08-09
refactor(core): harden security boundaries and runtime performance

Guikingone:
## Summary

- remediate the security and performance findings reported by the independent Kimi K3 audit
- harden PHAR parsing, `unserialize()`, generation-safe buffers, runtime-cache identity, include-depth handling, recursion/fiber boundaries, PDO callbacks, TLS state, and web handler isolation
- add compile-time web process models: plain `--web` keeps the original in-process `worker` hot path, while `pool` and `request` are explicit isolation opt-ins
- implement ownership-safe `ksort()` / `krsort()` default `SORT_REGULAR` behavior with stable `O(n log n)` hash-link sorting on ARM64 and x86_64
- add `get_object_vars()` and object-to-array conversion support required by the audited parity cases
- add TDD regression coverage for reproduced findings and refresh generated builtin documentation

## Security and runtime hardening

- bound archive decompression, PHAR manifests, signatures, and actual decoder output
- validate serialized input before allocation or hook execution, isolate reentrant decoder state, and make option failures catchable `TypeError`s
- replace raw buffer pointers with generation-checked handles and dedicated stale/exhaustion diagnostics
- preserve recursion budgets across exceptions, callbacks, eval boundaries, and fiber switches
- harden runtime-cache publication/pruning and include every emission feature, including PDO UDF adapters, in cache identity
- prevent pre-dispatch isolated-web cancellation from accumulating stale broker IDs

## Web process models

- `worker` remains the default and is exactly equivalent to plain `--web`; it preserves the original in-process throughput and process-global native state
- `pool` uses persistent supervised handler children, bounded concurrency, recycling quotas, and crash/timeout replacement without per-request fork cost
- `request` uses one disposable tracked/reaped child per request for the strongest process-global reset, accepting fork/COW/page-fault overhead
- broker-backed modes stream framed responses and support exact request cancellation, body/write deadlines, handler concurrency, and descendant-safe shutdown

## Array and object semantics

- return `bool` from `ksort()` / `krsort()` and preserve default `SORT_REGULAR` key ordering for numeric strings, large decimals, whitespace, mixed integer/string boundaries, and embedded-NUL-safe strings
- preserve COW and ownership for locals, properties, nested packed/hash arrays, boxed `Mixed` cells, and repeated key sorts
- promote non-empty packed `krsort()` receivers to integer-keyed hash storage instead of rejecting them
- replace the stable quadratic hash insertion sort with an allocation-free stable bottom-up merge sort
- keep PHP's optional sort-flag argument explicitly out of scope and tracked in #699
- expose `get_object_vars()` through the builtin registry and support object/mixed array casts on every supported target

## Maintainability

- split the monolithic `unserialize` emitter into format, validation, allocation, hook, and target-specific modules
- split runtime-cache orchestration from identity/integrity and lease/pruning policy
- split web request supervision, pool supervision, and child request execution
- isolate specialized key-sort by-reference place lowering from the generic rewrite

## Validation

- merged current `main` (`441480da1`) into the branch
- `cargo build`
- runtime-cache unit regressions: **9/9 passed**
- `elephc-web` crate tests: **81/81 passed**
- web end-to-end tests: **63/63 passed**
- web session end-to-end tests: **49/49 passed**
- key-sort codegen tests on macOS ARM64: **31/31 passed**
- key-sort codegen tests on Linux x86_64: **31/31 passed**
- key-sort codegen tests on Linux ARM64: **31/31 passed**
- key-sort diagnostics: **3/3 passed**
- assembly comment/alignment, `cargo fmt --check`, and `git diff --check`: passed
- builtin docs export/regeneration completed; docs audit: **0 errors**; site compatibility: **1038 pages validated**
- exact-head CI for `ccf56cd32` is pending

The full local suite was intentionally not run; CI remains responsible for the complete sharded supported-target matrix.


nahime0:
The PR originally made fork-per-request isolation the only execution model for `--web`. After looking at the performance impact - especially at concurrency 1 - I decided to change the approach.

Plain `--web` now keeps the original in-process worker path from `main`, with no additional request-time branching or IPC. The stronger isolation models remain available as compile-time opt-ins:

- `--web-isolation=pool` uses persistent, supervised handler processes.
- `--web-isolation=request` creates a disposable process for every request.

I think this gives us a better balance. `worker` remains the right default for throughput-sensitive applications, `pool` provides crash containment and additional concurrency without paying the fork cost on every request, and `request` is available when fully resetting native process-global state is more important than performance.

I also tightened the isolated implementations while making this change: handlers now have explicit PID and request-ID tracking, deterministic reaping, disconnect cancellation, execution and backpressure timeouts, acknowledged descriptor transfers, crash replacement, and complete descendant cleanup.

The release benchmarks show no regression for the default worker path compared with `main`. On the trivial handler benchmark, `pool` retained about 51% of worker throughput at concurrency 1 and 79% at concurrency 8; fork-per-request remains significantly more expensive, as expected.

| Client concurrency | `main` | `worker` | `pool` | `request` |
|---:|---:|---:|---:|---:|
| 1 | 14,858 req/s | 15,034 req/s | 7,641 req/s | 3,184 req/s |
| 8 | 25,539 req/s | 25,639 req/s | 20,171 req/s | 7,189 req/s |

Compared with the new `worker` mode:

| Client concurrency | `pool` throughput | `request` throughput |
|---:|---:|---:|
| 1 | 50.8% | 21.2% |
| 8 | 78.7% | 28.0% |

These measurements used release builds on macOS ARM64, one web worker, `--handler-concurrency 8` for the isolated modes, and a trivial echo handler driven by `hey`. The concurrency-1 values are medians from three 3-second runs; concurrency 8 is one 3-second run. The trivial handler intentionally emphasizes the fixed process/IPC overhead, so the relative difference should decrease as application work per request increases.

I updated the documentation as well, with concrete guidance and deployment examples for choosing each mode, including PDO/process-global lifetime and the lack of request affinity between pool children.

nahime0:
One other area I changed while reviewing this PR was the unserialize() runtime emitter. It had grown into a single 3,822-line file combining policy parsing, diagnostics, validation, decoding, context management, object hydration, and both target implementations, which made it increasingly difficult to review safely.

I split it into 11 cohesive modules under system/unserialize/, separating shared orchestration from the AArch64 and x86_64 policy, validator, decoder, and storage paths. The two recursive decoders remain slightly above the 500-line guideline because they are still cohesive leaf implementations; the surrounding orchestration and support modules are now below it.

This is intentionally a source-organization refactor only. I captured the generated assembly before the split and compared it afterward for macOS ARM64, Linux ARM64, and Linux x86_64: all three outputs are byte-for-byte identical. The focused unserialize() tests also pass on every supported target, including the deep-recursion security regression.

nahime0:
I also aligned ksort() with krsort() for nested array<mixed>[int] lvalues, including COW, write-back, TypeErrors, and regression coverage. Full $flags support is tracked separately in #699

nahime0 (approved):

PR #711, opened 2026-08-13
fix(ir_lower): a closure wrote through a `use (&$x)` capture as if the reference had a type

Guikingone:
A closure stored through a `use (&$x)` capture as if the reference carried a type. It does not — a PHP reference holds whatever was last written to it — so the value came back reinterpreted, silently and with no diagnostic.

```php
$b = 5;
(function () use (&$b) { $b = null; })();
var_dump($b);          // php: NULL    before: int(9223372036854775806)
```

That number is the raw null sentinel read as an integer.

## The failure took a different form for every pair of types

| program | php | before |
|---|---|---|
| `$a = 5;` → closure writes `9` | `int(9)` | `int(9)` ✅ |
| `$b = 5;` → closure writes `null` | `NULL` | `int(9223372036854775806)` |
| `$c = null;` → closure writes `7` | `int(7)` | `NULL` — the write never landed |
| `$d = "s";` → closure writes `null` | `NULL` | garbage bytes |

The first row is the reason this survived: **every existing by-ref test wrote back the type the variable already held.** Nothing in the suite ever crossed types through a reference.

It also mattered beyond its own shape. This is the exact channel that made a compound-assign hoist gate look sound during an earlier audit: the jury's probe was `use (&$k)` writing an int into an int, which works, so the channel was cleared when it should not have been.

## Cause

The EIR says it in one line. For `use (&$b)` where `$b` holds an int:

```
function __eir_closure_main_0(b: I64 php=int)
    v0: I64 php=null = const_null
    store_ref_cell v0 slot[0]
```

and the caller reads it back with `load_ref_cell` typed `int`. The reference cell carries the type the captured variable happened to hold **at capture time**, and both the store inside the closure and the read after it are typed by that. Any other type is reinterpreted through the slot.

## Fix

The rule was already in the tree, for a narrower case: a by-ref capture whose closure body contains `eval` was already widened to `Mixed`, on exactly this reasoning — the body can store anything through the reference, so the cell must be able to hold anything. `eval` is not what makes that true; **any write does.** This generalises the existing arm rather than adding a second mechanism beside it.

## Only captures the closure WRITES

An earlier revision of this PR widened every by-reference capture. That was correct but too broad, and CI caught it: it widened the enclosing local too, and several backend paths need a concrete type.

```
unsupported EIR backend feature: preg_replace_callback subject for PHP type Mixed
```

The trigger was elephc's own synthetic `RegexIterator` body, whose closure captures `$subject` by reference and only ever passes it to `substr()`. Eleven SPL tests that had been compiling stopped, across three linux-aarch64 shards.

A capture the closure only reads keeps its exact type now. `body_writes_local` decides, and **its polarity is the whole point**: every shape it cannot rule out answers "written", and both matches are exhaustive with no wildcard arm, so a new AST node is a compile error here rather than a silent "not written".

One refinement the measurement forced: counting every bare `$name` argument as a write left all eleven red, because that closure passes `$subject` to `substr()`. The builtin registry carries `ref_params` per position, so a bare argument counts only where the callee takes that parameter by reference. Methods, closures, dynamic targets and anything the registry does not carry stay conservative.

## The cost was not what this PR first claimed

An earlier revision reported the widening as roughly two extra allocations per call for a same-type accumulator, and framed it as a tax on working code. **That framing was wrong.**

```php
$total = PHP_INT_MAX - 1;
$add = function (int $n) use (&$total): void { $total += $n; };
```

| call | php | before |
|---|---|---|
| first | `int(9223372036854775807)` | `int(9223372036854775807)` |
| second | `float(9.2233720368548E+18)` | `int(-9223372036854775808)` |
| third | `float(9.2233720368548E+18)` | `int(-9223372036854775798)` |

PHP promotes an overflowing sum to float, so `$total += $n` stores `int|float` — elephc's own inference types it exactly that way, `Mixed` unless it can prove the addition cannot overflow. A cell typed from the captured `int` cannot hold it, and the sum wrapped around silently.

So the narrow cell was not cheaper-and-equivalent. It was cheaper **because it was wrong**, and the extra block is the price of representing what PHP actually stores. This row needs no `null` and no string anywhere — just arithmetic leaving the integer range, which makes an accumulator the shape most likely to meet it. Pinned by its own test.

**Deliberately not done:** sparing a same-type accumulator by comparing written types. The existing syntactic inference answers `PhpType::Int` for everything it does not recognise — `Variable` has no arm at all — which is the opposite polarity from what this gate needs. It would take a second type inference, one that must agree with the lowering's or the cell is wrong again. That is the failure mode this branch fixes, not a place to add one.

## Verification

- The four reachable type pairs and the overflow row, asserted on `var_dump` output rather than a cast — a value that merely *prints* as correct cannot pass.
- **695 of 696** tests in the closure, callable and by-ref families passed on the first run with the widening in place. The one failure was a row of the new test, not existing behaviour: an int-to-string case that the checker refuses in ordinary straight-line code too (`$e = 1; $e = "grown";` → `cannot reassign $e from int to string`). It pinned an unrelated gradual-typing over-rejection, so it was removed and the reason recorded in the test. Reassigning to `null` *is* allowed, which is what leaves the four rows above reachable.
- The full local suite, not just the families I assumed were affected — that assumption is what let the SPL regression through. codegen 2178 + 2177 + 2181, error_tests 1289, `--lib` 2658, smoke 254: **10 487 tests, no behavioural failure**. All ten batch timeouts replayed one at a time and passed in 4–23s.
- Zero warnings, doctests green, generated docs in sync.

Measured on macOS-aarch64; CI covers Linux ARM64 and x86_64.


Guikingone:
Thanks — the changelog point is right, and measurably so. The entry was 1651 characters against 470 and 318 for the two beside it, and more to the point it described the *investigation* (EIR signature, allocation counts, why an optimisation was deferred) rather than what changed for someone writing PHP. `AGENTS.md` asks for user-facing bullets. Trimmed to 531 characters in a6b3708, keeping the wrong behaviour, the corrected one, and the allocation cost — that last one is a real cost users pay, so it stays. The mechanism and the measurements live in the commit message and this description, where they already were.

One correction on the file overview, since a reviewer reading the summary would be misled by it. It says of `src/ir_lower/expr/closures.rs`: *"no concrete correctness defect was established."* The defect is established by measurement against `php -n`, and the table is in the description above:

```php
$b = 5;
(function () use (&$b) { $b = null; })();
var_dump($b);          // php: NULL    before: int(9223372036854775806)
```

That value is the raw null sentinel read as an integer. Two further pairs fail differently — writing an int into a variable holding `null` never lands, and writing `null` over a string yields garbage bytes — and all four rows are pinned by `test_by_ref_capture_writes_back_a_different_type` in this PR, asserted on `var_dump` output rather than a cast so a value that merely *prints* as correct cannot pass. Reverting `closures.rs` alone turns that test red.

The change is also not a new mechanism: the same widening already applied to by-ref captures whose closure body contains `eval`, on exactly this reasoning. `eval` is one way to store an unpredictable type through a reference, not the only one — any write is.

nahime0:
At first sight the PR is ok, waiting the full CI run before merging

nahime0 (approved):

PR #715, opened 2026-08-13
fix(strings): read a negative substr/substr_replace length as bytes omitted from the end

Guikingone:
A negative `$length` means "omit this many bytes from the end" in PHP. `substr()` and `substr_replace()` read it as an error instead, and returned a plausible-looking string with no diagnostic.

| call | php | before |
|---|---|---|
| `substr("hello", 1, -1)` | `"ell"` | `"ello"` |
| `substr("hello", 0, -2)` | `"hel"` | `""` |
| `substr("hello", -4, -1)` | `"ell"` | `"ello"` |
| `substr_replace("hello", "X", 1, -1)` | `"hXo"` | `"hX"` |

## Two faults, and the first is why the second stayed hidden

`-1` doubled as the in-band marker for "no length argument". An explicit `substr($s, 1, -1)` was therefore indistinguishable from the two-argument call and kept the whole tail. Every *other* negative length took a different path and was clamped to zero, returning the empty string where PHP returns a prefix.

The two shapes fail in **opposite directions**, which is worth stating because it defeats a casual probe: testing only `-1` concludes "keeps too much", testing only `-2` concludes "returns nothing". Neither alone describes the defect.

## The fix removes the sentinel rather than moving it

Whether a length was *passed* is known from the operand count at compile time, so it is never encoded in the length's own value. `substr` emits the clamp only when a length was given.

`substr_replace` cannot do the same: its omitted-length signal is consumed by the runtime helper, which cannot see the operand count. It now receives `i64::MAX`. The helper bounds the length by the remaining tail **before** adding it to the offset, so a saturating value runs through the subject end by the ordinary path with no sentinel test — and that same bound is what keeps `offset + length` from overflowing. `-1` is then free to mean what PHP means by it.

## Re-measured before reimplementing

Both fixes already existed, on `backup/pre-squash-merge` and `backup/sprintf-and-compliance-*`, marked done in my notes and **never merged** — `git merge-base --is-ancestor` says neither is an ancestor of `main`.

Rather than replay the cluster, each claim was measured:

| stranded commit | claim | measured on `main` |
|---|---|---|
| `5648b925b` | `substr` negative length | 🔴 live |
| `be9b2a997` | `substr_replace` negative length | 🔴 live |
| `347964176` | `array_slice`/`array_splice` negative length | ✅ **dead** — landed by another route |
| `ccbae7dc9` | `filesize`/`filemtime` return `false` | 🔴 live, tracked separately |

Three of four live, one dead. Replaying the cluster wholesale would have added noise to `array_slice`, which someone fixed differently in the meantime. Reimplemented rather than cherry-picked: those commits carry a `Co-Authored-By` trailer that `AGENTS.md` forbids.

`filesize`/`filemtime` are left out of this PR deliberately — they overlap a stream/stat branch already in flight, and `filemtime("/absent")` currently returns uninitialised stack memory (`int(6125002896)`), which deserves its own change and its own witness.

## Verification

- Twelve `substr` rows and ten `substr_replace` rows against `php -n`. The expected strings in the tests were taken from **php's own output**, not from what elephc printed.
- Half of each fixture is controls, on purpose: the two-argument form, a zero length, a length past the end and an offset past the end all share the code path the sentinel removal touched.
- The whole `strings::search` family passes, **45/45**, including the pre-existing `substr` and `substr_replace` tests.
- Zero warnings, doctests green, generated docs in sync.

Measured on macOS-aarch64; both architectures were edited, and CI covers Linux ARM64 and x86_64.


Guikingone:
Valid, and applied wider than suggested — the block was wrong on **all four** lines, not the two flagged.

Read off the caller (`lower_substr_replace_x86_64`), which pops the offset, then the replacement pair, then the subject pair, and confirmed against the helper's own spill slots (`mov QWORD PTR [rbp - 8], rax` — "preserve the subject string pointer"):

| line | was | is |
|---|---|---|
| subject | `rdi/rsi` | `rax/rdx` |
| replacement | `rdx/rcx` | `rdi/rsi` |
| offset | `r8` | `rcx` |
| length | `r9` | `r8` |

So `rcx` and `r8` are right for the two lines you flagged, and the subject/replacement lines needed the same treatment. Fixed in 64fa11f, with a note that the assignment is elephc's own rather than System V argument order — that assumption is presumably what let a reader skip the check in the first place.

For the record on provenance: my negative-length commit rewrote the *descriptions* on two of those lines while keeping their register names, so it carried a pre-existing error forward rather than introducing one. The AArch64 block directly above is correct and always was, which is how the x86 one rotted unnoticed — the sibling doc reads right, so nobody cross-checks the other.

Documentation only; `strings::search` still passes 45/45.

nahime0 (approved):

PR #723, opened 2026-08-16
fix: parse assignment in operand position

### LINKED ISSUE #722: Assignment as the right operand of a comparison is rejected — breaks Composer's ClassLoader
# Assignment as the right operand of a comparison is rejected — breaks Composer's ClassLoader

**Version:** elephc 0.26.3 (Homebrew, `illegalstudio/tap`)
**Platform:** macOS 15 (Darwin 25.6.0), aarch64
**PHP target:** default (8.5)

## Summary

elephc's parser rejects an assignment appearing as the right operand of a
comparison operator — `false !== $pos = strrpos($s, '/')`. This is valid PHP and
is explicitly permitted by the language, but elephc reports `Invalid assignment
target` and then loses sync with the enclosing class body, producing a cascade
of spurious errors to the end of the file.

This idiom appears in `vendor/composer/ClassLoader.php`, so **every Composer
project currently fails to compile**, regardless of what the application code
looks like.

## Minimal reproduction

```php
<?php
$s = 'a/b';
if (false !== $pos = strrpos($s, '/')) {
    echo $pos;
}
```

```
$ php bug1.php
1

$ elephc bug1.php --check
error[bug1.php:3:11]: Invalid assignment target
error[3:38]: Expected ';'
error[5:1]: Unexpected token at statement position: RBrace
```

## Expected

Parses and type-checks. PHP's grammar permits an assignment on the right-hand
side of a comparison even though `=` has lower precedence than `!==`; the manual
documents this on the operator-precedence page ("Although `=` has lower
precedence than most other operators, PHP still allows expressions similar to
the following: `if (!$a = foo())`").

## Root cause

The reported column (3:11) points at the `!==` operator, and the message is
`Invalid assignment target` — so the parser appears to bind the comparison
first and then treat the result as the assignment's left-hand side:

```
(false !== $pos) = strrpos($s, '/')     // what elephc parses
false !== ($pos = strrpos($s, '/'))     // what PHP parses
```

This is not specific to `!==`. The same mis-parse occurs whenever an assignment
appears as the right operand of any higher-precedence binary operator:

| Expression | elephc |
|---|---|
| `if (false !== $p = strrpos($s, "/"))` | `Invalid assignment target` |
| `if (null !== $p = getenv("X"))` | `Invalid assignment target` |
| `while (false !== $p = strrpos($s, "/"))` | `Invalid assignment target` |
| `$x = 1 + $y = 2;` | `Invalid assignment target` |
| `if (false !== ($p = strrpos($s, "/")))` | OK |
| `if ($p = strrpos($s, "/"))` | OK |

The last two show the parser handles assignment-in-condition fine when it is
parenthesized or is the whole condition — only the "assignment as right operand"
position is affected.

## Impact

`vendor/composer/ClassLoader.php` uses this idiom three times (lines 500, 522
and 551 in Composer 2.8):

```php
while (false !== $lastPos = strrpos($subPath, '\\')) {
if (false !== $pos = strrpos($class, '\\')) {
if ($this->useIncludePath && $file = stream_resolve_include_path($logicalPathPsr0)) {
```

Compiling any file that pulls in `vendor/autoload.php` therefore fails with 14
errors before reaching application code:

```
error[vendor/composer/ClassLoader.php:500:26]: Invalid assignment target
error[500:64]: Expected ';'
error[522:19]: Invalid assignment target
error[529:9]: Unexpected token at statement position: RBrace
error[561:5]: Unexpected token at statement position: Private
error[561:13]: Expected variable after 'static'
...
```

Since the README advertises compile-time Composer PSR-4/PSR-0/classmap
autoloading, this is likely blocking that feature for most real projects.

Note also that the file cannot be worked around by excluding it: elephc resolves
the Composer autoload graph itself, so a broken vendor file is pulled in via a
`use` statement in another file even when it is not `require`d.

## Secondary issue: error recovery

After the first error the parser does not resynchronize at a statement or class
member boundary — it emits `Unexpected token at statement position: Private`,
`Expected variable after 'static'` and unbalanced `RBrace` errors for the rest of
the file. The 12 follow-on errors are all artifacts of the single mis-parse on
line 500. Recovering at the next `}` or member declaration would make the real
cause much easier to spot. (Possibly related to the closed checker-recovery
issues #612 / #597, but this one is in the parser.)

## Workaround

Parenthesize the assignment:

```php
while (false !== ($lastPos = strrpos($subPath, '\\'))) {
```

For Composer this has to be reapplied after every `composer install`, so it is
not viable for real use.

s2x:
## Summary

Fixes #722.

PHP spells assignment as `variable '=' expr`. Because the left-hand side must be
a `variable` rather than an arbitrary expression, an assignment is grammatical
wherever a variable has just been parsed — including positions where operator
precedence alone would have ended the expression. That is why PHP parses

```php
false !== $pos = strrpos($s, '/')
```

as `false !== ($pos = strrpos($s, '/'))`, and why the manual notes on the
operator-precedence page that `if (!$a = foo())` is allowed despite `=` binding
looser than `!`.

The Pratt loop stopped on binding power alone, so by the time the assignment
operator was reached the comparison had already been folded into the left-hand
side. `is_assignment_expression_target()` then saw `false !== $pos` and reported
`Invalid assignment target`.

The fix keeps the loop going when the left-hand side is a valid assignment
target. That predicate already mirrors PHP's `variable` non-terminal — it admits
`Variable`, `PropertyAccess`, `DynamicPropertyAccess`, `StaticPropertyAccess`
and `ArrayAccess` over those — so the condition expresses the grammar rule
directly. The binding-power table is unchanged.

```diff
-            if l_bp < min_bp {
+            if l_bp < min_bp && !is_assignment_expression_target(&lhs) {
                 break;
             }
```

A non-variable left-hand side still takes the existing error path, so
`(1 + 2) = 3` and `foo() = 3` remain rejected.

## Why it matters

Composer's generated `vendor/composer/ClassLoader.php` uses this idiom three
times (lines 500, 522 and 551 in Composer 2.8):

```php
while (false !== $lastPos = strrpos($subPath, '\\')) {
if (false !== $pos = strrpos($class, '\\')) {
if ($this->useIncludePath && $file = stream_resolve_include_path($logicalPathPsr0)) {
```

Before this change, compiling anything that reaches `vendor/autoload.php` failed
with 14 errors inside Composer's own autoloader, well before any application
code. After it, `ClassLoader.php` parses cleanly end to end; the first
diagnostic on that file is now the deliberate compile-time-constant-path
restriction on `require $file` at line 576, which is a separate matter.

## Tests

Parser (`tests/parser_tests/expressions/assignments.rs`):

* `test_assignment_as_right_operand_of_comparison` — asserts the AST shape is
  `false !== (assignment)`, not an assignment to the comparison.
* `test_assignment_as_right_operand_of_arithmetic` — `$x = 1 + $y = 2;` nests the
  assignment under the `+`.
* `test_assignment_to_non_variable_still_rejected` — `(1 + 2) = 3` and
  `foo() = 3` continue to fail, so the relaxation does not admit assignments to
  arbitrary expressions.

Execution (`tests/ir_backend_smoke_test.rs`), covering runtime semantics rather
than just parse shape:

* `ir_backend_evaluates_assignment_as_right_operand_of_comparison` — prints `3`.
* `ir_backend_evaluates_assignment_as_right_operand_of_arithmetic` — prints `32`,
  pinning the order in which the two assignments happen.

All five outputs were checked against `php` on the same snippets.

`cargo test --lib` is green (1251 + 1379 across the module suites, 0 failures)
and `cargo fmt --check` is clean.

Two pre-existing failures on `main` are unrelated to this change and reproduce on
an unmodified checkout:

* `codegen_support::abi::tests::basics::test_emit_frame_prologue_rejects_undersized_frame`
  hangs and is eventually SIGKILLed.
* `cargo clippy` fails in `elephc-builtin-contract` with
  `approximate value of f{32,64}::consts::E found`.

### DIFF
--- a/src/parser/expr/pratt.rs
+++ b/src/parser/expr/pratt.rs
@@ -360,7 +360,7 @@ fn parse_expr_bp_inner(
         }
 
         if let Some((op, l_bp, r_bp)) = assignment_bp(&tokens[*pos].0) {
-            if l_bp < min_bp {
+            if l_bp < min_bp && !is_assignment_expression_target(&lhs) {
                 break;
             }
 
--- a/tests/ir_backend_smoke_test.rs
+++ b/tests/ir_backend_smoke_test.rs
@@ -6981,3 +6981,25 @@ fn unique_test_id() -> u128 {
         .expect("system time should be after unix epoch")
         .as_nanos()
 }
+
+/// Verifies that an assignment used as the right operand of a comparison evaluates with PHP
+/// semantics: the assignment happens first and its value is what the comparison sees.
+#[test]
+fn ir_backend_evaluates_assignment_as_right_operand_of_comparison() {
+    let source = "<?php $s = \"a/b/c\"; if (false !== $pos = strrpos($s, \"/\")) { echo $pos; }";
+    assert_eq!(
+        compile_and_run_ir_backend("assign_in_comparison", source),
+        "3"
+    );
+}
+
+/// Verifies that a chained assignment inside an arithmetic expression assigns the inner
+/// variable and folds its value into the surrounding sum, matching PHP.
+#[test]
+fn ir_backend_evaluates_assignment_as_right_operand_of_arithmetic() {
+    let source = "<?php $x = 1 + $y = 2; echo $x; echo $y;";
+    assert_eq!(
+        compile_and_run_ir_backend("assign_in_arithmetic", source),
+        "32"
+    );
+}
--- a/tests/parser_tests/expressions/assignments.rs
+++ b/tests/parser_tests/expressions/assignments.rs
@@ -236,3 +236,60 @@ fn test_parse_union_typed_assign() {
         other => panic!("Expected typed assign, got {:?}", other),
     }
 }
+
+/// Verifies that an assignment used as the right operand of a comparison binds tighter
+/// than the comparison, so `false !== $pos = strrpos($s, "/")` parses as
+/// `false !== ($pos = strrpos($s, "/"))` rather than as an assignment to the comparison.
+#[test]
+fn test_assignment_as_right_operand_of_comparison() {
+    let stmts = parse_source("<?php if (false !== $pos = strrpos($s, \"/\")) { echo $pos; }");
+    let condition = match &stmts[0].kind {
+        StmtKind::If { condition, .. } => condition,
+        other => panic!("expected If, got {:?}", other),
+    };
+    match &condition.kind {
+        ExprKind::BinaryOp { left, op, right } => {
+            assert_eq!(op, &BinOp::StrictNotEq);
+            assert_eq!(left.kind, ExprKind::BoolLiteral(false));
+            assert!(
+                matches!(&right.kind, ExprKind::Assignment { .. }),
+                "expected the comparison's right operand to be an assignment, got {:?}",
+                right.kind
+            );
+        }
+        other => panic!("expected BinaryOp, got {:?}", other),
+    }
+}
+
+/// Verifies that an assignment as the right operand of an arithmetic operator binds tighter
+/// than that operator, so `$x = 1 + $y = 2;` parses as `$x = 1 + ($y = 2)`.
+#[test]
+fn test_assignment_as_right_operand_of_arithmetic() {
+    let stmts = parse_source("<?php $x = 1 + $y = 2;");
+    match &stmts[0].kind {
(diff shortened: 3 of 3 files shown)


nahime0:
@Guikingone for me merge and forget if the CI is green

Guikingone (approved):

nahime0 (approved):

PR #730, opened 2026-08-18
feat: add PHP ext/curl on pinned libcurl

nahime0:
## Summary

First-class PHP `ext/curl` on every supported target. Programs get the full PHP 8.2–8.5 function, class, and constant surface on a checksum-pinned libcurl 8.21.0. There is no system, Homebrew, or distro `-lcurl`.

```bash
elephc native add curl
elephc app.php
# or: elephc app.php --with-curl
```

`extension_loaded('curl')` is true only when the bridge is linked. A curl-free program never declares the classes, never links `elephc_curl`, and never needs the native package.

## What shipped

- **35 functions, 6 classes, 689 constants.** 260 of 271 `CURLOPT_*` options work; the other 11 (blobs and three leftover callbacks) return `false` plus PHP's warning. Never an inert `true`.
- **Easy, multi, and share**, including PHP 8.5 `curl_multi_get_handles`, `curl_share_init_persistent`, and `CurlSharePersistentHandle` (gated by `--php-version`).
- **`CURLFile` / `CURLStringFile` / `curl_file_create`**, with array `CURLOPT_POSTFIELDS` as real `multipart/form-data`.
- **Six callbacks** (`WRITE` / `HEADER` / `READ` / `PROGRESS` / `XFERINFO` / `DEBUG`) through the existing runtime callable invoker. Throws abort the transfer, stay catchable, and never unwind through libcurl.
- **Stream options** `CURLOPT_FILE` / `INFILE` / `WRITEHEADER` / `STDERR` on the AOT path.
- **The same surface inside `eval()`**, except those four stream options (honest `false` + warning).
- **Pinned native stack:** libcurl 8.21.0, OpenSSL 3.5.8 (TLS for libcurl only — `openssl_*` / `hash()` stay on RustCrypto), zlib 1.3.2, nghttp2 1.70.0, libssh2 1.11.1.
- **25 protocols**, including HTTP/2, SCP/SFTP, and the usual mail/messaging set. Not LDAP, not HTTP/3.
- **Runtime CA discovery** (`$CURL_CA_BUNDLE`, then the baked path if it still exists, then seven root-owned distro locations), applied to both `CURLOPT_CAINFO` and `CURLOPT_PROXY_CAINFO`. Explicit user options replace or compose as in a stock libcurl.

User-facing contract and every documented PHP divergence: `docs/php/curl.md`.

## Out of this PR

- Bundled CA material for `FROM scratch` / distroless images.
- HTTP/3 (needs ngtcp2 + nghttp3 on a later curl pin).
- LDAP / LDAPS (catalog cannot yet declare system-library needs).
- `CURLOPT_FILE` / `INFILE` / `WRITEHEADER` / `STDERR` inside `eval()`.

This is **not** the v0.28 Zend `curl.so` consumer PoC. `extern "curl"` stays user FFI.

## Tests / CI

- Dedicated `curl-codegen-tests-*` jobs (4 shards × 3 targets) run the ~194 codegen fixtures after `elephc native install --locked`. They fail if the managed-native skip gate fires.
- `managed-native-smoke` builds the curl stack from source and compiles `examples/curl-get`.
- Catalog / Magician / prelude signature audits run in `curl-feature-contract`.

## Review notes

PHP 8 objects (`CurlHandle`, …) come from an injected prelude over internal `__elephc_curl_*` builtins, same pattern as `HashContext`. The bridge owns libcurl handles by never-reused `i64` ids. A few signatures throw instead of returning `T|false` because the checker cannot narrow those unions — listed in the docs, not silent.


crates/elephc-curl/src/ca.rs
```diff
@@ -0,0 +1,274 @@
...
+/// `configure`'s FIFTH probe, `/usr/local/share/certs/ca-root-nss.crt`, is DELIBERATELY
+/// NOT HERE, and it is the one place this list is knowingly narrower than curl's own.
+/// `/usr/local` is the Homebrew prefix on Intel macOS and is left writable by the admin
+/// user there, so a probe under it is a writable-location fallback — the exact thing the
+/// security posture for this feature rules out — and honoring it would let anything that
+/// can write a Homebrew prefix choose a process's trust anchors. Modern FreeBSD, the
+/// system that path is really for, populates `/etc/ssl/cert.pem` (entry 6) via `certctl`
+/// anyway; a FreeBSD host that only has the `ca_root_nss` file needs an explicit
+/// `CURLOPT_CAINFO` or `$CURL_CA_BUNDLE`.
+pub(crate) const CANDIDATE_CA_FILES: &[&str] = &[
```
nahime0, line 134: **[bug]** CA discovery only probes Unix distro PEM files (`/etc/ssl/certs/ca-certificates.crt`, `/etc/ssl/cert.pem`, …). A cross-built libcurl bakes no `CURL_CA_BUNDLE` at all (`configure` skips detection when `--host=` is set). iOS has none of those files and OpenSSL does not read the Security.framework keychain, so a verified HTTPS transfer on `ios-arm64` / `ios-sim-arm64` fails closed with no trust anchors unless the app sets `CURLOPT_CAINFO` or `$CURL_CA_BUNDLE`. Catalog, generated builtin docs, and `BRIDGES` all advertise those targets as first-class; `docs/php/curl.md` never mentions iOS; CI has no iOS curl compile/link coverage (only macos/linux executable shards). That is an unsupported path that is neither isolated nor documented, against the supported-target policy.

**Suggestion:** Either (a) implement an iOS trust path (Security.framework / app-bundled CA) and add iOS compile+link tests, or (b) isolate iOS curl as unsupported: refuse at compile/link with a diagnostic, keep catalog/docs honest, and document that iOS needs an explicit CA bundle if you still ship the archive. Do not leave generated docs claiming iOS support while default HTTPS cannot work.

nahime0: Resolved on `a5aae7602`.

- The managed curl recipe enables `--with-apple-sectrust` and disables baked CA file/path defaults for both `ios-arm64` and `ios-sim-arm64`.
- The bridge detects the `AppleSecTrust` feature and leaves `CURLOPT_CAINFO` unset, preserving Security.framework verification; explicit CA overrides keep their documented libcurl behavior.
- `docs/php/curl.md` now documents native iOS trust and override semantics.
- The dedicated device and simulator CI jobs build the native chain and Rust bridge, assert the AppleSecTrust symbols, compile a curl-using PHP static library, and complete the final host link. Both jobs are green on this SHA.

This fully addresses the original iOS trust-store finding.

nahime0 (commented):
## Follow-up review

The iOS CA trust-store blocker is resolved on `a5aae7602`: the managed curl build now uses Apple SecTrust on device and simulator, the bridge preserves native trust unless the application supplies an explicit CA override, and the behavior is documented. Dedicated iOS compile/final-link jobs verify the SecTrust-enabled archives and are green for both targets.

The original thread is resolved. GitHub currently reports that the branch conflicts with `main`; that is separate from this finding and still needs reconciliation before merge.

PR #733, opened 2026-08-19
feat: add PHP-compatible PCNTL support

nahime0:
## Summary

- add PCNTL support across the shared builtin contract, EIR lowering, target-aware code generation, runtime helpers, Magician, and the optional `elephc-pcntl` bridge
- implement `pcntl_fork()`, child waits/status decoding, `pcntl_exec()`, priorities, Linux CPU affinity/namespaces, Darwin QoS, `posix_setpgid()`, and `posix_setsid()`
- add Elephc's `pcntl_daemon()` convenience API over `daemon(3)`; hide it under `--strict-php`
- implement PHP-compatible signal registration, dispatch, masks, alarms, async signals, siginfo, synchronous waits, restart-syscall defaults, exception abort, reentrancy, and Fiber-switch rejection in both AOT and eval paths
- route compiled and eval signal records through backend-owned queues so one backend cannot consume the other's incompatible callable records; use backend-specific `sigaction` trampolines to keep registration changes race-free
- preserve delivery counts after self-pipe saturation with preallocated atomic overflow slots, and clear inherited pipe/overflow records after `fork()` or daemonization
- preserve omitted-vs-explicit-empty `pcntl_exec()` environments for positional, named, spread, AOT, and Magician calls
- support macOS AArch64, Linux AArch64, and Linux x86_64; refuse PCNTL on iOS and reject reachable PCNTL process/signal operations from hosted `cdylib`/`staticlib` exports
- add focused bridge, AOT/eval, target-safety, hosted-library, GC, named-argument, Fiber, overflow, strict-mode, and diagnostic regressions; publish examples, the PCNTL guide, and regenerated builtin references

## Testing

- `cargo build`
- `cargo clippy -p elephc-pcntl --lib -- -D warnings`
- `cargo test -p elephc-pcntl --lib -- --test-threads=1` — 25 passed
- `cargo test -p elephc-magician strict_mode` — 2 passed
- `cargo test --lib pcntl_runtime` — 3 passed
- `cargo test --test codegen_tests pcntl` — 60 passed on Linux x86_64
- `cargo test --test error_tests pcntl` — 19 passed
- `cargo test --test error_tests posix` — 2 passed
- `cargo test --test error_tests strict_php` — 30 passed
- `cargo test --test cdylib_tests test_host_process_builtins_are_refused_for_ios_targets`
- `cargo test --test cdylib_tests test_hosted_libraries_reject_reachable_pcntl_operations`
- compiled and ran `examples/pcntl/main.php` and `examples/pcntl-daemon/main.php`
- `cargo build --example gen_builtins`
- `python3 scripts/docs/extract_builtins.py --render --force`
- `python3 scripts/docs/audit_builtins.py`
- `python3 scripts/docs/elephc_builtins/validate_site_compat.py`
- `python3 scripts/audit_builtin_eir_boundary.py --enforce-target-architecture`
- assembly-comment alignment checks for every touched codegen/runtime emitter
- `git diff --check`

CI remains authoritative for the complete macOS AArch64, Linux AArch64, and Linux x86_64 matrix.

## Runtime note

Compiled and eval callable descriptors remain backend-private. Pending records are now routed to the backend that registered the active handler, so dispatching in the other backend leaves them untouched. When a pipe overflows, every delivery count is replayed; repeated overflow records for one signal use the latest captured siginfo snapshot. Standard non-realtime signals can still be coalesced by the operating system before the native handler runs.


nahime0:
@Guikingone can you review this?

Guikingone:
@nahime0 On it 👍🏻

Guikingone (commented):
## Review — PCNTL extension (`aae00ac84`)

I built this branch and the merge base `20fa89093`, ran 39 differential programs against php 8.5.6
(macOS ARM64, pcntl compiled in), then put the code in front of three adversarial jurors and
re-measured every load-bearing claim myself before repeating it here.

The process-control half is in good shape. The findings cluster in signal introspection, in
argument typing, and in one return-value path that corrupts silently.

Provenance is marked on each finding: **verified** means I wrote a fresh probe and ran it;
**relayed** means it comes from a juror and I have not re-run it.

A methodology note for anyone re-running this: the machine's `php.ini` loads Xdebug, which
rewrites `var_dump` output. Every `var_dump` case is a false positive unless you run `php -n` —
`pcntl` and `posix` are compiled into this build, so `-n` keeps them.

---

## F1 — a pcntl result returned from an **untyped closure** collapses to a wrong integer

**Severity: blocker. Verified.** Silent value *and* type corruption.

```php
$b = function () { return pcntl_strerror(22); };
var_dump($b());
```
```
php     string(16) "Invalid argument"
elephc  int(0)
```

Same shape, different builtin: `function () use ($s) { return pcntl_signal($s, SIG_IGN); }` gives
php `bool(true)` and elephc `int(1)`.

The discriminators are what make this precise, and all four hold:

| variant | result |
|---|---|
| `function () { return pcntl_strerror(22); }` | **`int(0)`** ✗ |
| `function (): string { return pcntl_strerror(22); }` | correct ✔ |
| `var_dump(pcntl_strerror(22));` direct | correct ✔ |
| `function () { return strtoupper("ab"); }` | correct ✔ |

So it is neither a general closure defect nor a general `pcntl_strerror` defect — it is the pcntl
return value on the *inferred* return path. An untyped closure is the most ordinary callback shape
in PHP, nothing warns, and the wrong values are plausible enough to travel (`int(0)`, `int(1)`).
`pcntl_strerror` is also the thinnest-covered builtin in the PR's own tests, which is exactly where
this landed.

## F2 — a runtime-valued bad handler name kills the process with an uncatchable fatal

**Severity: important. Verified.**

```php
$hn = "nosuchfn";
try { pcntl_signal(SIGUSR1, $hn); } catch (Throwable $e) { echo get_class($e); }
echo "still alive\n";
```
```
php     TypeError: pcntl_signal(): Argument #2 ($handler) must be of type callable|int, string given
        still alive
elephc  Fatal error: Call to undefined function <dynamic>()      (exit 1, catch never runs)
```

A handler name coming from config or user input turns a catchable php error into process death,
and the message names neither the function nor a line.

## F3 — `function_exists()` and `PHP_OS` guards do not gate compilation

**Severity: important. Verified.** Portable sources cannot be written.

```php
if (function_exists('pcntl_getcpu')) { echo pcntl_getcpu(); } else { echo "no-getcpu\n"; }
```
```
php     no-getcpu
elephc  error[2:45]: pcntl_getcpu() is not available for the macos-aarch64 target
```

The sharp part: `function_exists('pcntl_getcpu')` on its own correctly const-folds to
`bool(false)` — I checked — so the checker proves the branch dead and then hard-errors inside it
anyway. A `PHP_OS === 'Linux'` guard fails the same way and `PHP_OS_FAMILY` is undefined, so no
portable-source pattern exists for the platform split this PR itself introduces and documents at
`docs/php/pcntl.md:147`.

## F4 — AOT and eval disagree about the same source: signal-set arrays

**Severity: important. Verified.**

```php
pcntl_sigprocmask(SIG_BLOCK, ["9"]);        // php: bool(true)
```
```
elephc, compiled     error[2:39]: parameter $signals must be an indexed integer array, Array(Str) given
elephc, inside eval  bool(true)             ← identical source, matches php
```

Same for `["k" => SIGTERM]`. These are correct php programs, not error cases. Compile-time
strictness is defensible for provably-wrong code, but here it rejects working code *and* the two
backends contradict each other about it.

## F5 — `call_user_func` by-reference: AOT writes through silently, eval warns

**Severity: important. Verified, including the attribution.**

```php
$status = 0;
call_user_func('pcntl_waitpid', $pid, $status);
printf("status=%d\n", $status);
```

| | result |
|---|---|
| php | `Warning: … must be passed by reference, value given`, `status=0` |
| elephc compiled | **no warning, `status=768`** |
| elephc in `eval()` | warns, `status=0` |

**Fairness:** the write-through itself is pre-existing house behaviour, not pcntl's —
`call_user_func('sort', $a)` behaves the same way on this branch, which I verified. What belongs to
this PR is that the two backends disagree, and that `docs/php/pcntl.md:164-166` states that
"Dynamic `eval()` uses the same … wait outputs, and **warning behavior** as AOT code". As measured
that sentence is false in both halves — F4 and F5 each contradict it.

The eval-side warning also prints with no `Warning: ` prefix and no trailing newline, so it glues
onto the next line: my run captured `value givenstatus=0`. Every other warning string in the bridge
carries `"Warning: …\n"`.

## F6 — `pcntl_getpriority` / `pcntl_setpriority` emit no warning on OS failure

**Severity: important. Verified.**

```php
var_dump(pcntl_getpriority(999999));
```
```
php     Warning: pcntl_getpriority(): Error 3: No process was located using the given parameters
        bool(false)
elephc  bool(false)              (no warning, on stdout or stderr)
```

The errno is recorded correctly — `pcntl_get_last_error()` returns 3 on both — so only the warning
channel is missing. The bridge mirrors php's warning strings verbatim elsewhere (fork's four errno
messages are copied from `pcntl.c`), which is why this reads as an omission rather than a choice.

## F7 — `pcntl_exec()` refuses non-string array values at compile time

**Severity: important. Verified.** `src/builtins/system/pcntl_exec.rs:46-53` accepts only
`PhpType::Str | PhpType::Never` as the element type of `$args` and `$env`:

| program | php | elephc |
|---|---|---|
| `pcntl_exec('/bin/sh', ['-c', $cmd, 123])` | runs | compile error, `Mixed given` |
| `pcntl_exec('/bin/echo', [1, 2])` | prints `1 2` | compile error, `Int given` |
| `$port = 8080; … ['-c', $cmd, $port]` | runs | compile error, `Int given` |
| `… , ['K' => 123]` (env) | runs | compile error, `Int given` |
| all of the above with `'123'` | ok | ok ✔ |

php coerces elements with `zval_try_get_string`, so `['--port', $port]` with an integer `$port` is
idiomatic code that does not compile here. `Mixed` is in the rejected set too, so an `$args` array
assembled at runtime can fall foul of it even when every element really is a string. `exec.rs`'s
own doc says it copies an "already string-coerced" argument — the coercion is simply not happening
on the lowering side.

## F8 — `pcntl_signal_get_handler` returns a value that contradicts itself

**Severity: important. Verified.**

```php
pcntl_signal(SIGUSR1, 'namedHandler');
$h = pcntl_signal_get_handler(SIGUSR1);
```

| | php | elephc |
|---|---|---|
| `var_dump` / `gettype` | `string(12) "namedHandler"` / `string` | `NULL` / `NULL` |
| `$h === null` | false | **false** |
| `is_string` | 1 | 0 |
| `is_callable` | 1 | **1** |

The same holds for a closure and for `['Class','method']`. `SIG_IGN` → `int(1)`, `SIG_DFL` →
`int(0)` and never-set → `int(0)` are all correct.

The handler is **not** lost — it fires normally, which I verified in the same program. So this is
the introspection path returning a value whose type tag does not match its content. The damage is
that `var_dump`, `gettype`, `is_string` and `=== null` each answer differently from php, so the
usual inspect-then-restore idiom takes the wrong branch.

Read from the *other* backend it gets worse: an AOT-installed string handler read inside `eval()`
returns `int(0)`, indistinguishable from "no handler installed" and inviting an overwrite
(*relayed*). That makes three different answers for one installed handler.

## F9 — `pcntl_signal(SIGKILL, …)` returns `false` where php dies

**Severity: important (a decision to make explicit). Verified.**

```
php     Fatal error: Error installing signal handler for 9   (the process stops)
elephc  bool(false)  →  execution continues
```

Same for `SIGSTOP`, and `@` does not suppress php's fatal, so a program reaching this line behaves
completely differently. elephc's is the more civilised behaviour; whichever you keep, `docs/php/`
is the place to record it.

## F10 — `PRIO_DARWIN_THREAD` with a nonzero pid returns `false` instead of throwing

**Severity: important. Verified.** macOS-only.

```php
pcntl_getpriority(5, PRIO_DARWIN_THREAD);
```
```
php     ValueError: … Argument #1 ($process_id) must be 0 (zero) if PRIO_DARWIN_THREAD
                    is provided as second parameter
elephc  bool(false)
```

To be precise about scope: the *mode-range* `ValueError` is byte-exact. It is only php's second
EINVAL branch, the pid-must-be-zero rule, that is unimplemented.

---

## Smaller items — relayed, not independently re-run

- `pcntl_signal(SIGUSR1, SIG_IGN, [])` surfaces a raw `EIR backend error: unsupported EIR backend
  feature: pcntl_signal restart_syscalls for PHP type Array(Never)` with no `error[line:col]`,
  where php throws a catchable `TypeError` for `?bool`.
- `pcntl_setqos_class()` with no arguments succeeds on both backends; php 8.5 throws
  `ArgumentCountError` (its C impl is `ZEND_PARSE_PARAMETERS_START(1, 1)` despite the stub default).
- Eval `pcntl_sigprocmask(SIG_BLOCK, ["abc"])` throws `ValueError` where php throws `TypeError` —
  the wrong class breaks class-specific catches.
- `pcntl_signal(SIGALRM, true)` reports "bool given"; php 8.5 spells boolean literals in type
  errors, "true given".
- `extension_loaded('posix')` is false while `posix_setsid`/`posix_setpgid` ship and work.
- Linux `SIGRTMIN`/`SIGRTMAX` are hardcoded 34/64 (glibc); a musl-based php reports 35. Flagged as
  read-in-source, not asserted — relevant only if musl targets are in scope.
- `requirements.rs:29-66` lists 32 of 35 names, omitting `pcntl_daemon`, `posix_setpgid`,
  `posix_setsid`. Linking is unaffected; it only skews support metadata.

---

## What is solid

Fork and status decoding are exact across exit codes 0/1/42/255 and the 256/300/−1 wrap;
`WIFEXITED`/`WEXITSTATUS`/`WIFSIGNALED`/`WTERMSIG`/`WIFSTOPPED`; `waitpid` blocking, with
`WNOHANG`, and with `-1` over two children; `pcntl_wait`; the rusage out-parameter (17 keys, and
the macOS microsecond normalisation from `5a26c95e8` holds — `ru_utime.tv_usec` stays in
`[0, 1000000)`); errno and `pcntl_strerror` on the direct path.

Signal *delivery* is right everywhere I could reach it: external SIGTERM/SIGKILL/SIGINT to a forked
child produce the correct `WIFSIGNALED`/`WTERMSIG`; a child that installs a handler and exits from
it works for a closure, a replaced handler, and a named string handler — so the "retain string
signal handlers during pruning" fix in `4892a0037` does hold at runtime; `SIG_IGN` survives;
`pcntl_alarm` sets, cancels and fires; masks block and report correctly. A 500-signal storm against
a child with one deferred dispatch collapses identically to php and the child survives, which is
what the self-pipe-plus-overflow design exists to produce.

`pcntl_exec` matches on argv, env, exit code, missing binary, empty argv and numeric env *keys*,
and an embedded NUL in either the path or an argument produces php's message byte for byte.

The Fable juror additionally confirmed clean: **74/74 macOS constants exact** against
`ReflectionExtension('pcntl')`; `function_exists` target gating exact across nine functions;
`pcntl_waitid` runtime parity including `$info`; handler `$siginfo` parity; `async_signals(true)`
delivery; byte-exact `ValueError` texts across signal range, sigprocmask modes, exec NUL variants
and priority modes; named-argument and omitted-optional parity; and every one of the 35 builtins
carrying at least one test.

## The jury

| juror | material | outcome |
|---|---|---|
| GLM 5.3 | full bridge sources, no repo | no actionable finding |
| Kimi K3 | full bridge sources, no repo | no actionable finding |
| Fable (high) | the worktree, told to skip everything already measured | F1–F6, F10 and the relayed items |

GLM 5.3 and Kimi K3 were aimed at async-signal-safety, the fork race and the overflow queue. They
examined and dismissed, with reasons: the handler's errno save/restore; `copy_signal_siginfo`
performing only memory reads; `sigfillset` in `sa_mask`; partial writes (a 96-byte record is far
under the POSIX ≥512 `PIPE_BUF` guarantee, so a nonblocking write is all-or-`EAGAIN`); pipe
capacity before spill; the child restoring the parent's pre-fork mask; `pcntl_fork` with an
uninitialised queue; `waitpid` EINTR; `pid_t` widening; `getpriority`'s errno-clear-then-check;
`_SC_NPROCESSORS_CONF` versus online. Kimi flagged a pipe/overflow ordering inversion and then
correctly talked itself out of it — standard signals do not queue in php.

I report those two null results as results: for ~2950 lines of signal-handling code, two
independent passes finding nothing is information. Both also stopped at "cannot settle this from
the bridge alone" on `pcntl_exec`'s argument marshalling — which is where F7 came from once I
measured it.

**Refuted** by execution, each of which would have made a plausible-sounding false report: the
`{0,0}` timeout in `pcntl_sigtimedwait` matches php's `ValueError` rule exactly; the Fiber
rejection is both correct and reachable — php itself throws `FiberError` in all three shapes
tested and elephc matches the message; `pcntl_waitid` is *not* Linux-only (php has it on macOS), so
its `All` gating is right; and the ship-list bug class this repo has hit before does **not** recur
— `libelephc_pcntl.a` is in every list and the linker guard tests are derived from the bridges
table rather than pinned.

## Verification and caveats

- 39 differential programs plus targeted probes, php 8.5.6 oracle, macOS ARM64.
- `cargo test -p elephc-pcntl`: 25 passed. `cargo build`: warning-free.
- CI green on `aae00ac84` (129 pass, 1 skipping); MERGEABLE / CLEAN.
- **Not this PR:** `sort()`/`rsort()`/`usort()` on an array received through a by-reference
  parameter segfaults. It reproduces with a plain user function and no pcntl, identically on the
  merge base, so it is filed separately as #819. It matters here only because
  `pcntl_sigprocmask($how, $set, $old)` is one of the most natural ways to walk into it.
- **Platform caveat:** everything above is macOS ARM64. The Linux family — `waitid` on Linux,
  `unshare`, `setns`, `getcpu`, CPU affinity, `qos_class`, `daemon` — is untested by all three
  passes, and it is a large part of `2963f8dee`.
- elephc has no `posix_kill`/`getmypid`/`posix_getppid`, so a process cannot signal itself from
  PHP and every signal test has to arrange delivery from another process. Not this PR's job, but
  worth knowing if a `posix` PR is on the roadmap.

nahime0:
@Guikingone can you please re-review this?

Guikingone (commented):
## Re-review — PCNTL extension (`7f426fe91`)

I rebuilt the branch at `7f426fe91` and re-ran the ten findings from the first pass plus fifty new
programs against php 8.5.6 (`php -n`, macOS ARM64) — **60 differential programs in all**.
Everything below I executed myself in this pass; there is no relayed material this time.

**Nine of the ten findings are fixed, and the one that mattered most is gone: no silent value
corruption remains.** Every defect still standing fails loudly — a compile error or a fatal — which
is the right failure mode for what is left. The core process-control paths are unchanged and clean.

---

## Fixed — re-measured, byte-identical to php

| | was | now |
|---|---|---|
| **F1** untyped closure collapses a pcntl result to an int | `int(0)` for `string(16) "Invalid argument"` | correct |
| **F2** runtime-valued bad handler name → uncatchable fatal | `Fatal error: Call to undefined function <dynamic>()`, exit 1 | byte-exact `TypeError`, process survives |
| **F3** `function_exists()` / `PHP_OS` guards do not gate compilation | hard compile error inside a provably dead branch | folds; `PHP_OS_FAMILY` now defined |
| **F4** signal-set arrays rejected at compile time | `error[…]: must be an indexed integer array` | literals accepted, keys ignored, numeric strings coerced |
| **F6** priority failures emit no warning | silent `bool(false)` | both warnings, errno text exact |
| **F7** `pcntl_exec()` refuses non-string array values | compile error on `[1, 2]` | scalars coerced |
| **F8** `pcntl_signal_get_handler` contradicts itself | `NULL` for `string(12) "namedHandler"` | correct for strings **and** arrays |
| **F9** `pcntl_signal(SIGKILL, …)` returns `false` where php dies | `bool(false)`, execution continues | fatal, exit 255, `@` does not suppress it |
| **F10** `PRIO_DARWIN_THREAD` with a nonzero pid | `bool(false)` | `ValueError` (see R5 for `pcntl_setpriority`'s wording) |

F1 is worth a note because it was the blocker. All four discriminators now pass — untyped closure,
`fn() =>` arrow form, `pcntl_get_last_error`, `pcntl_wifexited` — and the fix is the right one:
`builtin_call_types` is threaded into `closure_signature_from_ast`, so the inferred return path
finally knows what the builtin returns instead of defaulting to an integer.

F8's practical payoff also checks out end to end. I installed a named handler, read it back,
restored the disposition from the value I had read, re-installed the read-back string, and let
`pcntl_alarm` fire: `int(0)` → `string(1) "h"` → `int(0)` → `string(1) "h"` → `h14`, identical to
php at every step. Reading an AOT registration from inside `eval()` now returns
`string(2) "nh"` where it used to return `int(0)`.

Smaller items from the first pass, all re-measured and now correct: the `restart_syscalls`
`TypeError` for `?bool`, `"true given"` for boolean literals, `extension_loaded('posix')`, the
`TypeError`-vs-`ValueError` class in eval for `["abc"]`, the three missing names in
`requirements.rs`, and the musl `SIGRTMIN` question answered by scoping the target matrix. Splitting
the `call_user_func` by-reference degradation out to #820 and narrowing the doc claim to
"direct-call warning behavior" is the honest resolution of old F5 — I re-verified that direct-call
warning parity now holds, including inside `eval()`.

Core paths re-checked and clean: `fork` / `waitpid` / `WIFEXITED` / `WEXITSTATUS` / `WIFSIGNALED`,
synchronous dispatch of a named handler under `pcntl_alarm`, and asynchronous dispatch
(`pcntl_async_signals(true)`) of a closure capturing by reference.

---

## Still open

### R1 — `pcntl_exec()` does not call `__toString()`

**Severity: important. Verified.** The scalar coercion added in this round stops at objects.

```php
class S { public function __toString(): string { return "obj"; } }
$o = new S();
echo "as-string: $o\n";          // both print "as-string: obj"
pcntl_exec('/bin/echo', ['a', $o]);
```
```
php     a obj
elephc  Fatal error: Uncaught TypeError: pcntl_exec(): array value could not be converted to string   (exit 255)
```

The same program proves the object *is* convertible one line earlier, so this is `pcntl_exec`'s
coercion and not the object. php uses `zval_try_get_string`, which honours `__toString`.

The non-Stringable case diverges too, in class as well as text:

| | php | elephc |
|---|---|---|
| `pcntl_exec('/bin/echo', ['a', new P()])` | `Error: Object of class P could not be converted to string` | `TypeError: pcntl_exec(): array value could not be converted to string` |

`catch (TypeError $e)` catches elephc's and not php's.

### R2 — the signal-set fix is literal-only

**Severity: important. Verified.** `fold_pcntl_signal_array_literal` rewrites an array **literal**
in argument position, so the same array in a variable never reaches it:

```php
$assoc = ["k" => SIGTERM];
pcntl_sigprocmask(SIG_UNBLOCK, $assoc);
```
```
php     bool(true)
elephc  error[3:41]: pcntl_sigprocmask() parameter $signals must be an indexed integer array,
                     AssocArray { key: Str, value: Int } given
```

`$sigs = [9]` in a variable is fine; it is the string-valued and associative shapes that only work
when written inline. The same source inside `eval()` returns `bool(true)` and matches php — so the
AOT/eval split from old F4 survives in this narrower form.

Leading-numeric strings are the other half. php converts and warns; the fold's `parse::<f64>()`
gives up and the checker then refuses the program:

```php
pcntl_sigprocmask(SIG_BLOCK, ["9abc"]);
```
```
php     Warning: A non-numeric value encountered
        bool(true)                                  ← php used 9
elephc  error[2:45]: … must be an indexed integer array, Array(Str) given
```

`[" 9 "]` and `["1e1"]` are handled correctly.

### R3 — `PHP_OS_FAMILY` is undefined inside `eval()`

**Severity: important. Verified.** This one is new: the constant was added to the AOT prescan and
the checker, but not to the eval constant surface.

```php
eval('echo PHP_OS_FAMILY, "\n";');
echo "after\n";
```
```
php     Darwin
        after
elephc  Fatal error: eval() runtime failed        (exit 1, "after" never prints)
```

```php
eval('var_dump(defined("PHP_OS_FAMILY"));');    // php: bool(true)   elephc: bool(false)
```

`PHP_OS` works in eval; only the constant this PR introduces is missing. The failure is also
opaque — no `Undefined constant "PHP_OS_FAMILY"`, no line, just `eval() runtime failed`.

### R4 — the polyfill idiom still does not compile

**Severity: important. Verified.** F3's fix covers one of the two portable-source shapes. The
skip-the-call shape now works exactly as documented. The fallback shape does not:

```php
if (!function_exists('pcntl_getcpu')) {
    function pcntl_getcpu() { return 7; }
}
echo pcntl_getcpu(), "\n";
```
```
php     7
elephc  error[3:5]: Cannot redeclare built-in function: pcntl_getcpu
        error[5:6]: pcntl_getcpu() is not available for the macos-aarch64 target
```

The guard folds correctly and the branch is kept — that part works. What fails is that the
redeclare check consults registry membership rather than target availability, so a name that does
not exist on this target still blocks the user's declaration, and the call site then binds to the
absent builtin anyway. Same for `pcntl_setns`.

**Attribution:** this is not solely yours. The identical shape fails for
`imagecreatetruecolor` without `--with-image` (differently: an `EIR backend error` with no
`error[line:col]`), so the registry-vs-availability conflict is house-wide and merely newly
reachable here. `function strlen() {}` is correctly refused on both sides, so the rule itself is
right for available builtins. I raise it because `docs/php/pcntl.md:163-164` now says target guards
are folded "so one portable source can isolate those calls", and the most common portable idiom in
PHP is the one that still cannot compile.

### R5 — `pcntl_setpriority`'s `ValueError` says "third parameter"

**Severity: nit. Verified.** php-src hardcodes the same sentence for both functions, ordinal
included, even though `$mode` really is `pcntl_setpriority`'s third argument:

```
php     … must be 0 (zero) if PRIO_DARWIN_THREAD is provided as second parameter
elephc  … must be 0 (zero) if PRIO_DARWIN_THREAD is provided as third parameter
```

`pcntl_getpriority`'s message is byte-exact. Computing the ordinal is the more correct English and
the wrong string.

### R6 — `["9.7"]` loses php's precision deprecation

**Severity: nit. Verified.** Both end up blocking signal 9; php also emits
`Deprecated: Implicit conversion from float-string "9.7" to int loses precision`, elephc emits
nothing.

---

## Not this PR — four leads I chased and dropped

Recording these so nobody re-opens them from this branch:

- **Conditional function declarations are hoisted and early-bound.** `echo bar(); if (…) { function
  bar() {…} }` prints a value where php fatals with `Call to undefined function bar()`. I suspected
  the new `prune_target_boolean_if`, but `if (true) { … }` — a condition the new pruning never
  touches — behaves identically, so this is house-wide.
- **Warnings carry no `in FILE on line N` suffix.** `Undefined array key "x"` is formatted the same
  way, so the pcntl warnings match house style rather than diverging from it.
- **Zero-argument builtin calls are compile errors, not catchable `ArgumentCountError`s.**
  `strlen()` behaves the same, so `pcntl_setqos_class()` is now consistent with the house rule; my
  earlier note about it succeeding is resolved.
- **`get_class($closure)` returns `string(0) ""` and `$closure instanceof Closure` is `false`.** I
  found this through `pcntl_signal_get_handler()` returning a closure and nearly filed it, but it
  reproduces on a plain `$c = function () {};` with no pcntl in the program. Worth its own issue;
  the consequence for this PR is only that `docs/php/pcntl.md:125-126`'s "closures remain `Closure`
  objects" is not currently observable — the value's identity is right (`$g === $c` is `true`), its
  class is not.

---

## Branch state

- CI was green on `7f426fe91` — 129 of 130 checks pass and one is skipped, including all 16 codegen shards and the full eval matrix.
- The branch is **22 commits behind `main`** and GitHub reports it conflicting. The conflict is
  confined to two files, `src/cli.rs` and `docs/compiling/cli-reference.md` — the `--with-…` flag
  table on both sides. Everything else auto-merges, including `src/ir/mod.rs`, `src/types/mod.rs`
  and `src/linker/bridges.rs`.
- While the PR reads conflicting, CI will not run again, so the green above is the last word until
  it is merged up.

## Verification and caveats

macOS ARM64 only; the Linux-only surface (`pcntl_setns`, `pcntl_unshare`, CPU affinity, `getcpu`)
was exercised only through its unavailability path. Everything measured with `php -n`, because this
machine's `php.ini` loads Xdebug and it rewrites `var_dump` output — every `var_dump` case is a
false positive without it. The optimizer changes in this round (`function_exists` / `PHP_OS` folding
and the pre-check branch pruning) affect every program, not only pcntl sources, so I probed them
separately: polyfill shapes for available builtins, namespace-prefixed and case-varied names,
user-function shadowing, undefined-variable behaviour after a branch is pruned, and eval parity.
Apart from R3 and R4 they behave.

nahime0:
@Guikingone Thanks again for the thorough re-review. All of the remaining in-scope findings have now been addressed on `a8ed1e961`:

- **R1 — `pcntl_exec()` object coercion:** AOT and eval now both use PHP string-context conversion for every argv/env value. `Stringable` objects invoke `__toString()` and preserve the resulting bytes; plain objects raise `Error` instead of being silently converted or reported as the wrong exception class. The eval path now goes through `eval_string_context_value()` before `string_bytes()`. Regression coverage includes Stringable argv and env values plus the non-Stringable case.
- **R2 — signal-set coercion:** signal arrays are normalized beyond inline literals, including associative shapes with ignored keys. PHP-style numeric prefixes are handled consistently: `"9abc"` emits `Warning: A non-numeric value encountered` and contributes signal 9. The same behavior is covered in eval.
- **R3 — `PHP_OS_FAMILY` in eval:** the constant is exposed through the eval constant surface as well as AOT.
- **R4 — guarded polyfills:** unavailable target-specific PCNTL names can be supplied by the canonical `if (!function_exists(...))` polyfill shape. I also fixed the house-wide counterpart you identified: guarded `imagecreatetruecolor` polyfills compile without `--with-image`, while genuinely available builtins remain protected from redeclaration.
- **R5 — Darwin priority wording:** the `pcntl_setpriority` `ValueError` now deliberately matches php-src's `second parameter` wording byte-for-byte.
- **R6 — float-string precision deprecation:** `"9.7"` still contributes signal 9 and now emits PHP's precision-loss deprecation in both AOT and eval.

The older `call_user_func` by-reference behavior remains explicitly out of scope here and is tracked separately in #820.

While bringing the branch up to the current `main`, I also resolved the merge conflicts and two CI-only regressions found by the fresh matrix:

- bridge extension metadata now supports one bridge exposing multiple PHP extensions, so `elephc_pcntl` reports both `pcntl` and `posix` in production and codegen fixtures;
- the x86_64 Fiber signal guard now aligns `rsp` around its throwing call, with a focused SysV audit regression test and explicit audit-hole documentation for the intentional eval-fatal realignment and Fiber stack hand-off.

Current state: **144 checks successful, 0 failed, 1 expected benchmark skip**; GitHub reports the PR **MERGEABLE / CLEAN**, and the branch is 0 commits behind `main`.

Could you please take a fresh review pass against `a8ed1e961` when you have time, especially over the eval `pcntl_exec` coercion, signal numeric warning/deprecation parity, and guarded-polyfill behavior?

Guikingone (commented):
## Third pass — PCNTL extension (`a8ed1e961`)

**All six findings from the last round are fixed.** I re-measured every one against `php -n`
8.5.6 on macOS ARM64 and they are byte-identical, including the parts that were hardest to get
right: `pcntl_exec()` now honours `__toString()` and raises php's own `Error` (not a `TypeError`)
for a plain object, in both backends; a NUL byte returned from `__toString()` still raises php's
positional `ValueError`; a throwing `__toString()` propagates unchanged; signal sets held in
variables work in every shape including associative; `"9abc"` warns and contributes signal 9;
`"12.7"` emits the float-string deprecation and contributes 12; `PHP_OS_FAMILY` resolves inside
`eval()`; and `pcntl_setpriority`'s `ValueError` reproduces php-src's copy-pasted "second
parameter" wording rather than the ordinal that would be better English.

Two decisions worth naming, because they do not show up in a diff summary. First,
`pcntl_sigprocmask(SIG_BLOCK, ["abc"])` used to be a compile error and now throws php's exact
`TypeError` at run time — the fix moved the family from compile-time refusal to runtime parity
instead of just widening the accepted set, which is the harder and better choice. Second, the
target-aware redeclaration rule in `src/types/checker/driver/functions.rs` has the right shape:
gate the redeclare error on availability rather than on registry membership, so a polyfill is
legal exactly where php would allow one. No regressions anywhere in the earlier corpus, and
fork/waitpid/status plus synchronous and asynchronous dispatch are all still byte-clean.

Three things to fix. None is a blocker, and the jury changed my mind about the first one's cause.

### Method

About 90 differential programs against `php -n` on this head, 111 in the corpus across the three
passes, plus an adversarial jury of three agents from two model families briefed to refute the
findings rather than grade them. Every claim is graded: **✅ proven** means I ran it and the bytes
are in the finding. The jury confirmed finding 2 unanimously, and split on findings 1 and 3 — in
both cases against my stated *mechanism*, and in both cases they were right, which is recorded
below. I re-verified every jury claim by execution before adopting it; two of their four additions
did not survive that. `php -n` is mandatory when re-running any of this: the machine's `php.ini`
loads Xdebug and it rewrites `var_dump` output.

---

## Should-fix

### 1 — `pcntl_exec()` answers one input class three different ways ✅ proven

An array where a string is expected produces three different behaviours depending on the shape,
none of them php's:

| shape | php | elephc |
|---|---|---|
| `['x', [1,2], 'z']` — heterogeneous argv | warning, argv gets `Array` | **silent empty argument** |
| `['A'=>'x', 'B'=>[1,2]]` — heterogeneous env | warning, env gets `Array` | uncaught `TypeError: pcntl_exec(): array value could not be converted to string`, exit 255 |
| `[[1],[2]]` or `['K'=>[1,2]]` — homogeneous | warning, gets `Array` | `error[2:25]: pcntl_exec() array values must be coercible to string, Array(Int) given` |

With the argument boundaries made visible:

```php
pcntl_exec('/bin/sh', ['-c', 'for a in "$@"; do printf "[%s]" "$a"; done; printf "\n"',
                       'sh', 'x', [1, 2], 'z']);
```
```
php     Warning: Array to string conversion
        [x][Array][z]
elephc  [x][][z]
```

It also holds for the shape real code produces, an argv built element by element (`$args[] = …`),
and `eval()` behaves the same as AOT.

**I originally filed the silent row as yours; the jury refuted that and I confirmed the
refutation.** The empty string comes from the house-wide boxed-`Mixed` string conversion, not from
anything this PR wrote — see the separate section below, where `echo`, `.` and `implode()`
reproduce it with no pcntl in the program. What is yours is the *inconsistency*: rows 2 and 3 are
this PR's own code — the tag ladder in `emit_exec_assoc_mixed_string`
(`src/codegen/lower_inst/builtins/pcntl_exec.rs`) knows tags 0/1/2/3/6/8 and throws for everything
else, and the `matches!` in `src/builtins/system/pcntl_exec.rs` accepts `Mixed` and `Union` while
rejecting `Array`, which is what splits homogeneous from heterogeneous.

There is a fourth split in the same family, and it is silent too: a **resource** in a
heterogeneous env array.

```php
pcntl_exec('/bin/sh', ['-c', 'env | grep "^[RS]=" | sort', 'sh'], ['R' => STDIN, 'S' => 'x']);
```
```
php     R=Resource id #1        elephc  R=0
        S=x                            S=x
```

**Fix:** route every element through one conversion, and make that conversion php's — `"Array"`
plus `Warning: Array to string conversion` for an array, `"Resource id #N"` for a resource. The
object path added this round already has exactly that shape. Fixing the shared helper (next
section) fixes the silent rows here for free; the ladder and the `matches!` still need to agree
with each other afterwards.

**Why the tests did not catch it:** `tests/codegen/pcntl.rs:702` is titled "Uses `__toString()`
for object values supplied through heterogeneous exec arrays" — the heterogeneous case was
considered, but only with an object in it. There is no case with an array or a resource element,
in `$args` or in `$env_vars`.

### 2 — a guarded polyfill inside a `namespace` does not compile ✅ proven

```php
<?php
namespace App;

if (!function_exists('pcntl_getcpu')) {
    function pcntl_getcpu() { return 7; }
}
echo pcntl_getcpu(), "\n";
```
```
php     7
elephc  error[7:6]: pcntl_getcpu() is not available for the macos-aarch64 target
```

In php the unqualified string argument to `function_exists()` names the *global* symbol, which
does not exist on macOS, so the guard is true, `App\pcntl_getcpu` is declared, and the unqualified
call resolves to it before falling back to the global table.

Three discriminators isolate it tightly:

| variant | php | elephc |
|---|---|---|
| the same declaration, **unguarded**, in `namespace App` | `7` | `7` ✔ |
| a plain namespaced user function `App\helper()` | `7` | `7` ✔ |
| the guarded polyfill with **no** `namespace` line | `7` | `7` ✔ |
| **guarded + namespaced** | `7` | compile error ✗ |

Both jurors traced the same mechanism independently, and it is worth repeating because the fix
follows from it: `name_resolver/symbols.rs::collect_symbols` descends into `NamespaceBlock` but
never into `If` bodies, so the guarded `App\pcntl_getcpu` is missing from the symbol table;
`resolve_function_name` then falls through to `canonical_builtin_function_name`, which is registry
membership with no target filter, and binds the call to the **global** `pcntl_getcpu`. Meanwhile
the declaration rewriter *does* recurse into `if`, so the declaration lands in `fn_decls` as
`App\pcntl_getcpu`. At the call site the new relaxation in `types/checker/builtins/mod.rs` asks
`has_function_decl_folded("pcntl_getcpu")`, the keys differ (`app\pcntl_getcpu` vs
`pcntl_getcpu`), the relaxation is skipped, and the availability error fires at 7:6.

This matters because `docs/php/pcntl.md:167-169` now advertises the idiom directly: "A source may
also declare a fallback function with a target-unavailable builtin name inside the negative
`function_exists()` branch." Namespaced files are where that pattern actually lives.

### 3 — the guarded image polyfill reaches the backend as a duplicate declaration ✅ proven

```php
if (!function_exists('imagecreatetruecolor')) {
    function imagecreatetruecolor($w, $h) { return "poly"; }
}
$im = imagecreatetruecolor(2, 2);
var_dump(is_string($im));
var_dump(imagesx($im));          // any second, unguarded gd call
```
```
php     bool(false)
        int(2)
elephc  EIR backend error: unsupported EIR backend feature:
        language construct elephc_img_create_truecolor
```

A positionless internal error on a source php compiles and runs. **I first reported this as
gated on `--with-image`; a juror predicted it would reproduce without the flag and I confirmed
that by execution** — the message differs from the one the flag produces (`runtime_call from PHP
type Str to PHP type Object("GdImage")`) but the shape is the same, and the no-flag form is the
realistic one: polyfill the function you are missing, call the rest of gd normally.

My original mechanism was wrong and the jury caught it: `imagecreatetruecolor` is not a registry
builtin at all, it is a prelude PHP function, so the relaxation in `types/checker/builtins/mod.rs`
never fires here. The real path is `image_prelude/detect.rs::program_uses_image`, which treats an
image function referenced only under a canonical `!function_exists` guard as user-owned and does
not enable the bridge. When a *second* gd call is not user-owned, the full prelude is injected,
`activate_guarded_image_polyfills` is skipped, and the prelude's declaration and the user's
guarded one both reach the backend under one name.

Separately, and worth a deliberate decision rather than a fix: when the polyfill is the *only* gd
reference, elephc silently prefers it over the real implementation.

```
php                bool(false)   ← real gd
elephc (no flag)   bool(true)    ← the polyfill
```

That is defensible under a demand-linked capability model, but it diverges from php-with-gd, it is
not written down anywhere under `docs/`, and it is observably inconsistent with the *unguarded*
call in the same compiler, which enables the capability and uses the real function. Whichever way
you settle it, `docs/php/` is the place to say so.

---

## Not this PR — but it is what makes finding 1 silent

**A boxed `Mixed` holding an array stringifies to `""`, and one holding a resource to `"0"`, with
no warning.** ✅ proven, and reproduced three ways with no pcntl anywhere:

```php
function f(mixed $x) { echo $x, "\n"; } f([1,2,3]);
                        // php: warning + "Array"      elephc: nothing at all
function g(mixed $x): string { return "[" . $x . "]"; } var_dump(g([1,2,3]));
                        // php: warning + "[Array]"    elephc: string(2) "[]"
var_dump(implode(",", [1, [2, 3], "z"]));
                        // php: warning + "1,Array,z"  elephc: string(4) "1,,z"
var_dump(implode(",", [STDIN, 'x']));
                        // php: "Resource id #1,x"     elephc: string(3) "0,x"
```

`(string) STDIN` on its own is correct, so it is specifically the boxed-`Mixed` path. This is the
shared helper behind `echo`, concatenation, `implode()` and now `pcntl_exec()`'s argv, and it is
the reason finding 1's first row loses data quietly rather than loudly. It deserves its own issue;
fixing it there removes the silent half of finding 1 and a good deal more.

## Not this PR — six other leads I chased and dropped

Each killed by a discriminator, not by opinion. Two are real defects that want their own issue.

- **`\0` and `\x00` escapes are not interpreted inside `eval()`.** This nearly landed on you as a
  `pcntl_exec()` finding, because in eval an argument whose `__toString()` returns `"a\0b"` is
  handed to `exec()` where php raises `ValueError: … must not contain null bytes`. The guard is
  fine — the NUL never exists:

  ```php
  var_dump("a\0b");                    // AOT   php: string(3)  elephc: string(3) ✔
  eval('var_dump("a\0b");');           //       php: string(3)  elephc: string(4) "a\0b"
  eval('var_dump(strlen("a\0b"));');   //       php: int(3)     elephc: int(4)
  eval('var_dump("a\x00b");');         //       php: string(3)  elephc: string(6)
  ```

  `\t`, `\n` and `\\` are correct in eval, so it is the octal/hex forms specifically. The
  reproduction has no pcntl in it and this PR does not touch the Magician parser — but it does mean
  php's NUL-byte refusal for `pcntl_exec()` can never fire under `eval()`, which is worth knowing
  given what that check is for. ✅ proven.
- **`Warning: Undefined variable` is not emitted inside `eval()`.** `eval('echo "[$Q]\n";')` warns
  in php, prints nothing in elephc; the value is `[]` on both sides. No pcntl involved. ✅ proven.
- **A real float element loses php's precision deprecation.** `pcntl_sigprocmask(SIG_BLOCK, [9.7])`
  omits `Deprecated: Implicit conversion from float 9.7 to int loses precision`. Not yours:
  `str_repeat("a", 2.7)` omits it too. Worth knowing anyway, because this round *did* implement the
  float-**string** variant, so the pair is now inconsistent: `"9.7"` warns and `9.7` does not.
- **A polyfill called before its own guard.** `echo pcntl_getcpu(); if (!function_exists(…)) {
  function pcntl_getcpu() … }` prints `7` where php fatals. Not yours: `if (true) { function bar()
  {…} }` behaves identically, and the new pruning never touches a non-target condition.
- **Extra arguments to a zero-parameter polyfill are refused at compile time.** Not yours: a plain
  `function myzero() {} myzero(1, 2)` is refused the same way.
- **`pcntl_signal_get_handler()` returns a closure that fails `instanceof Closure`.** Reproduces on
  a plain `$c = function () {};` with no pcntl in the program. The only consequence for this PR is
  that `docs/php/pcntl.md:126-127`'s "closures remain `Closure` objects" is not observable — the
  identity is right (`$g === $c` is `true`), the class is not.

## Nit

- **An object in a signal set is still a compile error.** `$a = [new O()];
  pcntl_sigprocmask(SIG_BLOCK, $a)` gives php a catchable `TypeError: … signals must be of type
  int, O given`; elephc gives `error[4:45]: … must be an array of integer-coercible scalar values,
  Array(Object("O")) given`. It is the last member of the family that did not move to runtime
  parity with `["abc"]` this round, so `catch (TypeError $e)` still cannot run for it. ✅ proven.

---

## Mergeability

- CI on `a8ed1e961`: **144 pass, 1 skipped, 0 failed** — I confirm your count.
- The branch is now **24 commits behind `origin/main`** and GitHub reads `CONFLICTING / DIRTY`
  again. Your "MERGEABLE / CLEAN, 0 behind" was accurate when you wrote it; `main` has since taken
  the 0.26.6 release plus #879. Conflicts are `README.md` (stats) and **`src/codegen/frame.rs`**,
  where `feat(cdylib): add recoverable owned string ABI` on `main` meets this branch's
  signal-handler frame work — that one deserves a human, not a merge driver.
- Worth a look while merging up: #879 (`control-flow normalization v2 for loop, switch, and
  function shells`) and the two `fix(optimize)` commits behind it land in `src/optimize/`, which is
  where this PR's pre-check `function_exists`/`PHP_OS` folding and target-branch pruning also live.
  They do not conflict textually. They interact semantically.

## Verification and caveats

macOS ARM64 only. `pcntl_sigwaitinfo` and `pcntl_sigtimedwait` are outside the oracle's reach —
php does not define them on Darwin — so the literal-fold path for those two names is untested
against php by me, as is the whole Linux-only surface beyond its unavailability path.
`a8ed1e961`'s x86_64 frame-alignment fix is on a path this host cannot execute; I am relying on
the CI matrix for it, as you are. Two jury additions I did **not** adopt: a predicted `TypeError`
for the resource-in-env case (the real behaviour is `R=0`, measured above) and a predicted
`runtime_call … GdImage` message for the no-flag image case (the real message names
`elephc_img_create_truecolor`). Both pointed at something real; neither was right about what
came out.

nahime0:
@Guikingone Thanks again for the thorough third pass. All findings from that review are now addressed on `65a5496fc`:

- `pcntl_exec()` now routes argv and env values through consistent PHP string-context conversion in both AOT and eval. Nested arrays warn and become `Array`, resources retain their `Resource id #N` form, and the object path remains aligned for Stringable and non-Stringable objects.
- Guarded namespaced polyfills are collected and resolved correctly.
- Guarded image polyfills remain user-owned when another image call enables the bridge. The demand-linked behavior for a polyfill-only source is documented.
- Signal-set object elements now reach the runtime coercion path and raise the PHP-compatible catchable `TypeError`.
- The shared issues exposed by the review were fixed as well: boxed Mixed array/resource stringification, eval octal and hex NUL escapes, eval undefined-variable warnings, float precision deprecation in signal coercion, and `Closure` instanceof behavior.
- Follow-up CI regressions were fixed for defined-null and quiet eval scope reads, the stale nested-array rejection test was corrected, and the slow extension bridge probe was split so each macOS test stays below the per-test timeout.

The branch was rebased onto the current `origin/main` at `24cf6e8e7`. Conflict resolution preserves both the new monitoring contract integration and the PCNTL/POSIX extension surface. It is currently 0 commits behind, GitHub reports `MERGEABLE / CLEAN`, and the post-rebase CI is fully green: 144 passed, 1 skipped, 0 failed.

Could you please take another look?

Guikingone (commented):
## Fourth pass — PCNTL extension (`65a5496fc`)

Everything I raised last round is fixed, and so are the five house-wide defects I had explicitly
attributed *away* from this PR — you went and fixed them anyway, which is why `pcntl_exec` now
agrees with php instead of agreeing with elephc. Verified on this head, byte-identical to `php -n`:

- **`pcntl_exec` array elements** yield `Array` plus `Warning: Array to string conversion`, in argv
  and env, inline / in a variable / appended, homogeneous and heterogeneous, AOT and `eval()`. Over
  20 000 iterations the warning count matches php exactly (20 000 both sides) and `--gc-stats`
  reports `allocs=180001 frees=180001` — the new conversion leaks nothing.
- **Boxed-`Mixed` array stringification** fixed at the shared helper: `echo`, `.` and `implode()`.
- **`eval()`**: `\0` and `\x00` escapes interpreted, `Warning: Undefined variable` fires, and
  `isset` / `empty` / `??` over undefined, defined-null and set variables are byte-identical.
- **Objects in a signal set** raise php's exact catchable `TypeError`; the **guarded image
  polyfill** plus another gd call compiles and uses real gd; `get_class($closure)` is `"Closure"`
  and `instanceof Closure` is `true`; the **namespaced guarded polyfill** of a *target-unavailable*
  builtin compiles.
- 53 previously-passing programs re-run with no regression.

The polyfill-only image case is now written down at `docs/php/image.md:17-21`, which is the right
way to close a design decision.

**But the round also introduced a blocker and four smaller regressions, all in the new
name-resolution and float-coercion paths.** I built `main` (`0b1d17038`) to separate them from what
was already broken; every "before/after" below is that comparison, not an inference.

### Method

About 100 differential programs against `php -n` 8.5.6 on macOS ARM64 across two compilers — this
head and `main` — out of a corpus of 134 over four passes, plus `--gc-stats` for ownership. An
adversarial jury of three agents from two model families was briefed to refute rather than grade;
it took my headline finding apart, and four of its six additions survived my own re-execution. **✅
proven** means I ran it and the output is in the finding.

---

## Blocker

### 1 — a namespaced polyfill of an *available* builtin no longer compiles ✅ proven

```php
<?php
namespace App;

if (!function_exists('str_contains')) {
    function str_contains(string $haystack, string $needle): bool { return false; }
}
var_dump(str_contains("abc", "b"));
```
```
php               bool(true)
elephc @ main     bool(true)                                              ✔
elephc @ this PR  error[7:10]: Undefined function: App\str_contains       ✗
```

This is the most common polyfill idiom in the ecosystem — `str_contains`, `str_starts_with`,
`array_is_list` shims sit in namespaced files all over Composer — and it compiled before this
branch.

The mechanism, which two jurors traced independently: name resolution runs before
`fold_constants_for_target`, so with `collect_symbols` now recursing into `if` bodies the resolver
binds the unqualified call to `App\str_contains`; the target fold then folds
`function_exists('str_contains')` to `true` and prunes the branch, deleting the declaration; and
`registry::lookup("App\\str_contains")` does not strip the namespace, so the checker finds neither a
user function nor a builtin.

**Fix:** either strip the namespace when falling back to the builtin registry, or make the symbol
collected from a conditional branch aware that the branch may be pruned. Note the asymmetry that
makes this pass CI: the *unavailable* builtin case (`pcntl_getcpu`) works precisely because the
guard folds the other way and the declaration survives.

---

## Should-fix

### 2 — the PR widens a compiler panic to class names that used to compile ✅ proven

```php
$legacy = PHP_VERSION_ID < 80100;
if ($legacy) {
    class DateTimeZone { public static function listIdentifiers() { return []; } }
}
var_dump(in_array('UTC', DateTimeZone::listIdentifiers()));
```
```
php               bool(true)
elephc @ main     bool(true)                                              ✔
elephc @ this PR  thread 'main' panicked at src/ir_lower/program/class_methods.rs:219:13:
                  checker method declaration DateTimeZone::listIdentifiers has no matching AST method
```

Same shape with `ArrayObject` panics on **both** compilers, so the panic class itself is older than
this branch — what the `collect_symbols` recursion changes is which names reach it. An unconditional
`class DateTimeZone { … }` at top level compiles on this head, so it is specifically the conditional
form. A panic with a backtrace note is the wrong diagnostic for any user program; the pre-existing
half deserves its own issue.

### 3 — `intval($f, 8)` now emits a deprecation php never emits ✅ proven

```php
$f = 42.9;
if (count([1, 2]) > 9) { $f = 1.0; }
var_dump(intval($f, 8));
```
```
php               int(42)
elephc @ main     int(42)                                                                    ✔
elephc @ this PR  Deprecated: Implicit conversion from float 42.9 to int loses precision
                  int(42)                                                                    ✗
```

`intval()` is an *explicit* conversion, so php stays silent. Two-argument `intval` routes a
non-string subject through `load_as_int`, whose float arm is now the deprecating helper.

### 4 — the deprecation prints the wrong float ✅ proven

```php
$g = 0.1 + 0.2;              var_dump(chr($g));
$m = 123456789.123456789;    var_dump(substr("abcdef", $m));
```
```
php     Deprecated: Implicit conversion from float 0.30000000000000004 to int loses precision
        Deprecated: Implicit conversion from float 123456789.12345679 to int loses precision
elephc  Deprecated: Implicit conversion from float 0.3 to int loses precision
        Deprecated: Implicit conversion from float 123456789.12346 to int loses precision
```

The message exists to name the precision that was lost, and it is rendered with `__rt_ftoa`'s
`%.14G` — which rounds the loss away. php uses shortest-round-trip; the house already has that
formatter with the same ABI (`__rt_ftoa_repr`).

### 5 — the new coercion has no "not representable" branch ✅ proven

php has two diagnostics in this family and the new path only ever emits one. In this PR's own
signal-set coercion:

```
pcntl_sigprocmask(SIG_BLOCK, [INF])
  php     Warning: The float INF is not representable as an int, cast occurred
          ValueError: … signals must be between 1 and 31
  elephc  Deprecated: Implicit conversion from float INF to int loses precision
          ValueError: … signals must be between 1 and 31
```

`NAN` makes php print *both* the warning and the deprecation; elephc prints only the deprecation.
The values and the `ValueError` are right; the diagnostics are not.

Where php treats it as a type error instead, the control flow diverges too:

```php
var_dump(substr("abcdef", 1.0E+20));
```
```
php     Fatal error: Uncaught TypeError: substr(): Argument #2 ($offset) must be of type int,
        float given                                                          (exit 255)
elephc  Deprecated: Implicit conversion from float 1.0E+20 to int loses precision
        string(0) ""                                                         (exit 0)
```

`NAN`, `INF` and `-INF` behave the same, returning `string(6) "abcdef"`. The values come from
AArch64's saturating `fcvtzs`. The pre-PR compiler was equally wrong here but silent; what is new
is that the program now asserts a specific php diagnostic while php is saying something else.

Both jurors independently noted that on x86_64 `NAN` sets ZF through `ucomisd`, so the `je done`
would skip the diagnostic entirely and the two arches would disagree on the same program. I cannot
execute that arm here and am not asserting it — worth checking while you are in the file.

### 6 — `function_exists()` denies a function the same program defines and calls ✅ proven

```php
if (!function_exists('pcntl_getcpu')) {
    function pcntl_getcpu() { return 7; }
}
var_dump(function_exists('pcntl_getcpu'));
var_dump(pcntl_getcpu());
```
```
php     bool(true)   int(7)
elephc  bool(false)  int(7)
```

`collect_top_level_user_functions` runs before the branch is pruned and deliberately skips `if`
bodies, so the post-declaration query folds against the builtin's target availability and answers
`false` — about a function that is right there and works.

---

## Worth addressing

### 7 — the polyfill idiom only works when the function name is a string literal ✅ proven

`$name = 'pcntl_getcpu'; if (!function_exists($name)) { function pcntl_getcpu() { return 7; } }`
gives `error[6:6]: pcntl_getcpu() is not available for the macos-aarch64 target`; php prints `7`. A
`const`, a concatenation and a variable all hit the same wall, because only a literal argument is
foldable. `docs/php/pcntl.md:167-170` describes the idiom without that restriction.

## Nit

### 8 — `Closure::fromCallable` reports `Undefined class: Closure` ✅ proven

`Closure::fromCallable('strtoupper')` gives `error: Undefined class: Closure` while
`$c instanceof Closure` is `true` and `get_class($c)` is `"Closure"`. A juror is right that the
change did not half-implement the method — `BUILTIN_CLASS_LIKE_SYMBOLS` only canonicalizes names for
constructs like `instanceof`, and the checker's class table never had `Closure`. What is new is the
contradiction between the two answers in one build.

---

## Not this PR — proven against `main`

- **A resource stored in a `Mixed` array stops being a resource.** `$a = [STDIN, 'x'];` gives
  `gettype($a[0]) === "integer"`, `is_resource($a[0]) === false`, `var_dump($a[0])` → `int(0)`,
  where php reports a stream resource. It reproduces identically on `main`, and
  `git diff origin/main...HEAD` touches none of `types/model.rs`, `lower_inst/arrays.rs`,
  `lower_inst/hashes.rs` or `codegen_support/value_boxing.rs` — `codegen_repr()` collapses
  `Resource → Int` at the boxing site, before the correct `Resource → tag 9` arm can be reached.
  **This is why the summary's "boxed Mixed … resource stringification" item is not actually
  fixed**: the shared `__rt_mixed_cast_string` does have a working tag-9 resource arm (I checked
  after a juror corrected me), and it never sees a resource tag. Worth its own issue — any
  `[$fh, $label]` or `$handles[] = $socket` loses the handle's identity silently.
- **A conditionally-declared class shadowing `ArrayObject` panics the compiler on `main` too** —
  see finding 2 for the half that is new.
- **A jury addition that did not survive:** a conditional `function var_export(...)` polyfill under a
  version guard compiles and returns php's value on this head; the predicted prelude-suppression
  regression does not reproduce for that name.

## Mergeability

`MERGEABLE / CLEAN`, 144 checks pass and one skipped, one commit behind `origin/main`. Finding 1
would not be caught by the current suite — the namespaced-polyfill test covers the
target-unavailable name, where the fold goes the other way.

## Verification and caveats

macOS ARM64 only; the x86_64 arm of the new coercion assembly is CI's word, and the `ucomisd` NaN
question above is specifically unverified here. The Linux-only pcntl surface is still exercised only
through its unavailability path, and `pcntl_sigwaitinfo` / `pcntl_sigtimedwait` remain outside the
oracle's reach because php does not define them on Darwin.

PR #734, opened 2026-08-20
feat: add recoverable owned-string cdylib ABI v3

nahime0:
## Summary

- add recoverable cdylib ABI v3 interoperability for exact `string -> string` exports while preserving existing scalar C signatures
- generate a deterministic `lib<stem>.h` with lifecycle declarations, status constants, ownership rules, and stable C-safe names for namespaced PHP exports
- recover from escaping PHP exceptions, allocation failures, dynamic `exit` paths, and guarded stack exhaustion without terminating the native host
- make `elephc_last_error()` distinguish an empty recorded message from the absence of an error, and make `elephc_free(NULL)` safe
- reject fatal paths reached through transitive calls, fixed object lifecycle, implicit object hooks, closures, `eval`, dynamic construction, foreign calls, or opaque dispatch
- keep ELF and Mach-O public symbol tables limited to lifecycle functions and explicit `#[Export]` trampolines
- fix exceptional cleanup for inlined owned return locals and static PCRE2 linking for the ARM64 musl workspace wrapper
- update the cdylib example, internals documentation, roadmap, and assembly-comment coverage checker

## ABI scope

ABI v3 applies the recoverable status boundary to every export. Existing scalar exports retain their C signatures and publish failure through `elephc_last_status()`. Exact by-value `string function(string)` exports return status directly and publish caller-owned `output_ptr` / `output_len` storage.

Array, object, callable, nullable, variadic, by-reference, multi-file, static-library, and multi-threaded runtime-context support remain outside this change. String results are binary-safe: `output_len` is authoritative and excludes the convenience trailing NUL. Successful results must be released with `elephc_free()`.

## Safety and recovery

- `exit` / `die` are rejected across transitive user calls, fixed constructors/destructors, implicit `__toString` / `__get` / ArrayAccess / Countable / JsonSerializable / Iterator / IteratorAggregate bodies, and statically invoked closures
- dynamic `exit` lowering also uses the existing cdylib escape as defense in depth
- `elephc_init()` arms the existing stack-limit guard; guarded exhaustion escapes as `ELEPHC_STATUS_RUNTIME_FAILURE`, leaving the host alive
- `eval` and opaque invocation surfaces are rejected when termination freedom cannot be proven
- allocation exhaustion inside an active cdylib boundary returns `ELEPHC_STATUS_ALLOCATION_FAILURE`
- escaping PHP exceptions return `ELEPHC_STATUS_PHP_EXCEPTION` and preserve an empty exception message as a non-NULL empty diagnostic
- exception cleanup walks abandoned PHP activation records so the host can safely call the library again
- concat scratch state is reset at boundary entry, completion, and lifecycle reset

## Testing

- `cargo build`
- `cargo test --test cdylib_tests` — 23 passed on Linux x86_64, including a real C host that survives deep-recursion failure and calls the library again
- `cargo test --lib test_stack_overflow_escapes_active_cdylib_boundary`
- `cargo test --lib emits_aarch64_owned_string_boundary`
- `cargo test --lib emits_x86_64_owned_string_boundary`
- `cargo test --lib test_emit_exit_recovers_through_cdylib_boundary`
- `cargo test --test codegen_tests test_direct_infinite_recursion_reports_controlled_fatal`
- `python3 scripts/check_asm_comments.py src/codegen/lower_inst/builtins/system.rs src/codegen/stack_guard.rs src/codegen_support/abi/bootstrap.rs src/codegen_support/cdylib.rs src/codegen_support/runtime/system/stack_guard.rs`
- `git diff --check`


nahime0:

Guikingone (commented):
## Review — recoverable owned-string cdylib ABI (`6e69a105c`)

Method: I built this PR's compiler and the merge base, compiled real shared libraries, and drove
them from C hosts I wrote to attack the contract rather than demonstrate it. Every claim below is
what a process actually did; where a host died, the exit code is the evidence.

**Most of the ABI is genuinely solid, and the parts that are hardest to get right are the parts
that hold.** A 31-check adversarial host passes completely:

| Checked | Result |
|---|---|
| success bytes, length, convenience NUL, `last_error` cleared | pass |
| embedded NUL in **and** out (`"a\0b"` → `"a\0b|3"`, len 5) | pass |
| exception → `STATUS_PHP_EXCEPTION`, outputs left NULL/0, message text | pass |
| **a working call immediately after an exception** | pass |
| empty exception message as a non-NULL empty string | pass |
| `elephc_free(NULL)`, NULL `output_ptr` / `output_len` / `input_ptr` rejection | pass |
| empty input, 1 MB result, second export, double `elephc_shutdown()` | pass |
| 20 000 alternating success/exception calls | pass |

Memory is clean, and I proved the measurement is sensitive enough to see a leak:

```
200k success + free      RSS +80 KB
200k exceptions          RSS  +0 KB      <- the failure path leaks nothing
20k x 64KB + free        RSS +192 KB
20k x 64KB, NO free      RSS +8000 KB    <- control: the detector works
```

`nm -gU` shows exactly the six lifecycle symbols plus the `#[Export]` trampolines, nothing else —
the visibility claim is true. Twenty-four fatal-path candidates (`range(1, PHP_INT_MAX)`,
`array_fill` to exhaustion, 2 M-element arrays, `chunk_split`/`str_replace`/`wordwrap`/`implode`
blowups, doubling concat) all converted to `STATUS_ALLOCATION_FAILURE` with the host alive. And
the static rejection really does fire on `exit`/`die`, `__destruct`, closures handed to
`array_map`/`usort`, named string callbacks, method calls, `unserialize`, `sprintf` with a
non-literal format, `str_repeat` with a non-constant count, `ob_start`, and a missing `require`.

Two things kill the host, and both have a fix already sitting in the tree.

---

### F1 — `exit()` reached through an implicitly-invoked magic method terminates the host

**Severity: blocker. Evidence: executed.**
`src/exports/safety.rs:104-114` (the `Op::ObjectNew` arm) and
`src/codegen/lower_inst/builtins/system.rs:810` (`emit_dynamic_exit`).

```php
<?php
class Boom { public function __toString(): string { exit(42); } }
#[Export]
function probe(string $s): string { $o = new Boom(); return "x" . $o; }
```

Compiles without a diagnostic. The C host prints `calling`, then **the process exits with code
42** — the PHP script's own exit code, chosen by the PHP author, delivered to the embedding
process. `HOST-ALIVE` never prints. Four entry points measured, each with a distinct code so the
route is unambiguous:

| construct | how it is reached | host exit |
|---|---|---|
| `__toString` | `"x" . $o` | 42 |
| `__toString` | `echo $o` | 43 |
| `offsetGet` | `$a['k']` | 45 |
| `offsetSet` | `$a['k'] = 1` | 48 |
| `__get` | `$b->missing` | 46 |

**Two layers, both fixable.**

*The analysis gap.* `validate_cdylib_call_graph`'s `Op::ObjectNew` arm hardcodes
`for method in ["__construct", "__destruct"]`. No other implicitly-invoked body is ever enqueued.
The asymmetry is its own proof: `__destruct` calling `exit` **is** rejected
(`reaches exit() through probe -> D::__destruct`), `__toString` calling `exit` is not.
`__invoke`, `__clone` and `__set` happen to be rejected too — but only incidentally, because they
lower through `Op::MethodCall`, which `opaque_dynamic_dispatch` already refuses. `__toString`,
`offsetGet`, `offsetSet` and `__get` do not go through that op, so nothing catches them.

*The runtime gap, which is the more important one.* `emit_cdylib_exit_escape` exists and works —
`grep -rn 'emit_cdylib_exit_escape' src/` returns exactly **one** call site,
`abi/bootstrap.rs:68` inside `emit_exit`. The `exit`/`die` lowering uses a **different** emitter,
`emit_dynamic_exit`, which emits a bare syscall with no boundary check:

```asm
mov x19, x0
bl  __rt_ob_flush_all
mov x0, x19
svc #0x80          ; no _elephc_boundary_active check, no __rt_throw_current
```

So the static rejection is the *only* thing between a cdylib export and a dead host, and any hole
in it — this one, or the next one — is unconditionally fatal. That the boundary-aware path exists
is the good news: making `emit_dynamic_exit` call `emit_cdylib_exit_escape` first turns every
future analysis gap into a `STATUS_RUNTIME_FAILURE` instead of a corpse. That is defence in depth
you have already written; it is just not wired to this emitter.

I would do both: traverse the implicit magic bodies, **and** route `emit_dynamic_exit` through the
escape.

Worth noting for calibration: `__toString` *throwing* is handled correctly — status 2, host alive.
Only `exit` escapes. This is a precise gap, not a broken boundary.

---

### F2 — the stack-overflow guard is compiled into the cdylib but never armed, so deep recursion SIGSEGVs the host

**Severity: blocker. Evidence: executed.**
`src/codegen/stack_guard.rs:36-44`, `src/codegen_support/runtime/mod.rs:73`.

```php
<?php
function deep(int $n): int { if ($n >= 100000) { return $n; } return deep($n + 1); }
#[Export]
function probe(string $s): string { return (string) deep(0); }
```

This is ordinary terminating code — a recursive walk 100 000 deep, with a base case.

```
php -n                     prints 100000
elephc, plain executable   Fatal error: Maximum call stack size reached. Infinite recursion?  (exit 255)
elephc, cdylib host        SIGSEGV, exit 139
```

Bisected: alive at depth 40 000, dead at 60 000. Unbounded and mutual recursion do the same.

**The guard is not missing — it is inert.** Counting in the emitted assembly for the same program:

```
_stack_limit prologue checks      executable 17    cdylib 18
__rt_stack_limit_init call sites  executable  1    cdylib  0
```

The checks are all there; the one-time call that publishes the floor is not, because
`emit_stack_limit_init_call` is emitted at *process entry* and a cdylib has no process entry. Your
own comment states the consequence exactly — `stack_guard.rs:40`: *"Until it runs, `_stack_limit`
is zero and every prologue check passes"* — and `runtime/mod.rs:73` warns that when the two sides
disagree *"the guard silently never fires."* That is precisely what happens here.

Calling `__rt_stack_limit_init` from `elephc_init()` (or once at first boundary entry) should turn
this segfault into the clean `STATUS_RUNTIME_FAILURE` the standalone target already produces.
Note the helper clobbers the C-ABI argument registers, so at boundary entry it has to run before
the arguments are staged — `elephc_init()` is the easier home.

---

### Nit

The PR title and body say **ABI v2**; the generated header says `ELEPHC_ABI_VERSION 3`, and the
last commit is `docs(cdylib): document recoverable ABI v3`. Worth reconciling before merge so the
release notes do not inherit the wrong number.

---

### Refuted

Five jurors ran: OpenCode against the worktree with a live compiler, GLM 5.2 / Kimi K3 /
MiniMax M3 against the inlined sources; Codex was out of credits. Eight hypotheses died.

- **GLM, blocker: `require "missing.php"` is not in the `exit|die|eval` match, so it fatals the
  host.** Refuted by execution — elephc resolves includes at compile time and refuses the build
  (`Required file not found`). The premise, that elephc has php's runtime `require` semantics, was
  the wrong part.
- **GLM, blocker: a closure handed to `array_map` bypasses the callback analysis because
  `string_callback_operand_index` only inspects string callbacks.** Refuted —
  `array_map(function ($x) { exit(1); }, $arr)` is rejected with `reaches exit()`, as are the
  `usort`-with-closure and `array_map('named')` variants.
- **GLM, important: an exception thrown from `__destruct` during unwinding of another exception
  fatals.** Refuted — status 2, host alive, both during unwinding and on a normal scope exit.
- **Kimi K3: `serialize(new Bomb())` with `exit` inside `__serialize`.** Refuted for that
  builtin — elephc rejects the build on an unrelated backend limitation. Its *reasoning* was
  right, though, and it is what pointed at the magic-method class in F1; `__toString` and
  `jsonSerialize` are the reachable instances.
- **OpenCode: `range(0, PHP_INT_MAX)` reaches `__rt_range_size_fail`'s raw syscall.** Refuted by
  its own probe — a `ValueError` is raised first and the boundary catches it (status 2).
- **OpenCode: `ob_start()`'s nested-handler fatal.** Refuted — `ObStart` is in the proof whitelist
  and is statically rejected.
- **OpenCode: casting an object with no `__toString`.** Refuted — that path uses `emit_exit`, the
  escape fires, host gets status 4 and survives. This is the contrast that shows the escape works
  when it is used.
- **A leak on the failure path.** Refuted — 200 000 exceptions moved RSS by 0 KB.

---

### Regression check and scope

`src/codegen/frame.rs`, `block_emit.rs`, `ir_passes/inline.rs`,
`runtime/arrays/heap_alloc.rs` and `strings/concat_scratch.rs` affect every compiled program, not
only cdylibs, so I ran a 43-example regression subset (strings, arrays, closures, exceptions,
objects, traits, generators, iterators, enums, references, variadics, sorting, JSON,
serialization) between the merge base and this PR: **42 identical, 1 differing only in `rand()` /
`mt_rand()` / `random_int()` output.** I intended the full 206-example corpus but this machine ran
out of disk mid-run; the subset is what I can vouch for, and I am flagging that rather than
implying wider coverage than I have.

CI on `6e69a105c` was green across 130 checks. The PR is currently **CONFLICTING / DIRTY**, so no
further CI will run until it is rebased.

nahime0:
Follow-ups from the post-Guillaume residuals, neither a merge blocker:

- #804 — finish the implicit-magic exit walk for `__set` / `__isset` / `__unset` / `__clone` (today they stay host-safe as opaque `MethodCall`)
- #805 — remaining raw `_exit` helpers must escape or be deleted (no known reachable host-killer; the same class already bit twice on this PR)

PR #735, opened 2026-08-20
feat: add PHP's complete iconv extension to native compilation and eval()

nahime0:
Adds PHP's `iconv` extension to elephc: all 10 functions and its 4 constants, on the
native compilation path and inside dynamic `eval()`, for `macos-aarch64`,
`linux-aarch64` and `linux-x86_64`. The compatibility table moves `iconv` from
**0/10 to 10/10**, AOT and eval alike.

| Function | Notes |
|---|---|
| `iconv()` | honors the platform `//TRANSLIT` and `//IGNORE` target suffixes |
| `iconv_strlen()` / `iconv_substr()` | count and slice by character in the selected charset |
| `iconv_strpos()` / `iconv_strrpos()` | character positions; out-of-range `$offset` throws `ValueError` |
| `iconv_mime_encode()` | RFC 2047 encoded-words with php-src's line-budget folding |
| `iconv_mime_decode()` / `iconv_mime_decode_headers()` | both decode modes; repeated field names become lists |
| `iconv_get_encoding()` / `iconv_set_encoding()` | the process-wide encoding trio |

## How it works

The conversions are performed by the platform `iconv` — glibc on Linux, GNU libiconv on
macOS — so the charsets your system supports are the charsets your program supports,
the same relationship PHP has.

- **`crates/elephc-iconv`** holds one Rust implementation of the extension's semantics.
  Compiled programs reach it through a panic-free `elephc_iconv_*` C ABI; Magician links
  the same crate as an `rlib`. Both backends therefore share one engine and one copy of
  the encoding trio rather than drifting apart.
- **Linking is pay-for-use.** Calling an iconv builtin auto-links the bridge;
  `--with-iconv` force-links it for indirect use; programs that never touch it neither
  reference nor link it.
- **The AOT lowering** stages every call into one uniform argument block, so all ten
  functions share one convention and one pair of runtime entry points on both
  architectures.

The engine follows `php-src`'s `ext/iconv/iconv.c` function by function, including the
behaviors the manual does not describe: an omitted `$encoding` resolves to
`iconv.internal_encoding` while an explicitly empty one resolves to `default_charset`;
the search pair scans one character at a time and can report a match found before a
malformed tail; the MIME decoders print their source charset as the literal `???`
placeholder.

## Verification

A 23,448-case differential corpus was generated from PHP 8.3 — every function, 12
charsets including `//TRANSLIT` and `//IGNORE` variants, malformed and fuzzed inputs,
all four decode modes, and the full `iconv_mime_encode()` option matrix — then replayed
through four independent execution paths. **All four agree with PHP on every case.**

| Path | How it was run |
|---|---|
| Rust engine | the crate's API directly |
| AOT `linux-x86_64` | compiled binary on the host |
| AOT `linux-aarch64` | cross-compiled, run under real arm64 glibc in Docker |
| Magician `eval()` | the whole driver loop inside one `eval()` |

Cases the corpus cannot express were checked separately against PHP: named and
out-of-order named arguments, static string-keyed unpacking, namespaced and
case-insensitive calls, `function_exists()`, `ReflectionFunction`, `--strict-php`
visibility, first-class callables for all ten functions, every `$options` receiver
shape, `PHP_INT_MIN`/`PHP_INT_MAX` offsets and lengths, NUL-carrying binary payloads,
200-byte charset names, 5,000-character payloads, and the encoding trio shared between
compiled code and `eval()` in both directions.

Committed coverage: codegen, error, eval-parity, heap-debug (clean heap for every result
kind), extension-discovery, CLI, and crate unit tests. `examples/iconv/main.php` produces
byte-identical output to PHP.

## Deliberate divergences from PHP

- `ICONV_VERSION` reports `unknown` — the spelling php-src itself uses when it cannot
  identify its provider — because the runtime libc version is not knowable while
  compiling ahead of time. Both backends answer the same value, so no program can
  observe a difference between them. `ICONV_IMPL` follows the target.
- A negative or `PHP_INT_MAX` `line-length` makes php-src abort with an
  integer-overflow fatal; elephc returns php-src's `Buffer length exceeded` warning and
  `false` instead.

Both are documented on the new `docs/php/iconv.md` page.

## Notes for reviewers

Three pre-existing defects surfaced during verification. **None is touched by this PR**
and each reproduces on `main` with no iconv call involved; they are recorded at the end
of `.plans/iconv-extension.md`:

1. A ternary whose branches call a user function produces a wrongly typed result slot —
   `$a = $cond ? f(1) : f(2);` with `f(): string` yields `0`.
2. A runtime-`null` optional argument is passed as its zero value rather than as an
   absent argument; core `substr()` behaves identically.
3. A positional array spread into a builtin (`str_replace(...$args)`) is rejected at
   compile time, while the documented static string-keyed form works.

The generated builtin reference is regenerated in its own commit (242 files, mostly
`sidebar: order` renumbering) so it does not bury the hand-written pages. Suggest
`topic:strings` as a maintainer label.

## Commits

`feat: add the elephc-iconv engine crate` · `feat: link the iconv bridge on demand` ·
`feat: declare the ten iconv builtin contracts` ·
`feat: compile PHP's iconv builtins to native code` ·
`feat: run PHP's iconv builtins inside eval()` ·
`feat: expose the ICONV_* constants on both paths` ·
`test: cover the iconv extension end to end` · `docs: add an iconv example` ·
`docs: document the iconv extension` · `docs: list --with-iconv among the bridge flags` ·
`docs: regenerate the builtin reference for iconv`


nahime0:

Guikingone (commented):
## Review — iconv extension (`d89b06344`)

Method, so you can weigh each claim: I built two compilers — this PR's head and the merge base
`b7325fe0`, where iconv does not exist — and differentially tested every claim against **php
8.5.6 on this machine (macOS, `ICONV_IMPL = libiconv`, `ICONV_VERSION = 1.11`)**, byte for byte.
Nothing below is reasoning about what php "should" do; each line is what php actually printed.

**The central design decision is the right one and it shows.** Delegating transcoding to the
platform `iconv(3)` instead of porting a charset engine means the parts that would have been a
minefield simply are not one:

| Surface | Cases | Divergences |
|---|---|---|
| `iconv()` — 41 charsets x 10 inputs x {plain, `//IGNORE`, `//TRANSLIT`} + round-trips | 1271 | **0** |
| `iconv_mime_encode()` — names x values x 11 option sets | 352 | **0** |
| `iconv_strlen/substr/strpos/strrpos` — 3 charsets | 420 | **0** |
| `iconv_mime_decode()` | 120 | 10 lines, one root cause |
| Large inputs (up to 20 000 chars, buffer growth) and embedded NUL bytes | 30 | **0** |
| 206-example corpus, merge base vs PR | 206 | 0 regressions |

Binary cost is genuinely on demand, measured: `<?php echo 1;` is **51 720 bytes on both
compilers**, byte-identical; a program that calls `iconv()` is 494 728.

Two things need a decision before merge.

---

### F1 — the `Q` decoder cannot fail, so malformed encoded-words are silently accepted

**Severity: blocker. Evidence: executed.**
`crates/elephc-iconv/src/mime/quoted_printable.rs:57` (`decode`), consumed at
`mime/decode.rs:435`.

`decode` returns `Vec<u8>`. It has no failure path. Its docblock states the premise:

> `_` becomes a space and `=XX` becomes the escaped byte; a malformed escape is kept literally,
> **which is what php-src's decoder does with unparsable input.**

That last clause is false. php rejects the encoded-word:

```php
<?php var_dump(iconv_mime_decode("=?UTF-8?Q?=ZZ?=", 0, "UTF-8"));
```
```
php     bool(false)                       + Warning: iconv_mime_decode(): Malformed string
elephc  string(3) "=ZZ"                   no diagnostic
```

Under `ICONV_MIME_DECODE_CONTINUE_ON_ERROR` php re-emits the **raw encoded-word** verbatim
(`=?UTF-8?Q?=ZZ?=`); elephc still returns `=ZZ`. Nineteen malformed/edge `Q` payloads measured,
**14 diverge** — every one in the same direction, elephc accepting what php refuses:

| payload | php (default) | php (CONTINUE_ON_ERROR) | elephc (both) |
|---|---|---|---|
| `=ZZ` `=Z` `=G0` `=0G` `=4` `=?` | `false` | raw word | `=ZZ` `=Z` `=G0` `=0G` `=4` `=?` |
| `a=ZZb` `a=Zb` `==41` | `false` | raw word | `a=ZZb` `a=Zb` `=A` |
| `a=` (trailing `=`) | `a` | `a` | `a=` |
| `= 41`, `=\t41` | `1` | `1` | `= 41`, `=\t41` |

`iconv_mime_decode_headers` inherits it — `Subject: =?UTF-8?Q?=ZZ?=` gives php `false` and elephc
`{"Subject":"=ZZ"}`.

**Why this is the blocker and not a nit.** `iconv_mime_decode` exists to parse untrusted mail
headers. php's contract is "`false` unless you explicitly asked me to continue"; a caller that
tests `=== false` is a caller that currently stops attacker-controlled bytes. Under elephc that
test never fires, and the bytes that come through are neither what was sent nor what php would
have produced.

**Fix.** The seam already exists. `base64::decode` sits at the same call site and the surrounding
code already routes a failure through `lenient()` at `decode.rs:438-455`. Make
`quoted_printable::decode` return `Option<Vec<u8>>` (or `IconvResult`), return `None` on a `=`
that is not followed by two hex digits, and hand it to that same path — strict mode gets
`IconvError::MalformedString`, lenient mode gets `append_literal_range` over the raw word, which
is exactly php's fallback. And drop the docblock sentence, which is what sent the implementation
this way.

None of your tests cover a malformed `Q` payload; adding one is the same edit.

---

### F2 — `iconv_strlen` / `iconv_substr` disagree with php on stateful charsets, and elephc is the one that is right

**Severity: important (a decision, not a defect). Evidence: executed.**
`crates/elephc-iconv/src/search.rs`.

`日本語` in ISO-2022-JP is `1b2442 467c 4b5c 386c 1b2842` — a shift-in escape, three two-byte
characters, a shift-out escape:

```
                       php            elephc
iconv_strlen           4              3
iconv_substr(0,1)      ESC$B 467c ESC(B    ESC$B 467c ESC(B     (agree)
iconv_substr(1,1)      4b             ESC$B 4b5c ESC(B
iconv_substr(2,1)      5c             ESC$B 386c ESC(B
iconv_substr(3,1)      38             ""
round-trip of slice 1  "K"            "本"
round-trip of slice 2  "\"            "語"
```

php loses the shift state after the first character: it counts four "characters" in a
three-character string and returns raw bytes that decode as ASCII punctuation. elephc converts
the whole subject to the UCS-4LE superset in one pass, so it counts 3 and every slice
round-trips. **elephc's answer is the correct one** — but `search.rs`'s docblock says "offsets and
lengths are character counts exactly like php-src", and for stateful charsets that is not true.

Nothing in the PR exercises a stateful charset: `grep -rn 'ISO-2022\|EUC-JP\|SJIS' tests/` over
the branch returns nothing. So this is untested in both directions.

Your call, and either is defensible — but please make it explicit rather than leave it implicit:
match php's quirk, or keep the correct behaviour and say so in `docs/php/` next to the other
documented divergences. One caveat on my measurement: php's iconv behaviour here may well differ
between libiconv and glibc, so I can only vouch for macOS/libiconv 1.11. Given
`d89b06344` ("stop pinning glibc's transliteration in the iconv tests") you have already met that
class of problem on this branch.

---

### Smaller things

**N1 — `default_charset` is hardcoded, and the reason given for it is not true.**
`encoding_state.rs:119` says "elephc has no ini surface, so `default_charset` is the fixed `UTF-8`
PHP ships with." elephc does have one: `--ini KEY=VALUE`, threaded from `main.rs:143` through
`pipeline.rs:72` into the preludes. Measured:

```
php -n                              iconv_get_encoding('internal_encoding') = UTF-8
php -n -d default_charset=ISO-8859-1                                        = ISO-8859-1
elephc                                                                      = UTF-8
elephc --ini default_charset=ISO-8859-1                                     = UTF-8   (flag accepted, ignored)
```

Hardcoding may still be the right call for now, but the justification should be "not wired up
yet", not "no ini surface".

**N2 — no `Deprecated:` for `iconv_set_encoding`, and your own example shows it.**
Running `examples/iconv/main.php` under php emits two
`Deprecated: iconv_set_encoding(): Use of iconv.internal_encoding is deprecated` lines (php 8.5
deprecates all three slots); elephc emits none, and that is the *only* difference in that
example's output. In fairness this is a platform gap rather than something the PR introduced —
`grep -rn '"Deprecated' src/ crates/` finds no runtime deprecation channel anywhere in elephc, so
there was nothing to hook into. Worth a line in the docs either way, since `iconv_set_encoding`
is a function this PR ships and the deprecation is the only diagnostic php attaches to it.

**N3 — `['line-length' => -1]`.** php raises a fatal
`Possible integer overflow in memory allocation (1 * 18446744073709551615 + 5)` and aborts;
elephc returns `false`. elephc's behaviour is better; noting it only so the divergence is on the
record. All 17 other option shapes I tried are byte-identical.

**`ICONV_VERSION`.** php reports `1.11` here, elephc reports `unknown`. `lib.rs:51` argues a
cross-compiler cannot know the runtime libc version, which is correct and I would not change it —
but the docblock's phrasing ("the `unknown` spelling php-src itself uses when it cannot identify
its iconv provider") reads as if php were also saying `unknown`, and on both my platform and any
normal build it is not. `ICONV_IMPL` matches exactly.

---

### Refuted

Nine hypotheses died, five of them the jury's. Three jurors ran on the inlined engine sources
(GLM 5.2, Kimi K3, MiniMax M3) and one agentic juror against the worktree with both compilers
(OpenCode); Codex was out of credits.

- **GLM: `iconv_strpos`/`strrpos` swallow the conversion notice when the first byte is illegal.**
  Refuted by execution — `iconv_strpos("\xffA", "A", 0, "UTF-8")` emits
  `Notice: iconv_strpos(): Detected an illegal character in input string` on both, same text.
- **GLM: CONTINUE_ON_ERROR drops a word truncated at end of input instead of re-emitting it.**
  Refuted — php returns `""` for `iconv_mime_decode("=?UTF-8?Q?hello", CONTINUE_ON_ERROR)`, not
  the raw text; elephc returns `""` too. Its predicted php behaviour was the part that was wrong.
- **GLM: `mime_encode` over-decrements the line budget for a non-ASCII field name.** Refuted —
  byte-identical across six value sizes straddling the fold boundary, including its own repro.
- **`base64::decode` is infallible too, so it must be permissive like `Q`.** Refuted: 44 malformed
  base64 payloads (bad padding, `=` mid-string, non-alphabet bytes, truncated quantum, `AA==AA`)
  are byte-identical to php.
- **The decoder's state machine mishandles folding / inter-word whitespace / unterminated words.**
  Refuted: 30 cases — `\r`, `\n`, `\r\n`, folded continuations, doubled spaces between adjacent
  words, nested `=?`, `=?UTF-8*en?`, `=?UTF-8?Q?a` with no terminator — all identical.
- **Buffer growth in `ffi.rs` breaks on long input.** Refuted: 1, 10, 100, 1 000, 5 000 and
  20 000-character inputs, identical lengths and MD5s.
- **The extension is not binary-safe.** Refuted: embedded NULs identical through convert, strlen,
  substr, strpos, mime encode and decode.
- **The eval-path iconv tests are invalidated by elephc's eval not processing `\xNN` escapes.**
  Refuted by reading them: they use literal UTF-8 (`"café"`, `"héllo"`, `"Prüfung"`), not escapes.
- **The 346 changed files regress something outside iconv.** Refuted: the 206-example corpus is
  identical between the merge base and the PR except for pid, cwd, inode, `rand()`, timestamps,
  pointer addresses, ephemeral ports and free-disk bytes — plus `examples/iconv`, which does not
  compile on the base for the obvious reason.

---

### Out of scope — pre-existing, verified on the merge base

Not yours to fix here, but they touch this surface so you should know:

- **A `null` length reaching a nullable-int parameter through an array element is read as 0.**
  `substr("abcdefgh", 2, [null][0])` returns `""` on the merge base too, where php returns
  `"cdefgh"`. `iconv_substr` inherits it — it accounted for 30 of the differing lines in my
  stateful-charset run, and for `iconv_substr($s, 2, $c[1], 'UTF-8')` inside a `foreach`.
  A literal `null` and a plain variable are both fine; only the array-element path is wrong.
- **`eval()` does not process `\xNN`, `\u{...}` or octal escapes in double-quoted strings.**
  `eval('echo bin2hex("caf\xc3\xa9");')` gives `caf\xc3\xa9` verbatim on both compilers, while
  `\t` and `\n` work. Same on the merge base.
- Warnings and notices go to stderr without php's ` in FILE on line N` suffix, and an uncaught
  exception prints only its first line, never the `Stack trace:` block. Both reproduce on the
  merge base with non-iconv builtins.

One process note: the PR is currently **CONFLICTING / DIRTY**, so no further CI will run until it
is rebased. The last run against `d89b06344` was green across all 130 checks, including the three
`Build & Archive` platforms and `PDO Live Databases`.

nahime0:
@Guikingone the review findings should be addressed on the current head — feel free to take another look.

Guikingone (commented):
## Re-review — iconv extension (`5a05d2550`, was `d89b06344`)

I rebuilt the branch at its new head and re-ran the measurements from the previous review against
php 8.5.6. Every point I raised has been addressed, most of them more thoroughly than I asked.
One residue of the blocker remains, and the fix surfaced two new divergences worth a decision.

### F1 (blocker) — fixed

`decode` now returns `Option<Vec<u8>>` and `None` routes through the existing strict/lenient seam,
which is exactly the shape the seam wanted. Re-running the 19 payloads from the previous review
plus 8 new edge shapes, in **both** modes:

**26 of 27 now match php exactly**, including byte-exact re-emission of the raw encoded-word under
`ICONV_MIME_DECODE_CONTINUE_ON_ERROR` — `=?UTF-8?Q?a=\t?=` comes back with its literal tab intact.
`iconv_mime_decode_headers` inherits the corrected behaviour. The two tests I said were missing are
there: `malformed_quoted_printable_words_respect_continue_on_error` over `=ZZ`, `=Z`, `a=Zb`,
`==41` in both modes, and `malformed_quoted_printable_header_fails`.

### The one residue — `=` followed by more than one whitespace

**Severity: important. Evidence: executed.** `mime/quoted_printable.rs`, the
`if matches!(first, b' ' | b'\t') { index += 3; continue; }` arm.

php skips the **whole run** of whitespace after `=` and then consumes one more character. This
code skips a fixed three bytes, so it is right for exactly one whitespace character and drifts by
one position for every additional one:

| payload | php | elephc |
|---|---|---|
| `= 41` | `1` | `1` ✔ |
| `=  41` | `1` | `41` |
| `=   41` | `1` | ` 41` |
| `=    41` | `1` | `  41` |
| `=\t\t41` | `1` | `41` |
| `= \t41` / `=\t 41` | `1` | `41` |
| `a=  41b` | `a1b` | `a41b` |
| `=  41=42` | `1B` | `41B` |
| `ab=  ` | **`false`** | **`ab`** |

The last row is the one that matters: it is the original F1 failure mode in miniature — elephc
accepting a word php rejects, so a `=== false` guard still does not fire. The others are wrong
output rather than a bypassed guard.

The fix is the same shape as the one you just made: replace the fixed `index += 3` with a loop
that advances past every `' '`/`'\t'`, then take one character, returning `None` if the input ends
first. Your existing test would extend to `=  41` in a line.

### Two new divergences, and I think elephc is right in both

Both are confined to `iconv_mime_decode_headers` with a malformed word under
`CONTINUE_ON_ERROR` — the raw re-emission path this PR just built. Single-word
`iconv_mime_decode` matches php in every shape I tried; it is only the header-block parser that
differs.

```php
$h = "Subject: =?UTF-8?Q?=ZZ?=\r\nFrom: a@b.c\r\n";
iconv_mime_decode_headers($h, ICONV_MIME_DECODE_CONTINUE_ON_ERROR, 'UTF-8');
```
```
php     {"Subject":"=?UTF-8?Q?=ZZ?=\r","From":"a@b.c"}     ← trailing CR leaks into the value
elephc  {"Subject":"=?UTF-8?Q?=ZZ?=","From":"a@b.c"}
```

With LF-only line endings it is worse on php's side — the header boundary is lost entirely and the
rest of the block is swallowed into the value:

```
php     {"Subject":"=?UTF-8?Q?=ZZ?=\nFrom: a@b.c"}          ← one header, From: absorbed
elephc  {"Subject":"=?UTF-8?Q?=ZZ?=","From":"a@b.c"}
```

Both look like php artifacts of re-emitting raw bytes without re-establishing the line boundary,
and elephc's answers are the defensible ones. Same situation as the stateful-charset call: I would
keep the current behaviour and put a line in `docs/php/iconv.md` next to the one you added, so the
divergence is recorded rather than discovered. Well-formed words are identical in every mode, so
nothing in normal use is affected.

### Everything else I raised is closed

- **F2 (stateful charsets)** — you made the call explicit and backed it: `search.rs` now says the
  UCS-4LE approach "intentionally avoids php-src's provider-dependent loss of shift state",
  `docs/php/iconv.md:77` documents it with the `日本語`/ISO-2022-JP example, and
  `stateful_charset_slices_round_trip` tests it. That was the gap — it was untested in both
  directions before.
- **N1 (`default_charset`)** — the justification is now "elephc's compile-time `--ini` surface does
  not wire that directive into this bridge yet", which is the accurate reason.
- **N2 (`Deprecated:`)** — `docs/php/iconv.md:163` records that php 8.5 deprecates all three slots
  and that elephc has no runtime deprecation channel. The example's output is now identical to
  php's apart from exactly those two lines.
- **`ICONV_VERSION`** — the docblock no longer implies php also reports `unknown`.

### Verification

- 27 `Q` payloads × 2 modes, plus 12 whitespace shapes, 8 trailing-`=` shapes and 10 header-block
  shapes, all against php 8.5.6 on macOS ARM64 / libiconv 1.11.
- A broad sweep of the rest of the surface — `iconv_strlen`, `iconv_substr` (positive, negative and
  past-end), `iconv_strpos`, `iconv_strrpos`, `iconv`, `//IGNORE`, `//TRANSLIT`, `iconv_mime_encode`
  in B and Q, `iconv_mime_decode` with two words and with a mixed good/bad pair,
  `iconv_mime_decode_headers`, `iconv_get_encoding`, malformed input, `ICONV_IMPL` — is identical to
  php on every line except `ICONV_VERSION`, which is the documented cross-compilation case.
- `examples/iconv/main.php` matches php except the two documented `Deprecated:` lines.
- `cargo test -p elephc-iconv`: 22 passed.
- CI is green on `5a05d2550` (129 pass, 1 skipping) and the PR is MERGEABLE / CLEAN.

### Refuted

Three jurors re-attacked the rewritten decoder. Two of them — GLM 5.2 and MiniMax M3 —
independently reported the same **blocker**: that the `break` on a trailing `=` silently accepts a
word php rejects, predicting `false` from php for `=?UTF-8?Q?ab=?=`. Both are wrong, and their
agreeing did not make them right. php drops a trailing `=` and succeeds, exactly as this code
does — measured across eight shapes (`ab=`, `a=`, `=`, `=41=`, `a_b=`, in strict and
continue-on-error, and through `iconv_mime_decode_headers`), all identical. The `break` is correct.

Kimi K3 derived php's whitespace-run rule from first principles and landed on the same divergence I
had already measured, which is a useful independent confirmation of the fix shape above but not a
separate finding.

PR #736, opened 2026-08-20
feat: add exception-aware dead-code elimination

nahime0:
## Summary

- Adds post-typecheck exception-flow analysis for AST dead-code elimination, using exact and constrained throwable domains instead of a single may-throw bit.
- Routes catch handlers in PHP source order from explicit throws, statically proven operator failures, fixed-point direct-function and exact-receiver method summaries, and caught-variable rethrows.
- Makes catch and finally guard invalidation path- and exception-type-aware while preserving conservative barriers for unresolved calls, open dispatch, external or trait-provided constructors, and complex control flow.
- Integrates immediately before the declaration-reachability pass landed in #714, allowing declarations referenced only by removed exceptional paths to be pruned before EIR lowering.

## Behavior

- Builds the throwable hierarchy from checker-provided class and interface metadata, with case-insensitive PHP symbol matching.
- Tracks exact thrown classes and constrained unknown domains, including source-order subtraction across catch handlers and the `Throwable = Exception | Error` root partition.
- Computes callable summaries to a fixed point for direct user functions and exact method targets.
- Preserves caught-variable domains through nested try blocks and simple aliases or reassignments so `throw $e` retains useful type information.
- Narrows catch-side invalidation to writes on paths that can actually enter the selected handler, including by-reference mutations performed by the throwing call.
- Excludes unconditional `exit` and `die` paths from finally-entry invalidation because PHP does not execute finally after process termination.

## Conservative boundaries

The optimizer retains an unknown throwable domain for dynamic or unresolved calls, late-bound/open instance dispatch, external constructors, trait-provided methods or constructors, and operand-dependent runtime failures that cannot be classified exactly. These paths continue to keep potentially matching handlers and their guard invalidations.

## Pipeline integration

The post-typecheck optimizer constructs exception summaries from the complete checked program. Exception-aware DCE then removes impossible handlers and non-observable paths. The declaration-reachability pass from #714 runs afterward, reconciles `CheckResult`, and prunes functions, classes, methods, externs, and libraries that are no longer reachable before EIR lowering.

A dedicated integration regression verifies that a function referenced only by a disjoint catch is removed by declaration reachability after exception-aware DCE drops that handler.

## Tests

- `cargo build`
- `cargo test --test codegen_tests dead_code_elimination::tries` - 51 passed
- `cargo test --lib optimize::tests::dce::tries` - 17 passed
- `cargo test --test codegen_tests declaration_reachability` - 42 passed
- `cargo test --lib optimize::reachability` - 74 passed
- Targeted `rustfmt --check` and `git diff --check`

CI remains responsible for the complete macOS AArch64, Linux AArch64, and Linux x86_64 matrix.


nahime0:

Guikingone (commented):
## Review — exception-aware dead-code elimination (`3a38f43d0`)

Method: I built this PR's compiler and the merge base `b7325fe0`, then held the pass to the
obligation an optimizer actually has — **observational transparency**. Every claim below is what
three binaries printed: `php -n` 8.5.6, the baseline compiler, and this PR's.

### The pass is sound on everything I could throw at it

**68 hand-written exception programs plus the 206-example corpus: zero PR-vs-baseline
differences.** The 11 differing examples differ only in `rand()`, pointer addresses, pid, inode,
ephemeral port, timestamps and free-disk bytes.

The hand-written set was built to make the analysis wrongly conclude "impossible", and covers
source-order handler selection, parent-class catches, `Throwable` and the `Exception`/`Error`
partition, case-insensitive catch spelling, rethrow of a caught variable, throwing constructors,
trait methods, closures, `array_map` callbacks, mutual recursion, interface-typed dispatch,
`parent::` calls, static methods, deep chains, anonymous classes, first-class callables, a throw
inside `finally`, `exit` in `try`/`finally` including the conditional form, by-reference mutation
by the throwing call, and `intdiv`/`%` by zero.

### And it demonstrably does something

Not just "no difference" — the pass earns its place on the shape it targets:

```php
class Never1 extends Exception { public function tag(): string { return "n1"; } }
class Thrown1 extends Exception {}
function onlyFromDeadHandler(): string { return "dead-fn"; }
try { throw new Thrown1("t"); }
catch (Never1 $e) { echo onlyFromDeadHandler(), $e->tag(); }
catch (Thrown1 $e) { echo "reached"; }
```

| symbol in emitted EIR | base | PR |
|---|---|---|
| `onlyFromDeadHandler` | 2 | **0** |
| class `Never1` | 8 | **0** |
| method `tag` | 2 | **0** |
| `Thrown1` (live) | 7 | 7 ✔ |
| total EIR | 260 lines | **186** (−26%) |

And with a live handler the emitted EIR is **byte-identical** to the baseline — it does not
perturb code it cannot prove dead. The composition with #714 works exactly as the description
claims.

### One thing worth deciding before merge, and it is not soundness

On the repository's own corpus the pass currently buys almost nothing. Across the **24 examples
that contain a `catch`**, comparing emitted EIR line counts base vs PR:

```
json-exception             base=650    pr=636    -14
(23 others)                                        0
────────────────────────────────────────────────────
total                      325302  →  325288      -14   (0.004%)
```

Fourteen lines, all in one example, for 3109 added lines across 19 files and a new 1090-line
analysis. The synthetic case above shows the pass *can* pay off handsomely, so this is not a claim
that the design is wrong — it is that the shape it pays off on (a handler that provably cannot
fire, holding declarations nothing else references) is rare in the code this project ships. A
Symfony-sized application with deep exception hierarchies may look quite different, and if you
have such a measurement it belongs in the PR description; the examples corpus does not make the
case on its own.

### Refuted

Four jurors ran — Greptile's automated review, plus GLM 5.2, Kimi K3 and MiniMax M3 on the
inlined analysis sources. Every finding died under execution.

- **Greptile, merge blocker: "a condition-side caught-variable rebind can still cause DCE to
  remove a reachable nested catch."** Refuted twice. The general case does not compile — elephc's
  checker rejects rebinding a caught variable across classes before DCE ever runs
  (`Type error: cannot reassign $e from AEx to BEx`), across all eight rebind positions I tried:
  `if` condition, `while` condition, ternary, plain statement, conditional branch, nested try,
  `??=`, and `switch` subject. And the reachable specialization — rebinding to a *sibling* under a
  common caught base, which is the shape the described mechanism would actually break — is
  correct:

  ```php
  class Root9 extends Exception {}  class L9 extends Root9 {}  class R9 extends Root9 {}
  try { try { throw new L9("l"); }
        catch (Root9 $e) { if (($e = new R9("r")) !== null) { throw $e; } } }
  catch (R9 $x) { echo "R:", $x->getMessage(); } catch (L9 $x) { echo "L:", $x->getMessage(); }
  ```
  php, baseline and this PR all print `R:r` — the post-rebind class selects the handler, not the
  pre-rebind one.

- **Kimi K3, important/blocker: "a `static::` or overridable `$this->` call inside a method
  declared in a base class gets the base class's non-throwing summary even when the called class's
  override throws."** Refuted by execution on four shapes — late-static-bound `static::target()`,
  `$this->target()` with the override in the child, the same reached through a parent-typed
  parameter, and a mixed array of base and child instances. All catch correctly on both compilers.

- **GLM 5.2, important: `class_alias` handler removed because `is_subtype` answers `false` for an
  alias absent from the hierarchy.** Refuted as a defect of this PR — and it flagged its own
  uncertainty honestly. `catch (AliasException $e)` does not catch on **either** compiler; the
  exception escapes uncaught, identically before and after. See the note below.

- **MiniMax M3: implicit `__toString` throws missed for casts and interpolation.** Refuted as a
  defect of this PR — `(string) $obj`, `"v=$obj"` and `implode` with a throwing `__toString` behave
  identically on both compilers. Worth saying that the new `string_conversion.rs` models these
  conservatively, which is the safe direction, even though the runtime cannot currently deliver
  them as catchable.

- Mine, all refuted by execution: a mixed-case parent name breaking the hierarchy walk
  (`class_reaches_root` normalizes on read); `class_reaches_root` receiving an un-normalized
  `base` at `hierarchy.rs:146` (the function normalizes its own parameter); interface-extends-
  interface and interface-via-parent chains; the `exit`/`die` exclusion under a conditional exit.

### Out of scope — pre-existing, verified identical on the merge base

These all reproduce on the baseline, so none is yours to fix here, but three of them sit directly
under this PR's machinery and you should know they are there:

- **A plain local scalar written inside `try` is not visible in the `catch`.** php prints
  `written`, elephc prints `init`. Properties, statics, globals, array appends and echo side
  effects all survive — it is locals only. This matters for review because the PR's most delicate
  new machinery is the *narrowed* guard invalidation over exactly those writes: end-to-end, a
  mistake in that narrowing cannot currently be observed, because the value is discarded
  downstream regardless. The unit tests are the only thing standing behind it.
- **A throwing `__toString` is not catchable** — uncaught fatal instead of the `Error`/exception
  php delivers, for `.` concatenation, `(string)` casts and interpolation alike.
- **`class_alias` names do not work in a `catch` clause** — `class_alias('MyException', 'Alias');
  catch (Alias $e)` never catches, and the exception escapes.
- Also unsupported on both, so untested by this corpus: multi-catch `catch (A | B $e)`, catching by
  interface, `declare(strict_types=1)` TypeError, `implode` over a `__toString` object, and
  `$cond && exit(1)`.

### Process

CI was green on `3a38f43d0` across all 130 checks. The PR is now **CONFLICTING / DIRTY**, so no
further CI will run until it is rebased.

nahime0:
Ok, I opened a follow-up to add more complex examples: #799. Meanwhile this PR is mergeable if there are no other defects.

PR #781, opened 2026-08-24
elephc monitor: PHP-level profiling, one command in every environment

Guikingone:
Profiles an elephc program at the **PHP level**, with one command that does not
change with the environment. Because elephc is the compiler, a profile names
your PHP functions and can say *why* runtime time was spent — heap allocation,
Mixed cell boxing, reference counting — which an interpreter-based profiler
cannot produce.

## One command, four targets

Profiling in production is normally a *different tool* answering a *smaller
question*: an approximation you learn to read differently from the exact numbers
you get on a laptop. Here both answers come from one instrumentation and agree
with each other, so the smaller question is a narrower slice of the same picture
rather than a second thing to reconcile.

| You have | You run | What happens |
|---|---|---|
| a source | `elephc monitor shop.php` | built with `--with-monitoring`, then read — full table |
| a binary | `elephc monitor ./shop` | read, if it carries monitoring — full table |
| a running service | `elephc monitor host:9411` | read through its endpoint: sampled, or `--exact` for one request |
| a local socket | `elephc monitor /run/app.sock` | the same, on the same machine |

A target this command **launches** is measured for its whole run and reports the
full per-function table, with every export: Speedscope, pprof, DOT, HTML.

A service **already running** answers through its endpoint, and there the default
is the sampler: folded stacks with their sample counts and per-stack
allocations — a statistical view of where CPU time goes, not a per-function
table. `--exact` returns the measured table for the next request that completes.
What sampled does not carry: per-function call counts, retained objects, and — in
a binary built with the combined `--with-monitoring` — per-route SQL and I/O
wait, because the exact runtime claims those slots. The exporters read the
sampled capture, and `--exact` with an export flag is refused rather than
accepted and silently ignored.

There is still no flag asking which *mechanism* to use. The one thing you choose
is which question you are asking of a live service, and the command says which
one it answered.

`https://` is for a service behind a TLS terminator: the endpoint itself speaks
the framed protocol above, not TLS, so pointing `https://` straight at
`ELEPHC_PROBE_ADDR` will not connect. When the scheme is `https://` the
certificate must validate against the system roots before a byte crosses.

## What it measures

`--with-monitoring` embeds the capability at compile time (or
`--with-monitoring=<names>` for named functions only). It is **dormant until
asked**: a monitored binary run on its own behaves and prints exactly like one
built without it, so the artifact you profile is the one you ship. A binary
built without it is **refused**, with both remedies printed, rather than
profiled approximately — a degraded answer that looks like the real one is worse
[...]

PR #790, opened 2026-08-25
feat(instr): the profiler hooks say which activation they belong to

Guikingone:
> **Stacked on #781.** Against `main` this shows 108 files, because #781's 31 commits
> are underneath it. **This PR is the last commit only — 5 files**, and the diff
> collapses to that once #781 merges. Draft until then.
>
> `git diff feat/php-monitor...feat/php-monitor-activation-id` is the review surface.

## What this fixes

Two known defects in the exact profiler had one cause, and the code in #781 said so:

> fixing it needs what the exit hook does not have — a way to tell which activation
> is ending — the same thing the recursive-catcher case needs, so both wait on one
> decision rather than two patches.

This is that decision.

The exit hook carried a function id. Two activations of one function are the same
thing to an id, so in ordinary recursive descent with error recovery — a `try`
around the recursive call, the deepest activation throwing — the frame the exit
resolved to was the corpse of the thrower rather than the frame returning:

```php
function f(int $n): void {
    if ($n > 0) { try { f($n - 1); } catch (\Exception $e) { recover(); } }
    else { throw new \Exception('x'); }
}
```

The activation that actually caught then stayed open on the shadow stack until its
own caller returned, and everything that caller did afterwards was charged to the
recursive function. Measured on a 120-tick fixture: **20 ticks moved off the
caller**, with the self times still summing to the root — so nothing looked wrong.

The same gap made an exit past the shadow-stack cap indistinguishable from a
tracked one. Activations beyond the cap are counted rather than pushed, and their
exits were reconciled by counting down — which a throw desynchronises, because it
destroys those activations without any exit arriving, so the count went on to eat
the exit of a frame that *was* tracked.

## What changed

`elephc_instr_enter` and `elephc_instr_exit` take the activation's frame pointer.

It is free to obtain. Every emitted function establishes a frame pointer in its
prologue — unconditionally; there is no leaf or frameless variant, and no tail
call — the enter hook runs last in that prologue and the exit hook first in the
epilogue, before the teardown. So `x29` / `rbp` already holds this activation's own
frame address at both sites, on both architectures, and passing it is one `mov`:

```asm
    ldr x2, [x9]          ; _gc_frees
    mov x3, x29           ; ← this activation
    bl  _elephc_instr_enter
```

Live frames have distinct addresses by construction, which is the only property the
runtime uses; an address reused by a later call is never compared against a frame
that has already returned.

**Two mechanisms go with it**, so this deletes more than it adds:

- `dropped_depth`, the count of activations past the cap, is gone — an exit whose
  frame is not on the stack closes nothing, which is exact where the count was a
  heuristic that a throw could desynchronise;
- `ambiguous_catch`, added in #781 to *report* the recursive guess the profiler had
  to make, is gone with the guess.

## What it costs

630,000 real calls (recursion, so nothing inlines away), each variant built by the
compiler that produced it, interleaved with the order alternating so drift lands on
both sides, child CPU time rather than wall clock, and a null comparison alongside:

| | before | after | delta | null |
|---|---|---|---|---|
| dormant | 27.24 ms | 27.51 ms | **+0.30 ns/call** (+0.68%) | 0.07 ms |
| profiled | 191.6 ms | 188.7 ms | −1.7 ns/call (−0.54%) | +0.6 ms |

Dormant is the tax on every binary that ships with monitoring: the `mov` runs
whether or not the hook does anything, and 0.30 ns/call is 2.8× the null floor —
small but real. While profiling there is **no measurable change**: the extra
argument and stored word pay for the two branches per exit that the
dropped-activation count no longer needs.

## Verification

- 49 instr / 33 probe / 1472 compiler / 129 pdo / 84 web tests, zero warnings,
  `aarch64-apple-darwin` and `x86_64-unknown-linux-gnu` both compile.
- Sentinel-mutated: replacing the frame-pointer match with the id match it replaces
  fails `a_recursive_call_that_throws_is_charged_to_the_activation_that_caught`;
  zeroing the stored pointer fails 14 tests.
- Emitted assembly checked directly, and end to end: a monitored binary compiled
  from a recursive PHP program that throws runs correctly and profiles correctly.

The tests gained a way to say *which activation* returns — `enter_sim` hands back
the frame pointer it assigned — which is the thing they could not express before.
The three tests that were about the removed mechanisms now assert the properties
those mechanisms were maintained for. One of them predicted this PR: *"the day an
activation identity reaches the exit hook, this test fails and says what it is
worth."* It did.

## Risk

This changes the hook ABI. A bridge archive built from an older `elephc-instr` and
linked against newer emitted code would receive a fourth argument it does not
declare. In practice `--with-monitoring` links both bridges together from the same
build, and `every_bridge_crate_is_in_the_build_lists` already holds the lists to
each other — but it is the thing to look at if something odd shows up. Both
mismatched compiler/binary pairs were also exercised by hand and profile correctly,
since the monitor↔binary protocol is untouched.


crates/elephc-probe/src/endpoint.rs
```diff
@@ -880,6 +898,12 @@ mod tests {
     /// Stub: nothing was armed, so cancelling is a no-op.
     extern "C" fn elephc_instr_capture_cancel() {}
 
+    #[no_mangle]
+    /// Stub: nothing holds a rendezvous these tests never open.
+    extern "C" fn elephc_instr_capture_blocked_by() -> i32 {
+        0
+    }
+
```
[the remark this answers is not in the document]
Guikingone: Correct, and it follows the convention rather than conflicting with it — `AGENTS.md` asks for `///` immediately before the function item **or its item attributes**, and this one sat between `#[no_mangle]` and the item.

Fixed in `6663a0ae2`. I checked the whole branch for the same shape rather than only this line: 28 occurrences exist across the changed files, but exactly one of them is introduced by this branch — this one. The other 27 predate it, including the three sibling stubs directly above, which is the shape I copied.

I moved all four stub docblocks anyway, since they are one contiguous block and correcting one of four would read as arbitrary. Doc-only: no signature, attribute or body changed, and the workspace still builds with no warnings.

crates/elephc-instr/src/lib.rs
```diff
@@ -376,33 +353,20 @@ impl State {
...
+        // exit for a tracked frame — the first was reconciled by counting
+        // dropped activations down, which a throw could desynchronise, and the
+        // second resolved to whichever frame carried the same id.
         let Some(index) = (match self.stack.last() {
             // The overwhelmingly common case, and the reason the search below
             // costs nothing in a program without exceptions.
-            Some(top) if top.id == id => Some(self.stack.len() - 1),
-            _ => self.stack.iter().rposition(|frame| frame.id == id),
+            Some(top) if top.fp == fp => Some(self.stack.len() - 1),
+            _ => self.stack.iter().rposition(|frame| frame.fp == fp),
```
nahime0, line 713: P2: a dropped activation can still match a retained frame whose native address was reused after an exception unwind.

At `MAX_STACK`, `note_throw` leaves the tracked frames in `self.stack` until the catcher exits. If that catch handler calls another instrumented function, `enter_at` drops the activation because the shadow stack is still full. The native call can reuse the frame pointer of one of the now-dead frames above the catcher, for example by calling the same fixed-frame-size function again. Its exit then reaches this `rposition`, matches the stale frame, closes everything above it, and only checks the mismatching function id after the stack has already been mutated.

I verified this with a temporary regression on `06d84f5`: fill the stack, record a throw caught at the root, enter and exit a dropped handler call using `stack[1].fp`, and assert the retained stack is unchanged. It fails with the stack reduced from `MAX_STACK` to only the catcher.

PR #791 does not cover this path: its current head predates the activation-aware hook ABI and still uses `dropped_depth`. Please add this collision regression here and ensure a dropped exit is ignored even when its frame pointer aliases an unwind-stale entry.

Guikingone: Confirmed and fixed in `9057e0214`. Your recipe reproduced on the first try, and the damage is larger than the grade suggests: the stack went from 65,536 frames to 1 — **65,535 frames closed by an exit that owned none of them**.

What makes it reachable is exactly what you describe, and it is worth stating as an invariant rather than a coincidence. The comment above the drop branch justified needing no bookkeeping like this:

> Its exit will arrive carrying a frame pointer that is not on the stack and close nothing, which is what makes this safe to do without tracking how many were dropped.

That holds while the shadow stack describes *live* frames. An unwind breaks it, and not by accident: the frames an exception destroyed stay on the shadow stack until the catcher exits — deliberately, since closing them at the instant of the throw is what keeps their cost on them rather than on the handler — while the native stack they occupied is already free. A call the handler makes is pushed onto that reclaimed space, so at a fixed frame size its frame pointer *is* the dead frame's address rather than merely able to collide with it.

I also checked the cheaper repair and rejected it: moving the `frame.id != id` test above the pop loop is not sufficient, because your own example is a handler calling the same function that occupied the slot, where the ids match.

So a dropped activation now records its frame pointer, and an exit carrying the innermost of those closes nothing and consumes it. Identities rather than a count — which is what this replaced, and what a throw could desynchronise. That same reasoning settles the other half: a throw destroys every dropped activation above it and none of their exits will ever arrive, so it clears the list outright. Left standing, those entries are worse than stale — the next dropped call reuses the freed address and *its* exit would be swallowed against a dead entry, leaving its own identity behind for the one after that.

Two regression tests, one per half, each verified by mutation. Worth noting why the existing dropped-exit test did not catch this: it exits with a frame pointer below every live frame, so it cannot collide and passes whatever the matching rule is.

Suites: `elephc` bins 1654, lib 1483, `error_tests` 1437, `elephc-instr` 60, `elephc-probe` 58, `elephc-web` 87, `elephc-pdo` 130. No warnings.

One note on reachability, which is why I agree with P2 rather than arguing it up: the drop path needs 65,536 live frames, and a default 8 MB stack dies of native exhaustion at roughly 51,200 — so this needs a larger stack to reach at all. It is fixed anyway because the failure is silent and total, and because the invariant it rested on was written down in a comment that had stopped being true.

crates/elephc-instr/src/lib.rs
```diff
@@ -339,12 +340,13 @@ impl State {
...
-        // the stack for a frame that was never pushed and pop everything on the
-        // way. The call still counts — it did happen — but nothing else does.
+        // so its inclusive time is never credited. Its exit will arrive carrying
+        // a frame pointer that is not on the stack and close nothing, which is
+        // what makes this safe to do without tracking how many were dropped. The
+        // call still counts — it did happen — but nothing else does.
         if self.stack.len() >= MAX_STACK {
             self.dropped += 1;
-            self.dropped_depth = self.dropped_depth.saturating_add(1);
+            self.dropped_fps.push(fp);
```
nahime0: [P2] Keep dropped-activation bookkeeping bounded.

`MAX_STACK` is explicitly the guard that keeps runaway recursion from growing the profiler state without bound, but this `push` creates a second shadow stack with no cap: every activation after 65,536 allocates another `usize`, and both `note_throw` and `reset` only `clear()` the vector, retaining its peak capacity for the lifetime of the thread. On the larger-stack configurations where this branch is reachable, one recursive spike can therefore keep growing (and permanently retain) profiler heap memory even though the tracked stack is capped.

The activation identities are not needed for this overflow case. A bounded active-dropped depth works because exits are LIFO; clear it on every throw (destroyed dropped calls have no exits), increment it for handler-created dropped calls, and consume it before searching the retained frame-pointer stack. That also handles the stale-frame-pointer collision fixed by `9057e0214` while preserving the cap; the `dropped_depth` implementation on #791/base is the existing model to reuse. Please add a regression asserting that overflow bookkeeping stays O(1)/bounded as depth grows past `MAX_STACK`.

Guikingone: Bounded, and without the LIFO assumption — `a2ff9f0`.

You are right that the cap was the point and that this vector walked around it. Two things changed:

- Nothing is recorded outside an unwind. An identity exists for exactly one reason: the shadow stack can hold frames that have **returned**, whose addresses the native stack has handed back out. That is true between a throw and its catcher exiting, and at no other time. Outside that window every frame on the stack is live, so an unknown frame pointer closes nothing on its own and needs no identity — which is what the original reasoning said, and what remains true of that case. So the runaway-recursion spike you describe records nothing at all.
- The list is released when the unwind ends, not merely cleared. `clear()` kept the peak capacity for the life of the thread, which was the second half of your point.

On reusing `dropped_depth`: a counter is O(1), but it is only correct if dropped activations exit in the order they were entered. That is the assumption a previous round attacked, and I did not want to buy the bound back with it — a generator or fiber resumed out of order leaves the counter describing a frame that is not the one exiting. Gating on `unwinding.is_some()` gets the same bound (nothing recorded when there is nothing stale to guard against) without depending on the ordering.

Two regressions, both mutation-verified:

- `dropped_identities_cost_nothing_outside_an_unwind` — fills the stack past `MAX_STACK`, asserts `dropped_fps` is empty **and** `capacity() == 0`, then runs a throw/catch cycle and asserts the capacity is handed back when the catcher exits.
- `a_dropped_exit_out_of_order_leaves_nothing_behind` — the membership case, now staged under an unwind since that is the only state where identities exist.

crates/elephc-instr/src/lib.rs
Guikingone: Follow-up on the last line of your report — "only checks the mismatching function id after the stack has already been mutated". That ordering is fixed too, in `a2ff9f0`.

I said above that moving the id test above the pop loop was not *sufficient*, which is true and is why the identity scheme is what closes the hole you found. But not-sufficient was the wrong reason to leave it where it was: as written it was a late abort rather than a guard, so a tracked exit whose frame pointer aliased a dead frame destroyed everything above it and only then noticed the names did not match. It now compares before anything is closed. One line, and it is the ordering the field's own invariant describes.

Guikingone:
### Arbitration round — `a2ff9f0`

Three more commits since the last review. The last one comes from a fifth reader asked to **arbitrate** the previous rounds rather than repeat them, and it returned NOT READY over a fix I had shipped two commits earlier. Flagging it here because a reviewer should not have to find this on their own:

**The epoch fix starved the capture it was protecting.** `offer_capture` disarmed unconditionally after winning the slot. A worker holding a slice for a capture that had since ended had that slice correctly discarded by the reader — and had already cleared `armed`, which belongs to the capture now running. No later request offered; the endpoint waited out its timeout and answered *"no request completed within 30s"* while requests were completing. A wrong profile traded for no profile and a false reason, which is worse than what it replaced. The offer now refuses to publish or disarm for a capture that has ended and gives the claim back; the reader re-arms when it discards a slice that reached it anyway.

**And its regression test asserted the opposite of what the code did.** The test staged the stale publish by writing header words by hand, which bypassed `offer_capture` and therefore bypassed the disarm — so it asserted "the capture stays armed" while the production path was clearing it, and stayed green. That is the exact class of defect this branch has spent its rounds correcting, committed inside the correction. `offer_capture` is now a thin reader of the current identity over `offer_capture_for`, which the test drives with a stale one: a seam in the production path, not a mock of it.

Also in the round:

- **`event_report` walked the event table to `count + 1`.** The route counter stops at `OTHER_ROUTE_INDEX` because overflow returns the shared bucket *without* claiming a slot — so the walk stopped one short of the id that bucket carries, and every exact I/O event charged to `<other>` was recorded and never printed. The existing overflow test checked name resolution and never the report.
- **The unwind's clamp guarded the clock and passed four counters through raw**, each subtracted the same way by `close_frame`. Either the guarded case is unreachable and the guard describes a phantom, or it is reachable and four counters wrap to ~1.8e19. All five are clamped.
- **Four comments that claimed more than the code holds**, found by reading them adversarially rather than by a failure — including one that denied SIGPROF exposure the PDO bridge's tokio workers actually have (they are built from the PHP thread and inherit an unblocked mask; that is profile pollution, not corruption, and it is written down now instead of denied).

Suites: `elephc` bins 1654, lib 1483, `error_tests` 1437, `elephc-instr` 67, `elephc-probe` 60, `elephc-web` 87, `elephc-pdo` 130. Workspace builds with no warnings. Every behavioural fix carries a regression test and each was verified by mutation.

Guikingone:
### Coroutines — `259d911`

Sweeping what else rests on the shadow stack turned up a case where the frame is not a second activation at all: it belongs to code that is not running. `docs/beyond-php/profiling.md` already had it written down as a known limit, with the fix in one sentence — *"pop the frame on yield and push it on resume"*. This is that.

**What it reported.** A `yield` and a `Fiber::suspend` switch stacks; they do not return. The body's frame stayed open across everything the consumer did next, which made it the caller of that work and the owner of its cost:

```php
foreach (producer() as $v) { $t += heavy(); }
```
```
drain     incl 100.0%  self  0.3%   self 36.6 µs
producer  incl  99.8%  self  0.2%   self 23.0 µs
heavy     incl  99.6%  self 99.6%   self 13.9 ms
```

and the call graph carried `producer → heavy ×3`. That edge does not exist — `drain` calls `heavy`, three times. A fiber body measured at 17 µs reported 99.7% the same way. Self time was the one column that stayed right, because the consumer's work was counted as a child; so the hotspot table was correct and everything shaped like a tree named the wrong function.

Now `producer` reads **0.3%**, with `drain → heavy ×3` and `drain → producer ×1` and nothing under `producer`.

**Closed, not set aside.** The first version parked the frame and rebased its stamps at the resume. It measured right and was wrong about the case that matters most: a coroutine that is never resumed. Most generators are abandoned — a `foreach` that breaks, a `current()` with no `next()` — and no exit hook ever arrives for them, so a frame waiting to be rebased reports a function that ran as one that did not (`self 0 ns` for 16 µs of work). A suspension really is the end of a span, so it closes the frame; that gets the abandoned case right *and* buys the whole of `close_frame` unchanged — self time, the caller's child charge, the inclusive span, the edge weight, the overdrawn check, the unwind's pending charge.

**What the sentinel caught that the tests did not.** Seven mutations, five killed on the first pass. Both survivors were real gaps and one was a real defect:

- The restore walked the parked group **in reverse**, putting a callee underneath its own caller. It survives every single-frame group — which is every group a program actually produces — so no test and no measurement could see it: the mutation that "broke" it made it correct. A group holds more than one frame only when frames sit *above* the suspending one, which is what an unwind leaves behind, and that is now the test.
- Reopening the inclusive span is indistinguishable from not reopening it with one activation live; it takes two activations of one function to tell them apart.

**Two limits, stated rather than papered over.** A suspension reached from *inside* a call the body made parks only that innermost frame — the hook is told one frame pointer, and the frames below it on that coroutine stack cannot be told from the consumer's without a record of where the coroutine began. `yield` is only ever lexically in the body, so every generator suspension is the exact case; `Fiber::suspend()` is the one that can be nested. And at most 4,096 coroutines can be suspended at once, past which the suspension is refused, the frame left where it was, and the count reported.

Suites: `elephc` bins 1654, lib 1483, `error_tests` 1437, `elephc-instr` 75, `elephc-probe` 60, `elephc-web` 87, `elephc-pdo` 130. No warnings. The emitted bracket was read in the assembly, and the three programs produce their correct output — which is what says the register saves around the hook are balanced.

src/codegen/lower_inst/static_method_calls.rs
```diff
@@ -233,7 +233,14 @@ pub(super) fn lower_static_fiber_suspend(ctx: &mut FunctionContext<'_>, inst: &I
...
     abi::emit_push_reg(ctx.emitter, abi::int_result_reg(ctx.emitter)); // preserve the boxed suspend value for target-specific argument loading
     abi::emit_pop_reg(ctx.emitter, abi::int_arg_reg_name(ctx.emitter.target, 0)); // pass the boxed suspend value as runtime helper argument 1
+    // The same park a `yield` gets, for the same reason: this call switches
+    // stacks rather than returning, so the frame stays open across whatever the
+    // fiber's caller does next and is read as that work's caller.
+    let value_arg = abi::int_arg_reg_name(ctx.emitter.target, 0);
+    let result_reg = abi::int_result_reg(ctx.emitter);
+    crate::codegen::frame::emit_instr_suspend(ctx, &[value_arg]);
     abi::emit_call_label(ctx.emitter, "__rt_fiber_suspend");
+    crate::codegen::frame::emit_instr_resume(ctx, &[result_reg]);
```
nahime0, line 243: **[P1] Restore the profiler frame before any non-returning resume path.** `__rt_fiber_suspend` does not always return here: it raises `FiberError` before switching when `Fiber::suspend()` is invalid, and after a switch it calls `__rt_throw_current` when `Fiber::throw()`/`Generator::throw()` supplied a pending exception. In all of those paths this post-call hook is skipped, so the activation remains parked; a local catch then runs without its frame, later suspend/exit hooks cannot reconcile it, and its work/callees are misattributed. The generator yield lowering has the same ordering. Please put the restore at the validated runtime resume boundary (and before rethrow), or otherwise make every exceptional/non-returning path restore the activation before control reaches PHP handlers.

src/codegen/lower_inst/generator_instructions.rs
```diff
@@ -64,7 +64,14 @@ pub(super) fn lower_generator_yield(ctx: &mut FunctionContext<'_>, inst: &Instru
     } else {
         "__rt_gen_suspend"
     };
+    // The yielded key and value are staged; parking here keeps the cost of
+    // staging them on the body that is suspending, and the two registers are
+    // saved because the hook is an ordinary call that takes its own arguments.
+    crate::codegen::frame::emit_instr_suspend(ctx, &[key_arg, value_arg]);
```
nahime0, line 70: **[P1] Cover the outer suspensions performed by `yield from`.** This brackets a direct `yield`, but `lower_generator_yield_from()` calls `__rt_gen_delegate` without hooks, and that helper invokes `__rt_gen_suspend_delegated` internally for every forwarded value. The outer generator therefore remains on the shadow stack while its consumer runs. With the current head, the repository generator example reports `pipeline()` at 88.4% self time even though its body only yields once and delegates to `base()`. Please bracket each outer delegated suspend/resume in the runtime (with the outer activation identity), and add a monitoring-active regression; the dormant output test cannot observe this.

crates/elephc-instr/src/lib.rs
```diff
@@ -373,40 +470,170 @@ impl State {
...
+            return;
+        }
+        if self.parked.len() >= MAX_PARKED {
+            self.parks_refused += 1;
+            return;
+        }
+        // Innermost first, with the rest still standing, so each span is charged
+        // to its own caller exactly as its return would have charged it.
+        let mut frames = Vec::new();
+        while self.stack.len() > index {
```
nahime0: **[P2] Do not park and reopen frames that an unwind already destroyed.** Between `note_throw()` and the catcher's exit, frames above the catcher remain only as stale shadow-stack entries. If that catcher yields or suspends, this loop closes those dead frames at the later suspension timestamp, stores them, and `resume_at()` reopens them as live callees. Subsequent work is then charged through activations that no longer exist. The new multi-frame test does not model this: it never calls `note_throw()`, so both frames are live. Please reconcile the active unwind using its killer timestamps and park only the surviving coroutine frames; the regression should exercise a catch handler that suspends after a real `note_throw()`.

crates/elephc-instr/src/lib.rs
```diff
@@ -373,40 +470,170 @@ impl State {
...
+            let frame = self.stack.pop().expect("the index is inside the stack");
+            frames.push((frame.id, frame.fp));
+            self.close_frame(frame, t, a, f, io, w);
```
nahime0: **[P2] Record trace spans when suspension closes a frame.** `exit_at()` appends `(id, frame.t_enter, t)` while tracing is enabled before calling `close_frame`, but this path calls `close_frame` directly. Every pre-yield segment is therefore absent from Chrome/Perfetto output, and an abandoned generator has no trace span at all even though the aggregate table accounts for it. Please route both exit and suspension through a shared trace-and-close helper, or emit the same bounded trace record here, with a trace-enabled regression.

crates/elephc-instr/src/lib.rs
```diff
@@ -373,40 +470,170 @@ impl State {
...
+        // Paired with the id, the way `exit_at` pairs them: a freed coroutine
+        // stack is handed back out, so a frame pointer alone names an activation
+        // only among the live ones.
+        let Some(index) = self.stack.iter().rposition(|frame| frame.fp == fp) else {
+            return;
+        };
+        if self.stack[index].id != id {
+            return;
+        }
+        if self.parked.len() >= MAX_PARKED {
```
nahime0, line 539: **[P3] A refused park can resume an abandoned activation at the same `(id, fp)`.** At the cap this returns with the current frame still on `stack`. If an older abandoned coroutine of the same function owns a parked group at a reused coroutine-stack address, the post-switch `resume_at()` finds that old group, removes it, and pushes a duplicate activation on top of the current one. The address-reuse test only changes the function id, so it misses this same-id case. At minimum, `resume_at()` should refuse to restore when the current `(id, fp)` is already live (the refusal case); a generation/token per park would make the pairing unambiguous.

nahime0 (commented):
Code/logic findings only, as requested. I left documentation, process, and still-running CI out of this review.

Guikingone:
### `yield from` — `eb3cc1d6`

The coroutine bracket in the previous commit sits at the `yield` lowering, in the generator body's own frame. A **delegated** yield never passes through it: `__rt_gen_delegate` drives the inner generator on the *outer* one's coroutine stack and forwards each of its yields out by calling `__rt_gen_suspend` itself, from hand-written assembly. So for the whole of a `yield from`, the outer body stayed open across the consumer's work exactly as every coroutine did before — the same defect, on the delegated half, and nothing in that commit went near it.

Measured on a generator that yields, delegates to another, then yields again, with the consumer doing real work per iteration:

```
outer   incl 52.6%   self 0.6%   calls 1   self 163.5 µs
```

52.6% is not a coincidence — two of the four iterations are driven through the delegation, and both were charged to `outer`. It now reads **0.3%**, for the 52 µs of work it did.

**What it costs is one level of nesting.** The park goes around the whole delegation rather than around each forwarded yield, because each forwarded yield is inside a runtime helper this cannot reach. The inner generator is started *by* that helper, so it enters while the outer body is already parked and is recorded as a callee of the consumer: `drain → inner` where PHP says `outer → inner`. Recovering that needs the suspension bracketed inside the delegate loop, which is shared by every generator in the program and therefore has no function id to check a frame pointer against — the id guard beside the pointer is what three review rounds put there, and it is not worth trading for an edge. The consumer's `foreach` is also what actually drives those resumes, so the flattened edge names something real. Written down in `docs/beyond-php/profiling.md` beside the other limits rather than left for a reader to find in a graph that looks plausible.

The delegating program's output matches `php -n` exactly (`79999600000`), which is what says the register saves around the two hooks are balanced at this site too. Plain generator and fiber cases unchanged at 0.3% and 0.1%.

---

Also closing the two audit items that needed checking rather than changing:

- **Every return path of an instrumented function runs its exit hook.** Five `FunctionContext`s take the enter hook and all five are followed by `emit_function_epilogue` in the same function body; every non-main `Return` goes through `emit_function_return_epilogue`; both epilogues run `emit_instr_exit` first, before the early return for cleanup-free functions; `Unreachable` traps rather than returning. The other two dozen `abi::emit_return` sites are hand-emitted helpers with no `FunctionContext`, and the one that has one marks its function synthetic — the first thing `emit_instr_enter` checks.
- **A zombie worker cannot splice a profile.** It *can* hold a claim, though: an unreaped process answers `kill(pid, 0)` and keeps the `/proc` entry the start id is read from, so it reads as alive and as the same process. That is a wedge, not a splice — the endpoint reports the claim it cannot revoke. `4b88c8f8` says so in `claim_may_be_taken`, as a property of that function rather than as a claim about whichever supervisor sits above it.

Suites: `elephc` bins 1654, lib 1483, `error_tests` 1437, `elephc-instr` 75, `elephc-probe` 60, `elephc-web` 87, `elephc-pdo` 130. No warnings. CI green on `4b88c8f8` including the eval codegen shard that timed out on an earlier push — that was runner variance (the whole shard ran 623 s against 492 s green), not a regression.

src/codegen/lower_inst/generator_instructions.rs
Guikingone: Already on the branch — `eb3cc1d6`, which this review predates (it was written against `4b88c8f8`). Same diagnosis as yours, same place: `__rt_gen_delegate` calls `__rt_gen_suspend_delegated` itself, so the bracket at the `yield` lowering never sees a forwarded yield.

Your 88.4% figure is the reason I am answering rather than just pointing at the commit, because **the example still prints 100% and always will**. `examples/generators/main.php` does 16 µs of work in total, and shares are relative to the largest inclusive time — `pipeline` is genuinely the biggest contributor there whatever the attribution does. So that example cannot tell the two behaviours apart in either direction.

Giving the consumer work the numbers can attribute does. Same shape, `heavy()` in the drain loop:

```
heavy     incl 100.0%  self 99.7%  calls 3  self 14.0 ms
pipeline  incl   0.3%  self  0.3%  calls 1  self 35.5 µs
base      incl   0.0%  self  0.0%  calls 1  self  4.1 µs
```

On my own four-line delegating case, `outer` went from **incl 52.6%** to 0.3% — 52.6% because two of four iterations are driven through the delegation, and both were charged to it.

One cost is written down rather than hidden: the park goes around the whole delegation, not around each forwarded yield, so the inner generator starts while the outer is already parked and reads as a callee of the *consumer* — `drain → inner` where PHP says `outer → inner`. Bracketing each forwarded suspension needs the hook inside the delegate loop, which is shared by every generator and so has no function id to check a frame pointer against; that is the same wall as your fiber comment below, and I did not want to buy an edge by dropping the id guard three rounds of review put there.

You are right that the dormant output test cannot observe any of this, and I have no automated monitoring-active regression — every number above is a manual measurement. That gap is real and I have not closed it.

crates/elephc-instr/src/lib.rs
Guikingone: Confirmed and fixed in `995b64c87`. You are right about the test too: it staged two **live** frames, which is a shape that cannot occur, and it never called `note_throw()`.

The reasoning that makes it a one-line rule rather than a reconciliation: a suspension is reached by a *call*, so a coroutine never has a live frame above it. Everything above the suspending frame is unwind debris by construction. So only the suspending frame is parked, and the rest is closed exactly the way an exit closes it — at the instant of the throw that killed each one, per frame, five counters clamped.

`exit_at`'s resync loop moved into `close_frames_above` and both callers use it. Two copies of that loop would only have had to be corrected twice.

The regression is the case you asked for: a catcher that throws, keeps its debris, then yields 1000 ns later. It asserts the dead frames are closed at the throw's instant and not at the suspension's, and that the parked group holds only `(catcher_id, catcher_fp)`. Mutation-verified — restoring the old `remove(index)` turns it red.

crates/elephc-instr/src/lib.rs
Guikingone: Confirmed and fixed in `995b64c87`. Both events end a span and only one recorded it, which is the same "two paths that must agree" shape as the resync loop above.

Both now go through one `trace_span(id, from, to)`, bounded by `TRACE_CAP` and counting into `trace_dropped` exactly as before. The regression turns tracing on, runs enter → suspend → resume → exit and asserts **two** spans — `(0, 100, 130)` and `(0, 500, 520)` — so it fails on a missing pre-yield stretch and equally on a span that wrongly spanned the suspension. Mutation-verified.

Worth noting which case made this more than cosmetic: an abandoned generator has no exit at all, so before this it had no span whatsoever while the aggregate table did account for it.

crates/elephc-instr/src/lib.rs
Guikingone: Confirmed and fixed in `995b64c87`. You are right that the address-reuse test only varied the function id and so missed the same-id case.

I did not add a generation token, because the state already carries the fact that settles it: a suspension takes its frame OFF the stack, so a resume can never legitimately find its own `(id, fp)` still live there. `resume_at` now refuses when it does — which is exactly and only the refused-park case, and needs no new field.

The regression fills the table to `MAX_PARKED` with an older abandoned group already sitting at the address, has the new activation's park refused, and asserts both that no duplicate is pushed and that the old group is still parked rather than consumed by somebody else's resume. Mutation-verified.

src/codegen/lower_inst/static_method_calls.rs
Guikingone: **Confirmed, not fixed, and not fixable where the bracket lives.** You are right on all three paths, and I read `emit_suspend` to check rather than take it:

- `suspend()` outside a fiber → `__rt_fiber_throw_state_error` → `__rt_throw_current`, before the switch;
- a live `unserialize()` → the same, also before the switch;
- a pending `Fiber::throw()`/`Generator::throw()` after the switch → `__rt_throw_current`.

All three skip the post-call hook, so the activation stays parked while a local `catch` in that same function runs. Its own exit then finds no frame and closes nothing. The first two are the worse shape in one way — the park happens before the helper has even decided it will throw, so an ordinary `try { Fiber::suspend(1); } catch (FiberError $e) { … }` outside a fiber loses its frame — and the third is the one real async code hits.

The `yield` lowering has the same ordering, as you say.

What I am *not* doing is a half-fix. The restore has to happen inside the helper, and the helper is shared by every coroutine in the program, so it has no function id to pair with a frame pointer — which is the guard three review rounds put there and the reason `resume_at` refuses a mismatched group at a reused coroutine-stack address. Two things make it tractable rather than a redesign, for whoever takes it:

- `__rt_throw_current` already carries the pattern: a null-checked function-pointer slot (`elephc_instr_throw_fn`), loaded and called from runtime assembly, that a dormant binary pays one load and a not-taken branch for.
- The identity is available at each of those points without an id: `[x29]` on AArch64 / `[rbp]` on x86_64 is the saved caller frame pointer, which *is* the `fp` the park was keyed on, read from a live frame chain rather than guessed. `rposition` already takes the newest group at an address, which is the one this suspension parked.

That is six emission points across two architectures plus the generator helper, in hand-written assembly I can only exercise on AArch64 locally — the `linux-x86_64` shard is the only oracle for the other. I would rather leave it standing and named than land it half-verified, so it is not in `995b64c87`.

The other four findings are fixed and answered in their own threads.

Guikingone: Fixed in `e2d93d1a`, along with the missing oracle you named in the other thread — the second is why the first is verifiable at all.

**The restore is inside the helper, keyed by the coroutine.** You were right that the call site cannot do it. What the helper has at all three points is `_fiber_current`, and that turns out to be the *better* key rather than a fallback: a coroutine has one suspension in flight at a time, so it names the parked group exactly. A frame pointer would not — read from the frame chain it gives the PHP function for a direct `Fiber::suspend()` and gives `__rt_gen_suspend`'s own frame for a `yield`, since generators reach the same helper one level deeper. The park now records `_fiber_current` beside the frame pointer for this reason and no other.

`elephc_instr_unpark_fn` follows `elephc_instr_throw_fn` exactly: a null-checked function-pointer slot, so a binary without the capability pays one load and a not-taken branch, and only on a path that is already raising. Six emission points across the two architectures, each inserted immediately before a `call`/`bl` that was already there — so the stack alignment at that point is unchanged by construction. `resume_at` and the new `unpark_at` share one `restore`, including the guard against opening a second copy of a live activation.

**What says it works.** `test_cli_monitor_restores_a_coroutine_resumed_into_a_throw` asserts an edge that must EXIST, where its sibling asserts ones that must not: a fiber body whose `catch` calls `heavy` has to show `body → heavy`. Without the hook the graph reads `drive → heavy` — a call the consumer never makes — which is exactly what the mutation produces:

```
edges without the unpark hook:
    drive -> heavy [label="100% (×1)"];
    drive -> body  [label="0% (×1)"];
```

Four mutations over the whole chain — the hook not emitted, the slot never filled, the park forgetting its coroutine, the coroutine hooks not parking at all — all four red.

One limit I will state rather than imply: I can only exercise AArch64 locally. The x86_64 sites are written from the same three anchors, and the `linux-x86_64` shard is the oracle for them.

Suites: `elephc-instr` 77, `elephc` lib 1483, bins 1654, `error_tests` 1437, codegen fibers 95, generators 83, cli 31. Eleven PHP programs match `php -n` byte for byte, the fiber-throw fixture included.

src/codegen/lower_inst/generator_instructions.rs
Guikingone: Closed too, in `e2d93d1a` — this was the sharper half of your comment and I said so rather than fixing it at the time.

There are now two tests that run with the profiler **active**:

- `test_cli_monitor_does_not_charge_a_coroutine_for_its_consumer` compiles a fixture with `--with-monitoring`, runs it under `monitor`, parses the `incl` percentage per row, and asserts the consumer dominates while `outer`, `inner` and `body` each stay under 10%. It then reads the `--dot` graph and asserts none of them has an edge to the function the consumer calls. Thresholds are loose deliberately: the gap is three orders of magnitude, so anything under 10% passes and every shape of the defect lands far above it.
- `test_cli_monitor_restores_a_coroutine_resumed_into_a_throw` covers the fiber P1, asserting the opposite direction — an edge that must exist.

Neutralizing the coroutine hooks turns the first red with `outer was charged 76.1% of the program while suspended`. That is the first test on this branch that could have caught the original defect; everything before it was either a unit test driving `State` directly, which cannot see an emitted hook, or the dormant-output test, which asserts the profiler stays quiet and is therefore blind to what it says when it speaks.

src/optimize/propagate/stmt/control.rs
```diff
@@ -229,12 +229,34 @@ pub(super) fn propagate_try_stmt(
...
+        crate::optimize::propagate::invalidation::block_invalidation(&try_body).apply(&mut catch_env);
+        catch_env
+    };
     let catches: Vec<_> = catches
         .into_iter()
         .map(|catch| crate::parser::ast::CatchClause {
             exception_types: catch.exception_types,
             variable: catch.variable,
-            body: propagate_block(catch.body, env.clone()).0,
+            body: propagate_block(catch.body, catch_env.clone()).0,
```
nahime0: **[P2] Remove the catch binding from the environment before propagating its body.** This shared `catch_env` still contains any incoming constant for the name bound by a catch when the try body does not write that name. For example, `$e = 'stale'; try { throw new RuntimeException('fresh'); } catch (RuntimeException $e) { return $e->getMessage(); }` reaches this branch with `$e` still known as the string `stale`; on the current head the optimizer substitutes that string and compilation fails with `method call receiver for PHP type Str`, while PHP returns `fresh`. `simulate_catch_constant_env()` already removes the binding for exit-path simulation, but the actual body propagation does not. Please derive an environment per catch, remove `catch.variable` from it before `propagate_block`, and add this shadowing regression.

nahime0 (commented):
Code/logic findings only on the current contributor head `3198886`.

crates/elephc-instr/src/lib.rs
```diff
@@ -373,40 +489,228 @@ impl State {
...
+    /// is skipped in all three, so without this the activation stayed parked
+    /// while its own `catch` ran: the handler's work was charged to whatever
+    /// frame was below, and the function's own exit later found no frame and
+    /// closed nothing.
+    ///
+    /// Keyed by the coroutine rather than by a frame pointer because that is what
+    /// the helper has. A coroutine has one suspension in flight at a time, so the
+    /// key is exact; `rposition` still takes the newest, which is that one.
+    fn unpark_at(&mut self, coro: usize, t: u64, a: u64, f: u64, io: u64, w: u64) {
+        let Some(at) = self.parked.iter().rposition(|group| group.coro == coro) else {
```
nahime0, line 583: **[P3] Only unpark a group when the current suspension actually parked it.** `unpark_at()` selects the newest group by `coro` even when the immediately preceding `suspend_at()` returned without parking. At `MAX_PARKED`, a new activation can reuse an abandoned coroutine's address under a different function id, have its park refused, and then enter one of the non-returning runtime error paths. This lookup finds the abandoned group; the `(id, fp)` live guard in `restore()` does not match the different live activation, so the stale group is consumed and pushed above it. A temporary regression reproduces the stack growing from one live frame to two. Please carry a success/generation handshake for the current park (or an equivalent transient refusal marker) so the runtime unpark cannot restore an older occupant, and cover the different-id reused-address case.

src/optimize/propagate/stmt/control.rs
Guikingone: Confirmed and fixed in `61706c07`. Your repro reproduced exactly — `method call receiver for PHP type Str`, where PHP answers `fresh`.

You put your finger on what makes it the same defect as the last two: `simulate_catch_constant_env` has always removed the binding, and the body propagation had not. That is the third time on this branch that two walks over one block disagreed about what is live in it — after `exit_at`'s resync loop, which a suspension had to be taught to share, and the trace span, which only the exit recorded. Derived per catch now rather than shared, since each clause binds its own name.

The regression asserts both directions:

```php
function shadowed(): string {
    $e = 'stale';
    try { throw new RuntimeException('fresh'); }
    catch (RuntimeException $e) { return $e->getMessage(); }
}
function untouched(): string {
    $keep = 'kept';
    try { throw new RuntimeException('x'); }
    catch (RuntimeException $e) { return $keep; }
}
// "fresh|kept"
```

The second half is deliberate: a name the try body cannot write and the catch does not bind still folds, so the fix stays a rule about the **binding** rather than a blanket clear of the handler's environment. Mutation-verified — dropping the `remove` turns it red.

crates/elephc-instr/src/lib.rs
Guikingone: Confirmed and fixed in `61706c07`, though not with a generation token — the state already carries what settles it, and using that turned out to be sound rather than a patch.

**A park drops any group standing under its own `coro`, before the cap check.** Two live coroutines never share a fiber address and one coroutine never has two suspensions in flight, so a match means the fiber that parked that group is *gone*. Nothing will ever resume those frames either: they were closed when they parked, and their own resume is keyed on a frame pointer their fiber no longer owns. So the abandoned occupant is not merely shadowed, it is removed — which also bounds the table better than the cap alone did, and makes your sequence unreachable rather than guarded. Either the older group is dropped and the new park succeeds, or there was none and the unpark finds nothing.

You were also right that my existing test missed it, and it turned out to be worse than that: `a_refused_park_does_not_resume_an_abandoned_twin` staged the twin under the **same** coroutine, which a park now removes before it can matter — so it was describing the behaviour this replaces. Rewritten around the pair that is still reachable: same `(id, fp)`, *different* fiber, which is a later generator of the same function landing on a reused coroutine **stack** under a new fiber object. That is what the live guard in `restore` is actually for, and it now has a test that can only pass because of it.

The new regression is yours: an abandoned group at a fiber address, the table full, a **different** function taking that address, its park refused, and then the non-returning path. It asserts the stack holds `[9]` and not `[9, 7]`, and that the dead group is gone rather than waiting to be found again. Mutation-verified — removing the `retain` turns it red.

Suites: `elephc-instr` 78, `elephc` lib 1483, bins 1654, `error_tests` 1437, codegen exceptions 61, cli 31, generators 83, plus the `--doc` gate.

nahime0 (approved):

PR #793, opened 2026-08-27
docs: the merged docs describe a profiler four commits behind the merged code

Guikingone:
#790 landed fourteen commits. Its CHANGELOG carries **one** entry for the coroutine work — the first of them — and four later commits changed behaviour with no user-facing record at all. Two pages still describe the profiler as it was before those four.

This is that gap, not a style pass.

### Four behaviour changes that merged without an entry

| commit | what changed |
|---|---|
| `eb3cc1d6` | A `yield from` left the delegating generator on the profiler's stack for the whole delegation — **52.6%** of a program's inclusive time for 52 µs of its own work, and an edge to a function it never called. The bracket sits at the `yield` lowering, and a delegated yield never passes through it. |
| `e2d93d1a` | The runtime's suspend helper has three exits that never return to the suspension site (`Fiber::suspend()` outside a fiber, a live `unserialize()`, a `Fiber::throw()` delivered on resume), and all three reach PHP handlers with the activation still parked. A fiber body whose `catch` did the work showed the *consumer* calling it. |
| `995b64c8` | A `catch` handler that yields swept the frames its exception had destroyed into the park; a suspension left no mark on the Perfetto timeline; a refused park could resume somebody else's group. |
| `61706c07` | A caught exception read as whatever its variable held before the `try` — which stops the compile rather than answering wrongly; and a suspension now releases what was parked under its own coroutine. |

Each gets an entry, in the section and the register the surrounding ones use.

### Two pages describing the old behaviour

- **`docs/beyond-php/profiling.md`** said the 4,096 cap was the whole story for parked coroutines. A suspension *also* releases what was standing under its own coroutine, since a fiber address is handed to the next fiber once the first is freed.
- Its **timeline** section said each call becomes one nested slice. A coroutine has not done that since a suspension started ending a span: a generator resumed three times is three slices, and an abandoned one still gets the slice it ran for.

### Two counts that had drifted

`docs/internals/architecture.md` — `runtime/strings/` is 97 files, not 95, and `src/codegen/shared_*.rs` is 4, not 3. `shared_state.rs` was in neither the count nor the description, though it is the module-wide state that dedupes callable descriptors and owns the label counter.

---

### What was checked and was already right

Recorded so the next pass starts further along rather than re-deriving it. Across all **1,130** pages:

- frontmatter complete everywhere; no page opens with an `#` heading Astro already renders from it;
- no broken relative link — the two candidates are PHP in prose (`$arr[0](1, 2)`, `$callbacks[0](...)`);
- every `src/` path the docs cite exists;
- every one of the **39** flags `src/cli.rs` parses, and every flag `elephc --help` and `elephc monitor --help` advertise, appears in `docs/compiling/cli-reference.md`;
- the phase list in `docs/compiling/compilation-pipeline.md` matches every `record_since` label in `src/pipeline.rs`, in order;
- the README's directory tree and the architecture module map name only directories that exist.

Docs only — no code, no tests touched.


Guikingone:
Folded into #794 — both change `docs/beyond-php/profiling.md` and reviewing them apart would mean reading that page twice. `650c913` is now the second commit there, unchanged; the mechanical checks were re-run after the merge (1,130 files, 0 frontmatter problems, 0 stray headings, 0 broken links).

PR #794, opened 2026-08-27
refactor(monitor): `--live` asks the program it launched, `--attach` reads it — neither is macOS-only now

Guikingone:
`--live` was refused on anything but macOS, and the reason given was true and beside the point: reading a process from the outside needs `/usr/bin/sample`, which ships there alone — but **`--live` launches the program**. It never had to read it from outside. It simply never asked.

The exact path has asked since it existed: it hands the child a socketpair before the fork, and that channel is the credential — only this process holds the other end, so there is nothing to copy, find in a log, or replay. The live path built its child with a bare `spawn()`, so the program was never told it was being monitored, its probe stayed dormant, and an external sampler became the only way to see anything.

**The platform limit was a consequence of a missing link, not a property of Linux.**

### What it needed

Nothing new measures anything. The probe already samples itself into a ring and already renders the folded profile the endpoint serves, so `sampled_answer()` is what goes over the channel. A thread serves it, started only when the channel is present — a program nobody asked about spawns nothing — and blocks SIGPIPE and SIGPROF for the two reasons the endpoint thread does.

Snapshots are cumulative, like the endpoint's answer; a window is the difference between two, computed by the caller. A probe that reset on every read would have to decide what a window is, and would lose one to any reader that disconnected.

`--attach` cannot be fixed the same way — it is handed a pid already running under someone else's control, with no channel in, so reading it from the outside is not a shortcut there, it is the whole job. That half is [below](#attach).

### Three defects of my own, all found by running it

- **The parent kept its copy of the CHILD end.** While it holds one, the socket has a writer that never writes — so when the program exits, `recv` does not report EOF, it waits for us forever. The live loop hung there instead of noticing the target was gone.
- **Asking also wakes the exact profiler**, which writes its table to stderr at exit. The live view printed that raw dump under the table, as if the program had written it. A live view has to filter as the lines arrive, since it runs as long as the program does and a pipe nobody reads fills and blocks the writer.
- **The snapshot was collected into a map**, which keeps the *last* of two equal stacks rather than their sum. `folded_text_to_display` emits the same display stack twice when a folded line carries a native leaf, so samples went missing from the window and from every window after it.

`EINTR` is handled on both sides: a signal arriving mid-wait is not the monitor leaving, and treating it as one would stop serving snapshots silently — which the other end reads as "the target is gone".

### Verification

`test_cli_monitor_live_needs_no_external_sampler` is deliberately **not** gated on macOS, which is the point of it. It asserts a second window appears, that a PHP function is named (so frames are symbolised rather than an empty table anyone could produce), that the program's own output survives, and that no `elephc-probe:`/`elephc-instr:` line reaches the operator.

The control that settles whether the new transport is faithful: the **endpoint** path, untouched by this and reading the same `sampled_answer()`, reports the same profile for the same binary.

Four mutations over the whole path — channel never opened, probe never serving, parent keeping the child end, stderr unfiltered — **all four red, each for a different one of the test's four assertions**.

`elephc-probe` 60 · `elephc` lib 1488 · bins 1659 · `error_tests` 1437 · codegen cli 32 · the `--doc` gate. No warnings.

### <a name="attach"></a>The other half: `--attach` on Linux

`--attach` was then the last thing in `monitor` a platform could refuse, and it could not be answered by asking, so elephc reads the process itself: seize each thread, stop it, read `pc` and `fp`, walk the frame chain, resume. The syscalls and the loop driving them are alone in `ptrace.rs`, because they are the one part no test on a macOS host can run; the ELF symbol table, the naming and the folding were separated for that reason and are exercised everywhere.

Four more defects, none of which announced itself, and two of which produced a **wrong answer** rather than an error:

- **`function_symbols` returned the linker's order and `symbolize` binary-searches.** The parameter is named `sorted` and nothing had ever called both. An unsorted binary search does not fail — it answers `None` for nearly every address, and a 12-second spin loop came back as 251 samples of `<native>`, which reads as *this program is not PHP*. The producer now guarantees the order its only consumer requires, asserted where it is produced: a test that sorted first would pass against the bug.
- **`display_stack` published the mangled spelling.** Ordered, not cosmetic: the prefix that says PHP-or-helper is exactly what demangling removes, so classifying after it leaves every PHP function native.
- **The worker tree was one pid.** I shipped `pids.first()` with a comment explaining why one was enough; it was the limitation restated. Workers share the image but not the bias, so each derives its own — reusing the root's names the wrong functions, which is worse than a short table.
- **A detach that never took.** `PTRACE_DETACH` needs a *stopped* tracee and restarts it as it detaches; asked of a running one it fails with `ESRCH` silently. Seizes are per window and a traced thread cannot be seized twice, so window two seized nothing — and an empty window is how `--attach` learns its target is gone. The live view announced the program had ended, eight seconds early. A related one: `interrupt_and_wait` reported one failure for two things, so a failed wait returned before any resume and left the thread stopped.

Reading from the outside asks two things of a target that asking does not, and each refusal names its remedy: a stripped binary says to rebuild with `--keep-symbols` (elephc strips by default, because nothing *inside* a program reads its symbols), and a kernel that forbids tracing names the `yama/ptrace_scope` setting.

### I can now run the Linux path, and did

The earlier version of this said I could only type-check for Linux and that CI's shards were the oracle. That is no longer true: the attach half is exercised against real running processes in a Linux container (`--cap-add=SYS_PTRACE`).

| exercised on Linux | result |
|---|---|
| one-shot capture | `{main}` → `outer_driver` → `inner_hot_loop`, 74.4% |
| worker tree | 672 samples over 2 processes at different bases, both named |
| `--attach --live` | full 8s, 4 windows, ends when the target exits |
| stack depth | a 9-frame chain naming all five levels |
| `--out` / `--html` / `--dot` | all three written, all naming the function |
| absent pid · stripped binary | `exit=1`, each saying which |
| threads left stopped | none; `state=R` throughout, utime climbing |

The two ordering fixes are mutation-verified. The detach and resume fixes are measured before and after instead, since no unit test on this host reaches ptrace: before, the live view exited at 2s having seized 0 of 1 target in its second window with `TracerPid` still set; after, the full duration.

`elephc` bins 1678 · lib 1488 · codegen cli 37 · `elephc-probe` 61 · `elephc-instr` 79; clippy and `fmt` clean. On Linux: `monitor::` 61, clippy clean on `src/monitor`. CI green across 127 jobs.

### Docs

`--attach` was called macOS-only in four places, one of which also still said `--live` reads a process from the outside — untrue as of the first half of this PR.

### DIFF
--- a/crates/elephc-probe/src/lib.rs
+++ b/crates/elephc-probe/src/lib.rs
@@ -1107,6 +1107,18 @@ pub unsafe extern "C" fn elephc_probe_init(table: *const SymtabEntry, len: usize
         flag.write(1);
         ASKED.store(true, Ordering::Relaxed);
         arm_timer();
+        // The same channel that said "you are being monitored" can also answer
+        // "what have you sampled so far" — but only for a monitor that said it
+        // would ask. The exact path spawns, waits for the program to finish and
+        // reads its output; a thread parked in `recv` for the whole of that run
+        // is one nobody wanted, and one more thing between the program and its
+        // exit.
+        if POLLED.load(Ordering::Relaxed) {
+            std::thread::Builder::new()
+                .name("elephc-probe-control".to_string())
+                .spawn(serve_control_channel)
+                .ok();
+        }
     }
 
     // fork() RESETS interval timers in the child (POSIX; `man 2 fork`), so a
@@ -1124,7 +1136,7 @@ pub unsafe extern "C" fn elephc_probe_init(table: *const SymtabEntry, len: usize
     // The remote endpoint is opt-in: a Unix socket path in ELEPHC_PROBE_ADDR
     // turns it on. A background thread accepts connections, runs the build-key
     // handshake, and serves the folded profile — so a live production process
-    // can be profiled by `elephc monitor --probe-host` without SIGPROF from
+    // can be profiled by `elephc monitor <addr>` without SIGPROF from
     // outside and without suspending the process.
     if !key.is_null() {
         if let Ok(path) = std::env::var("ELEPHC_PROBE_ADDR") {
@@ -1133,6 +1145,119 @@ pub unsafe extern "C" fn elephc_probe_init(table: *const SymtabEntry, len: usize
             }
         }
     }
+}
+
+/// Request byte the monitor sends on the control channel to ask for a snapshot
+/// of what has been sampled so far.
+///
+/// One byte, because the channel carries exactly one question. Anything else is
+/// ignored rather than answered: fd 3 is an ordinary number and this thread must
+/// never invent a reply to a protocol it does not own.
+const CONTROL_SNAPSHOT_REQUEST: u8 = b'S';
+
+/// Serves sampled snapshots to the process that launched us, over the control
+/// channel it already holds the other end of.
+///
+/// The channel IS the credential. A socketpair created before the fork cannot be
+/// opened, guessed, or replayed by anything else on the machine, which is why
+/// this path needs no key handshake where the network endpoint does — the same
+/// reasoning that lets init treat its magic byte as "you are being monitored".
+///
+/// What this buys is `--live` without an external sampler. Reading a process
+/// from the outside needs a tool that ships on macOS alone, so the live table
+/// was macOS-only — for a program `monitor` had launched ITSELF, and could
+/// therefore simply have asked. Answering here is the ask.
+///
+/// Snapshots are cumulative, exactly as the endpoint's answer is; a caller that
+/// wants one window subtracts two of them. Keeping the accumulation on this side
+/// would mean the probe deciding what a window is, which is the caller's
+/// question, and would lose a sample to every reader that ever disconnected.
+fn serve_control_channel() {
+    // Both for the reasons the endpoint thread blocks them. SIGPIPE: a monitor
+    // that exits mid-write must not take the profiled process down with it.
+    // SIGPROF: `ITIMER_PROF` is delivered to the PROCESS and the kernel picks
+    // any thread not blocking it, so the sampler would otherwise interrupt this
+    // thread and fill the ring with the profiler's own frames instead of the
+    // ones running PHP.
+    unsafe {
+        let mut set: libc::sigset_t = std::mem::zeroed();
+        libc::sigemptyset(&mut set);
+        libc::sigaddset(&mut set, libc::SIGPIPE);
+        libc::sigaddset(&mut set, libc::SIGPROF);
+        libc::pthread_sigmask(libc::SIG_BLOCK, &set, std::ptr::null_mut());
+    }
+    loop {
+        let mut request = [0u8; 1];
+        // Safety: CONTROL_FD is the socketpair end init identified; this thread
+        // is its only reader.
+        let read = unsafe {
+            libc::recv(
+                CONTROL_FD,
+                request.as_mut_ptr() as *mut libc::c_void,
+                1,
+                0,
+            )
+        };
+        if read < 0 {
+            // A signal that arrived mid-wait is not the monitor leaving. Treating
+            // `EINTR` as the end would stop serving snapshots silently, and the
+            // live loop on the other side reads that as "the target is gone" and
+            // stops with the program still running.
+            if std::io::Error::last_os_error().kind() == std::io::ErrorKind::Interrupted {
+                continue;
+            }
+            return;
+        }
+        if read == 0 {
+            // A real EOF: the monitor is gone, which is the end of the reason
+            // this thread exists.
+            return;
+        }
+        if request[0] != CONTROL_SNAPSHOT_REQUEST {
+            continue;
+        }
+        let profile = sampled_answer();
+        let bytes = profile.as_bytes();
+        let header = (bytes.len() as u32).to_le_bytes();
+        if !control_send_all(&header) || !control_send_all(bytes) {
+            return;
+        }
+    }
+}
+
+/// Writes every byte of `data` to the control channel, or reports that it could
+/// not.
+///
+/// A short write on a stream socket is ordinary, not an error, and a reply that
+/// stops halfway is worse than no reply: the reader is length-prefixed and would
+/// block waiting for a body that is never coming.
+fn control_send_all(data: &[u8]) -> bool {
+    let mut sent = 0usize;
+    while sent < data.len() {
+        // Safety: writing `data`'s own bytes to a socket this thread owns.
+        let wrote = unsafe {
+            libc::send(
+                CONTROL_FD,
+                data[sent..].as_ptr() as *const libc::c_void,
+                data.len() - sent,
+                0,
+            )
+        };
+        if wrote < 0 {
+            // Same reasoning as the read side: an interrupted write has sent
+            // nothing and can simply be retried, where reporting failure would
+            // truncate a reply the reader is length-prefixed to wait for.
+            if std::io::Error::last_os_error().kind() == std::io::ErrorKind::Interrupted {
+                continue;
+            }
+            return false;
+        }
+        if wrote == 0 {
+            return false;
+        }
+        sent += wrote as usize;
+    }
+    true
 }
 
 /// Starts sampled collection, because someone authenticated and asked for it.
@@ -1403,8 +1528,29 @@ const CONTROL_FD: i32 = 3;
 /// buffered when the child looks. A stray inherited socket on the same descriptor
 /// says nothing and is ignored.
 const CONTROL_MAGIC: &[u8] = b"ELEPHC-MONITOR-1";
+/// The same marker, from a monitor that intends to POLL this process for
+/// snapshots rather than read its output at the end.
+///
+/// Same length as the plain one, so the peek that decides whether fd 3 is ours
+/// is unchanged — it reads a fixed sixteen bytes and compares. Written before
+/// the fork like the other, which is what makes this free of a race: the child
+/// cannot look before the parent has decided.
+///
+/// The distinction earns its keep by NOT starting a server for a run that will
+/// never ask. `--live` polls; the exact path spawns, waits for the program to
+/// end and reads its output, and a thread parked in `recv` for the whole of that
+/// is a thread nobody wanted.
+const CONTROL_MAGIC_LIVE: &[u8] = b"ELEPHC-MONITOR-L";
 /// Returned to the spawning monitor only after the marker was consumed.
 const CONTROL_ACK: &[u8] = b"ELEPHC-MONITOR-ACK-1";
+
+/// Whether the monitor that started this process said it would poll it.
+///
+/// Set by the marker check, read by init. A plain word rather than a return
+/// value because `control_fd_present` answers a yes/no question that three
+/// callers already depend on, and widening it would make every one of them
+/// carry a distinction only one of them uses.
+static POLLED: AtomicBool = AtomicBool::new(false);
 
 /// Whether this process was started by `elephc monitor`.
 ///
@@ -1447,9 +1593,11 @@ fn control_fd_present() -> bool {
             buf.len(),
             libc::MSG_PEEK | libc::MSG_DONTWAIT,
         );
-        if read != CONTROL_MAGIC.len() as isize || buf != CONTROL_MAGIC {
+        let polled = buf == CONTROL_MAGIC_LIVE;
+        if read != CONTROL_MAGIC.len() as isize || (buf != CONTROL_MAGIC && !polled) {
             return false;
         }
+        POLLED.store(polled, Ordering::Relaxed);
         // It is ours: consume the marker so nothing downstream reads it back.
         libc::recv(
             CONTROL_FD,
@@ -1732,9 +1880,73 @@ fn decode_hex(text: &str) -> Option<Vec<u8>> {
     Some(out)
 }
 
+/// How far the fold has already carried, as a ticket number.
+///
+/// The ring's head only grows, so this is a watermark: everything below it has
+/// been taken out of the buffer and added to `CARRIED`, and re-reading it would
+/// count the same samples twice.
+static CARRIED_THROUGH: AtomicU64 = AtomicU64::new(0);
+
+/// Everything carried out of the ring so far.
+///
+/// Touched only by a reader — the control thread answering a snapshot, or the
+/// exit-time dump — and never by the signal handler, which is what makes a lock
+/// safe here at all.
+static CARRIED: std::sync::Mutex<Carried> = std::sync::Mutex::new(Carried::new());
+
+/// The cumulative half of the answer: stacks, allocation weights, and the count
+/// of samples that produced a walkable stack.
+struct Carried {
+    folded: std::collections::BTreeMap<Vec<String>, u64>,
+    allocated: std::collections::BTreeMap<Vec<String>, u64>,
+    valid: u64,
+}
+
+impl Carried {
+    const fn new() -> Self {
+        Self {
+            folded: std::collections::BTreeMap::new(),
+            allocated: std::collections::BTreeMap::new(),
+            valid: 0,
+        }
+    }
+}
+
+/// Adds one fold's worth of ring to what has been carried, and answers the
+/// running total.
+///
+/// A poisoned lock answers the fresh fold rather than panicking: a profiler that
+/// takes the program down because its own bookkeeping tripped is worse than one
+/// that reports a short window.
+fn carry_forward(
+    taken: u64,
+    folded: std::collections::BTreeMap<Vec<String>, u64>,
+    allocated: std::collections::BTreeMap<Vec<String>, u64>,
+    valid: u64,
+) -> (
+    std::collections::BTreeMap<Vec<String>, u64>,
+    std::collections::BTreeMap<Vec<String>, u64>,
+    u64,
+) {
+    let Ok(mut carried) = CARRIED.lock() else {
+        return (folded, allocated, valid);
+    };
+    for (stack, count) in folded {
+        *carried.folded.entry(stack).or_default() += count;
+    }
+    for (stack, allocs) in allocated {
+        *carried.allocated.entry(stack).or_default() += allocs;
+    }
+    carried.valid += valid;
+    CARRIED_THROUGH.store(taken, Ordering::Relaxed);
+    (carried.folded.clone(), carried.allocated.clone(), carried.valid)
+}
+
 /// Renders the current folded profile for the endpoint responder.
 pub fn current_folded_profile() -> Option<String> {
-    unsafe { folded_profile() }
+    // Cumulative: this is the answer `monitor --live` subtracts one window from
+    // the previous, and the endpoint's own contract says the same.
+    unsafe { folded_profile_with(true) }
 }
 
 /// How long the first sampled answer waits for collection to produce something.
@@ -1829,6 +2041,24 @@ pub unsafe extern "C" fn elephc_probe_dump() {
 /// a concurrent handler may add a sample mid-read (endpoint) — a raced slot
 /// only skews one count, never corrupts memory.
 unsafe fn folded_profile() -> Option<String> {
+    folded_profile_with(false)
+}
+
+/// The same fold, optionally CUMULATIVE across calls.
+///
+/// `cumulative` is what the snapshot answer needs and the exit dump does not.
+/// The dump reads once, with the timer disarmed, and "everything still in the
+/// ring" is exactly right for it; a reader that asks every window needs the
+/// total, because the ring stops growing the moment its head wraps and a
+/// consumer subtracting one answer from the last then sees nothing at all.
+///
+/// Kept as a parameter rather than made unconditional because a fold that
+/// carries is not idempotent, and both the dump and the tests fold more than
+/// once expecting the same answer twice.
+///
+/// # Safety
+/// As `folded_profile`.
+unsafe fn folded_profile_with(cumulative: bool) -> Option<String> {
     let table = TABLE_PTR.load(Ordering::Relaxed) as *const SymtabEntry;
     let table_len = TABLE_LEN.load(Ordering::Relaxed);
     if table.is_null() || table_len == 0 {
@@ -1850,8 +2080,26 @@ unsafe fn folded_profile() -> Option<String> {
 
     let head = region_head()?;
     let base = REGION.load(Ordering::Relaxed);
-    let taken = head.load(Ordering::Relaxed) as usize;
-    let available = taken.min(RING_SLOTS);
+    let taken = head.load(Ordering::Relaxed);
+    // Fold by TICKET, not by slot, and only the tickets not folded already.
+    //
+    // The ring holds the last `RING_SLOTS` samples, so folding all of it answers
+    // "the recent window" — which stops growing the moment the head wraps. The
+    // answer is documented as CUMULATIVE and `monitor --live` subtracts one from
+    // the previous on that basis, so past 8192 samples every delta went zero or
+    // negative and the live table emptied while the program ran flat out. What
+    // is cumulative is not the ring; it is what has been carried out of it.
+    //
+    // A slot caught mid-write is skipped, as it always was, and the watermark
+    // still advances past it: one sample lost rather than the same slot counted
+    // again on the next read. The reader that falls more than a full ring behind
+    // loses what was overwritten, which no fixed ring can avoid.
+    let oldest_live = taken.saturating_sub(RING_SLOTS as u64);
+    let from = if cumulative {
+        CARRIED_THROUGH.load(Ordering::Relaxed).max(oldest_live)
+    } else {
+        oldest_live
+    };
     let mut valid = 0u64;
     let mut folded: std::collections::BTreeMap<Vec<String>, u64> = std::collections::BTreeMap::new();
     // Allocations charged to each stack, kept apart from the sample counts: they
@@ -1859,7 +2107,8 @@ unsafe fn folded_profile() -> Option<String> {
     // profile whose bars mean two things at once.
     let mut allocated: std::collections::BTreeMap<Vec<String>, u64> =
         std::collections::BTreeMap::new();
-    for index in 0..available {
+    for ticket in from..taken {
+        let index = (ticket % RING_SLOTS as u64) as usize;
         // Acquire the depth gate before reading the PCs the handler stored; a
         // torn or in-flight slot with depth 0 is skipped.
         // Read the sequence first: odd means a handler is inside this slot right
@@ -1916,6 +2165,14 @@ unsafe fn folded_profile() -> Option<String> {
         }
         *folded.entry(stack).or_default() += 1;
     }
+    // Carried out of the ring before anything is rendered, so what the answer
+    // reports is every sample since the process started and not merely the ones
+    // still in the buffer.
+    let (folded, allocated, valid) = if cumulative {
+        carry_forward(taken, folded, allocated, valid)
+    } else {
+        (folded, allocated, valid)
+    };
     // A dormant binary took no samples and recorded no events. Saying
     // "elephc-probe-samples: 0" would still announce a profiler to anyone
     // reading the program's own stderr, which is exactly what
@@ -2508,6 +2765,116 @@ mod tests {
             !mid_write.as_deref().unwrap_or_default().contains(name),
             "a slot a handler is inside must not fold, however readable it looks: {mid_write:?}"
         );
+    }
+
+    /// The cumulative answer keeps growing after the ring's head has lapped.
+    ///
+    /// The ring holds the last `RING_SLOTS` samples, so folding all of it
+    /// answers "the recent window" — and `monitor --live` subtracts one answer
+    /// from the previous on the documented promise that it is CUMULATIVE. Past
+    /// 8192 samples the count stopped growing, so every later delta was zero or
+    /// negative and the live table emptied while the program ran flat out. The
+    /// CLI fixtures run six seconds at about a kilohertz and never reach the
+    /// lap, which is why nothing saw it.
+    ///
+    /// Staged rather than run for real: a fixture long enough to take 8192
+    /// samples is eight seconds of CPU in a unit test, and what is under test is
+    /// the arithmetic across the lap, not the sampler.
+    #[test]
+    fn a_lapped_ring_still_answers_a_growing_total() {
+        let _serial = ROUTE_TEST_LOCK.lock().unwrap_or_else(|e| e.into_inner());
+        let name = "hot\0";
+        let end = "<end>\0";
+        let symbols = [
+            SymtabEntry {
+                address: 0x1000,
+                name_ptr: name.as_ptr() as u64,
+                name_len: name.len() as u64,
+            },
+            SymtabEntry {
+                address: 0x2000,
+                name_ptr: end.as_ptr() as u64,
+                name_len: end.len() as u64,
+            },
+        ];
+        let mut region = vec![0u8; REGION_BYTES];
+        let base = region.as_mut_ptr() as usize;
+
+        let saved_region = REGION.swap(base, Ordering::Relaxed);
+        let saved_ptr = TABLE_PTR.swap(symbols.as_ptr() as usize, Ordering::Relaxed);
+        let saved_len = TABLE_LEN.swap(symbols.len(), Ordering::Relaxed);
+        let saved_stopped = STOPPED.swap(false, Ordering::Relaxed);
+        reset_carried();
+
+        let answers = unsafe {
+            let head = region_head().expect("mapped");
+
+            // One sample, taken as ticket 0 and still in the ring.
+            head.store(1, Ordering::Relaxed);
+            region_word(base, 0, 0).store(1, Ordering::Release);
+            region_word(base, 0, PC_WORD0).store(0x1000, Ordering::Relaxed);
+            region_word(base, 0, SEQ_WORD).store(slot_seq_settled(0), Ordering::Release);
+            let first = folded_profile_with(true);
+
+            // Asked AGAIN with nothing new taken. The total must not move: a
+            // sample still sitting in the ring has already been carried, and a
+            // fold that re-reads it counts it twice — which grows the total for
+            // the wrong reason and would hide the lap defect from this test.
+            let asked_twice = folded_profile_with(true);
+
+            // The head has now lapped: `RING_SLOTS` further samples were taken
+            // and the newest of them landed back in slot 0, overwriting the one
+            // above. All the reader can still SEE is one sample; the total it
+            // must report is two.
+            let lapped = RING_SLOTS as u64 + 1;
+            head.store(lapped, Ordering::Relaxed);
+            region_word(base, 0, 0).store(1, Ordering::Release);
+            region_word(base, 0, PC_WORD0).store(0x1000, Ordering::Relaxed);
+            region_word(base, 0, SEQ_WORD).store(slot_seq_settled(lapped - 1), Ordering::Release);
+            let after_lap = folded_profile_with(true);
+            (first, asked_twice, after_lap)
+        };
+
+        REGION.store(saved_region, Ordering::Relaxed);
+        TABLE_PTR.store(saved_ptr, Ordering::Relaxed);
+        TABLE_LEN.store(saved_len, Ordering::Relaxed);
+        STOPPED.store(saved_stopped, Ordering::Relaxed);
+        reset_carried();
+
+        let (first, asked_twice, after_lap) = answers;
+        let first = first.unwrap_or_default();
+        let asked_twice = asked_twice.unwrap_or_default();
+        let after_lap = after_lap.unwrap_or_default();
+        assert!(
+            first.contains("elephc-probe-samples: 1"),
+            "the first answer counts the one sample taken: {first}"
+        );
+        assert!(
+            asked_twice.contains("elephc-probe-samples: 1"),
+            "asking twice must not count the same sample twice: {asked_twice}"
+        );
+        assert!(
+            after_lap.contains("elephc-probe-samples: 2"),
+            "a lapped ring must still answer a GROWING total; a count that stops at what the \
+             buffer still holds is what emptied the live table: {after_lap}"
+        );
+        // Matched on the tail rather than on `"hot 2"`: the fixture's symbol
+        // names carry their C terminator, so the folded line reads `hot\0 2`.
+        assert!(
+            after_lap
+                .lines()
+                .any(|line| line.starts_with("elephc-probe: ") && line.ends_with(" 2")),
+            "and the per-stack weight has to grow with it: {after_lap}"
+        );
+    }
+
+    /// Clears what previous folds carried, so one test's samples are not another
+    /// test's total. The accumulator is process-wide by design.
+    fn reset_carried() {
+        CARRIED_THROUGH.store(0, Ordering::Relaxed);
+        if let Ok(mut carried) = CARRIED.lock() {
+            *carried = Carried::new();
+        }
     }
 
     /// Two writers a lap apart cannot agree on a settled value.
--- a/crates/elephc-probe/src/lib.rs
+++ b/crates/elephc-probe/src/lib.rs
@@ -1114,10 +1114,11 @@ pub unsafe extern "C" fn elephc_probe_init(table: *const SymtabEntry, len: usize
         // is one nobody wanted, and one more thing between the program and its
         // exit.
         if POLLED.load(Ordering::Relaxed) {
-            std::thread::Builder::new()
-                .name("elephc-probe-control".to_string())
-                .spawn(serve_control_channel)
-                .ok();
+            start_control_channel(CONTROL_FD, || {
+                std::thread::Builder::new()
+                    .name("elephc-probe-control".to_string())
+                    .spawn(serve_control_channel)
+            });
         }
     }
 
@@ -1143,6 +1144,25 @@ pub unsafe extern "C" fn elephc_probe_init(table: *const SymtabEntry, len: usize
             if !path.is_empty() {
                 endpoint::spawn(path);
             }
+        }
+    }
+}
+
+/// Starts the control server, disconnecting the authenticated channel on failure.
+/// A receive timeout cannot distinguish a slow server from one that never started.
+/// Closing the unserved peer lets the monitor report failure and leave PHP running.
+/// The spawn operation is injected so resource exhaustion can be tested without
+/// exhausting the application's own thread or memory limits.
+fn start_control_channel(
+    fd: libc::c_int,
+    spawn: impl FnOnce() -> std::io::Result<std::thread::JoinHandle<()>>,
+) {
+    if spawn().is_err() {
+        // SAFETY: init owns this authenticated socket and no server was created.
+        // Shutdown also disconnects any inherited duplicates of the peer.
+        unsafe {
+            libc::shutdown(fd, libc::SHUT_RDWR);
+            libc::close(fd);
         }
     }
 }
@@ -2863,6 +2883,51 @@ mod tests {
                 .any(|line| line.starts_with("elephc-probe: ") && line.ends_with(" 2")),
             "and the per-stack weight has to grow with it: {after_lap}"
         );
+    }
+
+    /// A refused thread spawn must expose EOF, not an unserved socket that times out.
+    #[test]
+    fn failed_control_thread_start_disconnects_the_monitor() {
+        use std::io::{Read, Write};
+        use std::os::fd::IntoRawFd;
+        use std::os::unix::net::UnixStream;
+
+        let (mut monitor, mut server) = UnixStream::pair().unwrap();
+        monitor.set_read_timeout(Some(std::time::Duration::from_millis(200))).unwrap();
+        server.write_all(&CONTROL_ACK).unwrap();
+        start_control_channel(server.into_raw_fd(), || {
+            Err(std::io::Error::from_raw_os_error(libc::EAGAIN))
+        });
+        let mut ack = [0; CONTROL_ACK.len()];
+        monitor.read_exact(&mut ack).unwrap();
+        assert_eq!(ack, CONTROL_ACK);
+        let mut byte = [0];
+        assert_eq!(monitor.read(&mut byte).expect("startup failure must not become a timeout"), 0);
+    }
+
+    /// Successful startup retains the peer so the server can answer its monitor.
+    #[test]
+    fn successful_control_thread_start_keeps_the_channel_open() {
+        use std::io::{Read, Write};
+        use std::os::fd::{FromRawFd, IntoRawFd};
+        use std::os::unix::net::UnixStream;
+
+        let (mut monitor, server) = UnixStream::pair().unwrap();
+        monitor.set_read_timeout(Some(std::time::Duration::from_secs(2))).unwrap();
+        let fd = server.into_raw_fd();
+        start_control_channel(fd, || {
+            std::thread::Builder::new().spawn(move || {
+                // SAFETY: ownership transfers only after the server thread starts.
+                let mut server = unsafe { UnixStream::from_raw_fd(fd) };
+                let mut request = [0];
+                server.read_exact(&mut request).unwrap();
+                server.write_all(&request).unwrap();
+            })
+        });
+        monitor.write_all(b"S").unwrap();
+        let mut answer = [0];
+        monitor.read_exact(&mut answer).unwrap();
+        assert_eq!(&answer, b"S");
     }
 
     /// Concurrent endpoint/control readers must count each settled ticket once,
--- a/crates/elephc-probe/src/lib.rs
+++ b/crates/elephc-probe/src/lib.rs
@@ -1880,18 +1880,10 @@ fn decode_hex(text: &str) -> Option<Vec<u8>> {
     Some(out)
 }
 
-/// How far the fold has already carried, as a ticket number.
-///
-/// The ring's head only grows, so this is a watermark: everything below it has
-/// been taken out of the buffer and added to `CARRIED`, and re-reading it would
-/// count the same samples twice.
-static CARRIED_THROUGH: AtomicU64 = AtomicU64::new(0);
-
 /// Everything carried out of the ring so far.
 ///
-/// Touched only by a reader — the control thread answering a snapshot, or the
-/// exit-time dump — and never by the signal handler, which is what makes a lock
-/// safe here at all.
+/// Shared by control-channel and endpoint readers, never by the signal handler.
+/// The lock covers ticket selection, folding and publication as one transaction.
 static CARRIED: std::sync::Mutex<Carried> = std::sync::Mutex::new(Carried::new());
 
 /// The cumulative half of the answer: stacks, allocation weights, and the count
@@ -1900,14 +1892,18 @@ struct Carried {
     folded: std::collections::BTreeMap<Vec<String>, u64>,
     allocated: std::collections::BTreeMap<Vec<String>, u64>,
     valid: u64,
+    /// First ticket not yet included in these totals; protected by the same lock.
+    through: u64,
 }
 
 impl Carried {
+    /// Creates an empty accumulator before any ring tickets have been consumed.
     const fn new() -> Self {
         Self {
             folded: std::collections::BTreeMap::new(),
             allocated: std::collections::BTreeMap::new(),
             valid: 0,
+            through: 0,
         }
     }
 }
@@ -1915,10 +1911,9 @@ impl Carried {
 /// Adds one fold's worth of ring to what has been carried, and answers the
 /// running total.
 ///
-/// A poisoned lock answers the fresh fold rather than panicking: a profiler that
-/// takes the program down because its own bookkeeping tripped is worse than one
-/// that reports a short window.
+/// The caller holds the accumulator lock from ticket selection through publication.
 fn carry_forward(
+    carried: &mut Carried,
     taken: u64,
     folded: std::collections::BTreeMap<Vec<String>, u64>,
     allocated: std::collections::BTreeMap<Vec<String>, u64>,
@@ -1928,9 +1923,6 @@ fn carry_forward(
     std::collections::BTreeMap<Vec<String>, u64>,
     u64,
 ) {
-    let Ok(mut carried) = CARRIED.lock() else {
-        return (folded, allocated, valid);
-    };
     for (stack, count) in folded {
         *carried.folded.entry(stack).or_default() += count;
     }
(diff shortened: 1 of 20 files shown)


Guikingone:
### Folded in #793 — same sector, one PR

`650c913` (the docs audit) is now the second commit here, since both change `docs/beyond-php/profiling.md` and reviewing them apart would mean reading that page twice. #793 is closed in favour of this.

What it carries:

**Four behaviour changes that merged with #790 and left no user-facing record.** Its CHANGELOG has one entry for the coroutine work — the first of fourteen commits — and the four after it changed behaviour with nothing written down: `yield from` leaving the delegating generator on the profiler's stack (52.6% of a program's inclusive time for 52 µs of its own work), the three non-returning suspend paths reaching PHP handlers with the activation still parked, a catch handler that yields sweeping its exception's dead frames into the park, and a caught exception read as whatever its variable held before the `try` — which stops the compile rather than answering wrongly.

**Two pages describing the old behaviour.** The 4,096-coroutine cap was presented as the whole story when a suspension also releases what was parked under its own coroutine; and the timeline section said each call becomes one nested slice, which a coroutine has not done since a suspension started ending a span.

**Two counts that had drifted** in `docs/internals/architecture.md`: `runtime/strings/` is 97 files, not 95, and `src/codegen/shared_*.rs` is 4, not 3 — `shared_state.rs` was in neither the count nor the description.

It also records what was checked and was already right, so the next pass starts further along: across all 1,130 pages, frontmatter is complete, no page opens with an `#` Astro already renders, no relative link is broken, every `src/` path cited exists, every one of the 39 flags `src/cli.rs` parses is documented, and the phase list matches every `record_since` label in `src/pipeline.rs` in order.

The mechanical checks were re-run after the merge: 1,130 files, 0 frontmatter problems, 0 stray headings, 0 broken links.

src/monitor/channel.rs
```diff
@@ -97,6 +97,94 @@ pub(crate) fn open_control_channel() -> Option<ControlChannel> {
...
+/// `None` on any failure — a child too young to have a thread reading yet, one
+/// that exited, a short read. The live loop treats it as an empty window, which
+/// is what an unsampled window is.
+pub(crate) fn request_snapshot(channel: &ControlChannel) -> Option<String> {
+    let fd = channel.parent;
+    let request = [CONTROL_SNAPSHOT_REQUEST];
+    // Safety: `fd` is this process's end of the socketpair, owned by `channel`.
+    let sent = unsafe {
+        libc::send(fd, request.as_ptr() as *const libc::c_void, 1, 0)
+    };
```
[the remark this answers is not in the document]
Guikingone: Checked rather than applied, and the check says this one does not happen here.

Rust's std sets `SIGPIPE` to `SIG_IGN` at startup, so the write returns `EPIPE` instead of killing the process. Measured, on this machine, with the socket's peer closed first:

```
send returned -1, errno = Broken pipe (os error 32)
REACHED THIS LINE — the process was not killed by SIGPIPE
SIGPIPE disposition = SIG_IGN (ignored)
```

`request_snapshot` checks `sent != 1` and returns `None`, and the live loop reads that as "the child is gone" and ends — which is the intended behaviour when the child exits during a window, and the reason the final report still prints.

The distinction is worth keeping straight rather than papering over, because the other half of this protocol *is* exposed: `serve_control_channel` runs inside the compiled program, where SIGPIPE is default-terminate, and that is why it blocks the signal on its own thread rather than relying on a disposition it does not own.

Your other two findings were real and are fixed in `6009f99f`.

crates/elephc-probe/src/lib.rs
```diff
@@ -1041,6 +1041,13 @@ pub unsafe extern "C" fn elephc_probe_init(table: *const SymtabEntry, len: usize
         flag.write(1);
         ASKED.store(true, Ordering::Relaxed);
         arm_timer();
+        // The same channel that said "you are being monitored" can also answer
+        // "what have you sampled so far". Only started when it is present, so a
+        // program nobody asked about spawns no thread.
+        std::thread::Builder::new()
+            .name("elephc-probe-control".to_string())
+            .spawn(serve_control_channel)
+            .ok();
```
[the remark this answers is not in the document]
Guikingone: Confirmed and fixed in `6009f99f`, though not at the place you pointed at — deliberately, and I think for the better.

You are right that a discarded spawn error leaves the descriptor open with nobody behind it, and `recv` on a live peer that never writes does not return. A live view would hang on a program that is still running, which is worse than a missing profile because it gives no reason.

I put a **receive timeout on the monitor's end** rather than a guard at the spawn. The hang is what every cause has in common — a failed spawn, a thread that died later, a child too busy to answer, and whatever else nobody has thought of — and the caller already treats "no answer" as "the target is gone", so the recovery path exists and is exercised. Guarding the one cause you found would have left the others.

Five seconds: generous against the answer itself, which waits out a 400 ms warm-up before reporting an empty ring, and finite against everything else.

src/monitor/mod.rs
```diff
@@ -211,16 +207,55 @@ pub(crate) fn run(cmd: MonitorCommand) -> i32 {
...
+    // writing into it.
+    if let Some(stderr) = child.stderr.take() {
+        std::thread::Builder::new()
+            .name("elephc-monitor-stderr".to_string())
+            .spawn(move || {
+                use std::io::BufRead as _;
+                for line in std::io::BufReader::new(stderr).lines().map_while(Result::ok) {
+                    if line.starts_with("elephc-instr") || line.starts_with("elephc-probe") {
+                        continue;
+                    }
```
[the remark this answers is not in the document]
Guikingone: Confirmed and fixed in `6009f99f`, and it was worse than the live path: the **exact** path has carried the same prefix-only rule since before this branch, spelled out separately. So a program printing `elephc-instrumentation disabled by config` had that line deleted by the tool watching it, on both paths, silently.

The rule is now the shape the profiler actually emits — a first token ending in a colon:

| line | verdict |
|---|---|
| `elephc-instr: hot calls=1 incl_ns=5` | profiler |
| `elephc-instr-edge: a -> b count=1` | profiler |
| `elephc-probe-samples: 934` | profiler |
| `elephc-instrumentation disabled by config` | **the program's** |
| `elephc-probes are not enabled here` | **the program's** |

One function, used by both paths. Being written twice is how they would have drifted apart anyway.

Tested in both directions — six real profiler lines and six a program might write — and verified by mutation: restoring the prefix-only rule turns it red naming `elephc-instrumentation disabled by config`.

src/monitor/channel.rs
```diff
@@ -74,6 +74,41 @@ pub(crate) fn compile_php_monitored(source: &str) -> Result<PathBuf, String> {
...
+    let timeout = libc::timeval { tv_sec: 5, tv_usec: 0 };
+    // Safety: setting a documented option on a socket this process owns.
+    unsafe {
+        libc::setsockopt(
+            channel.parent,
+            libc::SOL_SOCKET,
+            libc::SO_RCVTIMEO,
+            &timeout as *const libc::timeval as *const libc::c_void,
+            std::mem::size_of::<libc::timeval>() as libc::socklen_t,
+        );
```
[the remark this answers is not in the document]
Guikingone: Confirmed, and it is a regression in the fix I made for your previous pass — which is the useful kind of finding to get. Fixed in `a79a0e17`.

`request_snapshot` returned `Option`, so "did not answer in time" and "there is nobody there" arrived as the same `None`. The live loop ends on `None`, and a live view ending is what **reaps** the target — the program only outlives the view while the view is still running. So a busy program that missed one five-second window was stopped, mid-profile, by the tool watching it.

Three outcomes now, because two cannot express it:

- **Answered** — the window.
- **Late** — the deadline passed with *nothing* taken off the stream, so the reply is still to come and the next window can ask again. The loop skips the window and says so once, rather than burying the table under the same line.
- **Gone** — closed, broken, or timed out **part way**. That last one is the subtle member you may not have had in view: the child can be perfectly alive, but half a reply has been consumed, and asking again would read the rest of it as the next answer's length. The channel is finished even though the process is not.

The regression drives a real socketpair with a real receive deadline and nobody answering, because what is under test is what the kernel reports rather than what the code assumes it will. Mutation-verified: collapsing the clean timeout back into the fatal case turns it red on `a target that has not answered YET must not be reported as gone`.

src/monitor/channel.rs
```diff
@@ -113,6 +148,144 @@ pub(crate) fn control_channel_activated(channel: &ControlChannel) -> bool {
...
+/// Matching the prefix alone would swallow a program that writes
+/// `elephc-instrumentation disabled` on its own stderr: a line the author wrote,
+/// removed by the tool that was supposed to be watching. Silently, which is the
+/// part that makes it worth a rule rather than a `starts_with`.
+pub(crate) fn is_profiler_line(line: &str) -> bool {
+    let Some(first) = line.split_whitespace().next() else {
+        return false;
+    };
+    first.ends_with(':') && (first.starts_with("elephc-instr") || first.starts_with("elephc-probe"))
+}
```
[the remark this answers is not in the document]
Guikingone: Fair, and narrowed further in `a79a0e17` — though I want to be straight about what is left rather than claim it is gone.

The rule was "first token starts with `elephc-instr`/`elephc-probe` and ends in a colon", which still takes `elephc-instr-custom:` from a program that chose that name for its own channel. It is now the **nine exact prefixes**, gathered from `elephc-instr` and `elephc-probe`:

```
elephc-instr:            elephc-probe:
elephc-instr-edge:       elephc-probe-alloc:
elephc-instr-query:      elephc-probe-io:
elephc-instr-query-dropped:   elephc-probe-samples:
elephc-instr-trace:
```

What remains is a program that writes one of those nine exactly — a name collision with the tool's reserved output, and not something any text rule can resolve while both share one stream. Removing it for good means giving the profiler its own channel rather than stderr, which is a contract the exact path has depended on since before this branch and is not something to change inside it.

What the list does buy is the direction of failure. The two ways of being wrong are not equal: an unrecognised profiler line reaches the operator as a visible, harmless extra row, while a program's line matched by mistake is deleted with nothing to say it existed. So a new kind of profiler line has to be added to that list, and until it is, it leaks — and that is the side to fail on.

The test covers the nine, plus five a program might write, including the two your comment's shape would still have swallowed.

src/monitor/channel.rs
```diff
@@ -113,6 +148,216 @@ pub(crate) fn control_channel_activated(channel: &ControlChannel) -> bool {
...
+    const MAX_SNAPSHOT: usize = 64 * 1024 * 1024;
+    if len > MAX_SNAPSHOT {
+        return Snapshot::Gone;
+    }
+    let mut body = vec![0u8; len];
+    match recv_exact(fd, &mut body) {
+        Read::Filled => {}
+        // Mid-message, so the channel is desynchronised whatever the child's
+        // state: it cannot be asked again, even though it may be alive.
+        Read::TimedOutClean | Read::Ended => return Snapshot::Gone,
```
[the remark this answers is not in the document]
Guikingone: Confirmed and fixed in `79d9da3b`. You are right, and worse: I named this exact case in the previous commit's own message and did not follow it through.

Being mid-message is a property of the **message**, not of the read in progress. The header is already consumed, so the stream owes exactly `len` bytes — nothing else can be parsed out of it, which is what makes waiting longer safe here and nowhere else. Three deadlines now, each of which already measures a stall with nothing arriving at all.

**The first attempt at this did not work, and the test is how I know.** It retried only once a body byte had landed (`filled > 0` inside `recv_exact`) — but the header is a separate read, so the ordinary shape (header, then a stall before any body byte) starts at `filled == 0` and fell straight through to `Gone`. The retry never fired for the case it was written for.

The test caught it because it was made to cross a real deadline. It did not at first: the writer paused 300 ms against a five-second socket, so it proved "pieces are waited for" while never reaching the path under review. The deadline is settable now and the test asks for one second — what is worth testing is what happens *when* a deadline passes, and a test too slow to keep is a test that gets shortened until it stops reaching it.

Mutation-verified: removing the patience turns it red on `a child sending its reply in pieces was reported as gone, which reaps it`.

One thing I did not change, and want to be explicit about rather than leave implied: ending the view **does** stop a target that is still up. That is what `--live` does with a program it launched — the program exists in order to be watched, and leaving it orphaned when the watch ends would be its own surprise. What it should not do is that silently, so it now says whether the view ended because the program finished or because the channel to it was lost.

Guikingone:
Blocker accepted, and the framing was the useful part: `Gone` means only *the channel broke*, never *the program ended* — a program that finishes normally leaves through `try_wait`. So every path that reached the reap through `Gone` was killing a running target because our own socket went quiet.

Two things changed.

**The answer now carries whose fault it is.** `run_live` returns a `LiveOutcome` rather than a bare code, and a `Gone` that arrives while the child is still alive sets `leave_target_running`. The operator is told which pid is still theirs instead of watching it disappear:

```
elephc monitor: lost the channel to the target; it is still running as pid 4711
and is left alone. Reporting what was collected.
```

**The decision moved somewhere a test can reach it.** This crossed two decisions — whether to kill, and whether to wait — that lived at the call site as an `if` plus the statement after it. Only the first was reachable, which is why the first version of this fix passed its own test while leaving the bug: the early return sat in the caller, not in the rule. They are one function now, over the four facts that decide it:

```rust
fn disposition(still_running: bool, code: i32, live: bool, lost_channel: bool) -> Disposition
```

`LeaveAlone` skips the `wait()` too — waiting is exactly what would hang this process on a program that is still working.

Verified by mutation: removing the `lost_channel` clause turns `only_a_capture_that_owns_the_target_stops_it` red on `disposition(true, 0, true, true)`, the live-view-over-a-live-program case. The test also pins that an already-exited child is still collected, so a lost channel never trades a kill for a zombie.

Suites: `-p elephc --bins` 1665, `--lib` 1488, `codegen::cli` 37, `elephc-probe` 61, `elephc-instr` 79 — all green; clippy clean on `src/monitor`.

Guikingone:
#798 is folded in here and closed — it was this branch plus six commits, so two
PRs were showing the same work twice. Nine commits now, and the two halves of one
change: `--live` stopped being macOS-only by *asking* the program it launched;
`--attach` cannot ask, so it now reads a Linux process itself with `ptrace`
instead of being refused.

The description is rewritten rather than appended to, because the first half
made two claims the second half falsified — that `--attach` keeps its platform
branch, and that I could only type-check the Linux path rather than run it.
Both are corrected in place.

Four further defects came with the second half, and two of them produced a wrong
answer rather than an error: a binary search over an unsorted symbol table
(251 samples of `<native>` from a PHP program), and a detach that never took
(the live view announcing the target gone, eight seconds early). Details and
before/after measurements are in the description and in the comments above.

Guikingone:
Confirmed and fixed in `f689abddc7` — and the finding is sharper than "an edge
case", because the code already knew about this collision and guarded it in the
one place that can fail to.

The comment above `await_activation` spells out the hazard: the reply is
length-prefixed, the ACK's first four bytes are `ELEP`, that decodes as a 1.3 GB
length, the size bound refuses it as `Gone`, and `Gone` ends a live view on a
healthy program. What it did about it was wait two seconds for the ACK first.

A wait that can TIME OUT cannot be the only thing standing between two messages.
A child that boots slower than the deadline sends its ACK straight into the
snapshot read — so the guard meant to prevent the collision is precisely the
thing that misses it.

`request_snapshot` now recognises the ACK in that position, consumes it, and
reads the reply behind it. It treats itself as mid-message while doing so, with
the same stall-and-retry the body loop already uses, so a half-taken ACK is never
left for the next window to parse as a length. That holds whatever the child's
boot takes, which a deadline never could.

One thing your report did not mention that follows from it: the caller now also
stops re-waiting once a snapshot is answered. An answer proves the child is past
init regardless of who consumed the ACK, and without that every later window
would spend its full deadline waiting for an ACK that had already been taken.

The regression test stages the race as it actually happens — both messages queued
before the read begins, nothing having drained the ACK first — because what needs
proving is that the reader copes alone. Mutation-verified: with the new branch
disabled it fails on exactly the path you named, `a queued ACK must not read as a
dead channel`.

Re-measured on Linux afterwards: one-shot attach 253 samples with `inner_hot_loop`
at 74.3% and the target alive after sampling, live attach running its full 14s
across 8 windows.

Guikingone:
Two models audited this branch independently (Kimi K3 and GLM 5.3, given the
diff and the runtime ABIs, nothing else). They converged on the same two
defects, and both are real — I confirmed each by reading the code, not by
counting votes.

**The stack was walked after the thread was let go.** `sample_thread` read the
registers under the stop, called `resume`, and only then called `walk` — which
is not arithmetic on those registers, it reads the target's memory, up to
`MAX_DEPTH` frames. Every one of those reads was racing a thread that had
already been restarted. `can_follow` rejects the gross tears; it does not
reject a stale frame record above a stack that shrank between two reads, which
is aligned and still climbing and so passes both tests while naming a function
that was never on the stack. The module doc claimed "the target is stopped for
the duration of each sample" and `read_memory` justified its cost by how long
it holds the target stopped. Neither was true. Fixed in `e594c75`.

**The resume threw the target's signals away.** A seized tracee reports every
signal to its tracer and waits; `PTRACE_CONT` with `0` does not mean "nothing
was pending", it means "discard it". At 99 Hz that is most of them — a
profiled program silently lost the children it meant to reap and the shutdown
it was asked for. `wait_for_stop` now returns the status it was already
receiving and throwing away, and `pending_signal` separates a stop this tracer
asked for from one the program was about to receive using the `PTRACE_EVENT_*`
bits rather than the signal number: a group-stop under `SEIZE` reports
`SIGSTOP` and the interrupt stop reports `SIGTRAP`, and handing an untraced
`SIGTRAP` back would kill the target on the first tick. Four unit tests pin
that, and they run wherever the file compiles.

**One more, from GLM alone:** `recv_exact` never reset `stalls` when bytes
landed, so the budget accumulated across the whole message. A child that sends
a piece, pauses past the deadline, sends another and pauses again is cut off on
the third pause while it is visibly still writing — the opposite of the
distinction the comment above it draws. `Ended` becomes `Gone`, and `Gone` ends
the live view and reaps the target, so this killed healthy programs for being
slow. `b6aedc9` resets the budget on progress and rewrites
`request_snapshot`'s contract paragraph, which still described a `None`
returned "on any failure" — the shape from before `Snapshot` existed, folding
together the exact two answers the enum was added to separate.

Three findings I did NOT act on, with reasons:

- **`waitpid` has no timeout**, so a thread parked in uninterruptible sleep
  stalls the sampler with no bound. Real, and latent — I have no Linux host
  here to reproduce the D-state, and a `WNOHANG` poll loop is a behaviour
  change I am not willing to ship unmeasured.
- **aarch64 pointer authentication** would leave signing bits in the saved LR.
  Musl aarch64 userspace is not PAC-enabled by default, so this is a question
  about targets we do not currently build for.
- **`PTRACE_INTERRUPT` on an already-stopped tracee**: Kimi read this as a
  freeze until the window ends, GLM as a swallowed signal. They cannot both be
  right, and the fix for the second covers the case that actually occurs. I did
  not guess at the first.

Both models also re-derived the register indices independently — x86_64
`user_regs_struct` rbp=4, rip=16, 27 words; aarch64 `user_pt_regs` x29=29,
pc=32, 34 words — and agreed with what is here. That, and the protocol logic in
`channel.rs`, came back clean from both.

Guikingone:
Valid, and the mechanism turned out to be worse than either of us described. I
measured it rather than reasoned about it, and the first explanation I wrote was
wrong, so here is what a socketpair actually does.

`control_channel_activated` was not reading the ACK, it was **consuming** —
`recv` for twenty bytes with `MSG_DONTWAIT | MSG_WAITALL` and no `MSG_PEEK`.
`MSG_WAITALL` asks the kernel to hold on until all twenty bytes are there, and
**`MSG_DONTWAIT` does not override it on macOS**. Given a queued reply shorter
than twenty bytes — an empty snapshot is a four-byte length word and nothing
else, which is exactly what a child that has sampled nothing yet sends:

| flags | the probe | the reply afterwards |
|---|---|---|
| `DONTWAIT\|WAITALL`, no deadline | **never returns** | — |
| `DONTWAIT\|WAITALL`, 1 s deadline | returns after 1.002 s | **destroyed** |
| `DONTWAIT\|PEEK` | returns in 1.9 µs | intact |

So the path you found is real, and it has three effects rather than one:

1. the child boots slower than the 2 s activation deadline, so `activated` stays
   `false`;
2. `request_snapshot` takes the ACK off the front, then times out on the reply
   header and answers `Late` — correctly, nothing of the reply was consumed;
3. `run_live` sees `!activated` and asks again. The reply is now in the buffer,
   so that question **spends a full 5 s receive deadline** and then **eats the
   reply**. The next length word is parsed out of the middle of a message, which
   answers `Gone`, and `Gone` ends the view and reaps a target that was replying
   perfectly well. On the channel that carries no deadline — the check made once
   the program has finished — the same question never comes back at all.

Fixed by making the probe answer the question instead of spending the answer:
`MSG_PEEK`. That is what its own doc already claimed it did ("asks without
blocking"), and it makes `request_snapshot` the single consumer of the stream —
the one that knows the framing. Recording activation on a `Late` that happened
to consume the ACK would have closed your exact path but not this one: any
question landing on any queued reply had the same effect, and on a channel with
no deadline it hangs whether or not activation was ever recorded.

One consequence worth naming: the ACK is now always still in front of the first
reply, so `request_snapshot`'s `ELEP` branch goes from a path that only ran under
load to the ordinary one. It was already covered by
`a_late_activation_ack_does_not_end_the_snapshot_read`; it is now exercised on
every run.

The regression test queues a reply, asks, and reads the bytes back to prove they
are intact. Two details in it are deliberate. It reads back by raw `recv` rather
than through `request_snapshot`, because the old code does not answer wrongly
there, it blocks on a reply it has already swallowed. And the reply is longer
than twenty bytes, because at four bytes the old code hangs the test instead of
failing it — a test that hangs reports nothing. At this length the bug returns
immediately, twenty bytes poorer, and the read-back says so.
`the_control_ack_proves_the_capture_was_activated` now also asks twice: the
consuming version reported an ACK it had just destroyed, so the second question
answered `false`.

nahime0 (changes requested):
The defects you already found and listed look closed. Two of the same class are still open, and several docs still describe the old macOS-only sampler for `--live`.

## Must-fix

**1. `elephc monitor hot.php --live` never embeds the probe.**

`run` skips `run_instrument` when `cmd.live` is set (`src/monitor/mod.rs`), then compiles a `.php` target with `compile_php_target` (`--debug-info` only). `compile_php_monitored` (`--with-monitoring`) already exists and is what the exact path uses. `require_monitoring` is also skipped for `.php`. Live then opens a polled channel and waits for an ACK the child will never send.

`test_cli_monitor_live_needs_no_external_sampler` hides this: it pre-builds `./hot` with `--with-monitoring` and monitors the binary. Please compile the source on the live path the same way the exact path does, and cover `monitor foo.php --live` without a pre-built binary.

**2. Linux `--attach`: a failed seize is reported as the target dying, and never names yama.**

`ptrace::attach_window` drops every tid whose `seize` fails and returns empty when none succeed. `local::capture_display` turns that into `None`; `run_live` then breaks (“empty window is how we learn the target is gone”) and `run_once` prints “the program may have exited”.

Default `yama/ptrace_scope=1` is exactly this path: `image_for` can read `/proc` and symbols, then every `PTRACE_SEIZE` of a non-descendant gets `EPERM`. `attach_refusal_hint` / `explain_ptrace_scope` exist but are only appended to `image_for` errors in `attach_image`, never to seize failure. The docs say the refusal names the setting; that is false on the common case. `--cap-add=SYS_PTRACE` does not exercise it.

Please distinguish seize-fail from target-gone, and attach the yama hint there.

## Docs still on the old story

These still say `--live` / `--attach` use `/usr/bin/sample` and need macOS:

- `MONITOR_USAGE` in `src/monitor/command.rs` (“Those two use an external sampler, which only macOS ships”)
- `docs/compiling/cli-reference.md` — table rows were updated; the body still says both “read a process from the outside with `/usr/bin/sample`” and “need macOS”. The matrix row for `--live` / `--attach` is still “external sampled CPU only”.
- `src/monitor/mod.rs` module header, and the “still need the external sampler” comment in `run`
- `src/monitor/sampled.rs` header (“macOS-only”)
- `docs/beyond-php/profiling.md` matrix (`--live` / `--attach` “sampled externally”)

## Smaller

- `request_snapshot` that consumes a late ACK and then returns `Late` leaves `run_live`’s `activated` false, so the next window burns the 2s `await_activation` deadline.
- `explain_ptrace_scope` has no test, despite being split out so the wording is testable.
- ptrace unit tests cover only `pending_signal`. The `.php` live path and seize-fail ≠ death need an in-tree oracle if they can be had without `SYS_PTRACE`.
- aarch64 PAC bits on saved LR, as you already noted — leftover is fine.

Not asking to reopen a `waitpid` timeout.

Guikingone:
Both must-fixes confirmed by reading the code, and both were exactly as you
described. Thank you — the second one in particular is a failure mode I had no
way to hit from this machine.

## 1. `--live` on a `.php` source never embedded the probe

Confirmed, and the reachability makes it worse than it looks: `run` returns
through `run_instrument` for a source that is neither `--live` nor `--attach`,
and `--attach` returns before the compile, so **the `.php` branch that called
`compile_php_target` was only ever reached by `--live`**. That helper's single
call site was the one path where it was wrong, so it is gone — the live path now
compiles with `compile_php_monitored`, and `compile_php_target` is deleted rather
than left as a trap.

`test_cli_monitor_live_compiles_the_source_with_the_probe` runs
`monitor hot.php --live` and nothing else — no `--with-monitoring` step — and
asserts the table names a PHP function and that "did not answer within the
window" never appears. Without the probe, both fail.

## 2. Linux `--attach`: a refused seize was reported as a dead target

Confirmed. `attach_window` filtered with `seize(*tid).is_ok()` and returned an
empty vector when nothing survived, which is the same value a window that merely
sampled nothing returns — and attach reads an empty window as proof the target
has gone. `attach_refusal_hint` was only ever appended to an `image_for` error,
and under `ptrace_scope=1` `image_for` **succeeds**: `/proc/<pid>/maps` and the
executable are readable, so the image is built and the symbols load, and only
then does every `PTRACE_SEIZE` come back `EPERM`. The one setting the operator
had to change was named nowhere on the one path that needs it.

Three changes:

- `attach_window` returns `Result`. Failure to attach at all is an error;
  failure *during* a window is still silence, which is right.
- The seize errors are kept rather than dropped, with `EPERM` outranking
  whatever else is held — a thread that exited between the `/proc` read and the
  seize is ordinary and says nothing; a refusal is the whole diagnosis.
- `seize_refusal` builds the message, and `capture_display` now has three
  answers rather than two: `Err` (cannot read this target), `Ok(None)` (read it,
  saw nothing), `Ok(Some(_))`. `run_once` prints the refusal instead of "the
  program may have exited"; `run_live` prints it and ends the view rather than
  ending silently.

## Docs

All five: `MONITOR_USAGE`, `docs/compiling/cli-reference.md` (body and the
`--live` / `--attach` matrix row), the `src/monitor/mod.rs` module header,
`src/monitor/sampled.rs` (header, "Called from", and the macOS-only note, which
now says what it is macOS-only *for*), and the `docs/beyond-php/profiling.md`
matrix rows. The `run` comment you meant is the one that read "still need the
external sampler"; it now says why they still keep their own path, which is that
neither is what `run_instrument` does.

## Smaller

- **`Late` after consuming the ACK.** `Snapshot::Late` now carries
  `activation_seen`, set when the read took the ACK off the front before giving
  up on the reply behind it, and `run_live` does `activated |= activation_seen`.
  Pinned by `a_late_answer_reports_the_ack_it_consumed_on_the_way`.
- **`explain_ptrace_scope`** has a test: each setting names itself and says what
  to do, `0` explains nothing (sending an operator to change a setting that is
  already right is its own bug), and an unknown value still refuses rather than
  going quiet.
- **`seize_refusal`** has one too: `EPERM` names something changeable, `ESRCH`
  does not read as a permission problem, and no threads at all says so.
- **PAC** left as noted, and I have not touched the `waitpid` timeout.

## Rebased onto `origin/main`

The branch was 278 commits behind and the PR had gone `CONFLICTING`/`DIRTY`,
which is why CI had stopped firing on it. Rebased onto `6d14535e7` — 15 commits
replayed, one conflicted.

The conflict surface was small: the branch touches 16 files, main touched 1443,
and only 5 are common. `CHANGELOG.md` kept both sides — the branch's entries
under `[Unreleased]`, main's released `[0.26.6]` section untouched below it (47
and 24 bullets, no verbatim overlap; which of them the release note keeps is a
question for release time, not for a rebase). `docs/internals/architecture.md`
conflicted twice on the source-tree listing, and I resolved each against the
filesystem rather than by which side was newer: there really are four
`shared_*.rs` files, so the branch's line was the accurate one, and
`numeric.rs`, `resource_ids.rs`, `round_mode.rs` and `sysv_call_alignment.rs`
really exist, so main's fuller listing was.

Main touched `src/monitor/` in exactly one place across those 278 commits — a
doc-comment block in `elephc-probe/src/handshake.rs` — so there was no
auto-merge to be suspicious of in the code this PR changes.

## How this was verified, and its limit

`cargo check -p elephc --all-targets` is clean and warning-free on the rebased
tree; 64 monitor unit tests pass, and both live CLI tests pass.

The new CLI test was then run against the defect put back, because a test that
has never been red proves nothing. With `compile_php_monitored` compiling
`--debug-info` again, `test_cli_monitor_live_needs_no_external_sampler` **still
passes** — which is the direct evidence it could not have caught this — and the
new one fails with the defect in plain sight:

```
the live table named no PHP function, so the source was compiled without the
probe and the child had nothing to answer with:
Compiled 'hot.php' -> 'hot'
done522466
```

The program compiled, ran, and printed its own output. There is no table at all.

`ptrace.rs` is `#[cfg(target_os = "linux")]`, so none of it compiles on the
macOS host I work from. I type-checked the whole file — `attach_window` and
`seize_refusal` included — for `x86_64-unknown-linux-gnu` in an isolated crate
with stubs for the rest of `monitor`, and I ran the new pure tests natively by
lifting their text verbatim out of the file rather than retyping it. The rebase
left `src/monitor/` byte-identical (`git diff` against the pre-rebase tip is
empty there), so that verification still describes what is on the branch.

What I still cannot do here is exercise a real refused seize. The in-tree oracle
you asked about is the wording, which is now covered, and the branch that chooses
between refusal and silence, which is not — reaching it needs a live tracee under
`ptrace_scope=1`, and `--cap-add=SYS_PTRACE` grants exactly what makes it not
happen. If you want that covered, the shape that would work is a Linux job that
runs the attach against a sibling process with `ptrace_scope` left at its default
and asserts on the message rather than on the profile. Say the word and I will
add it; I did not want to guess at your CI's constraints.

Guikingone:
Both of the things I left open are closed, and on the first one I was wrong in a
way worth stating.

## The in-tree oracle for "a refused seize is not a death"

I said reaching that branch needed a live tracee under `yama/ptrace_scope=1`,
which no container flag can stage — `--cap-add=SYS_PTRACE` grants exactly what
stops the refusal happening. That is true of the *setting* and wrong about the
*branch*.

**A task cannot trace its own thread group.** Pointing `attach_window` at this
process makes every step before the seize succeed — the image is ours,
`/proc/self/maps` is ours, so `bias_of` returns a bias and the loop reaches the
seize — and only then is every `PTRACE_SEIZE` refused. That is the same shape
scope 1 produces for a process the tool did not launch, and it needs no
capability, no second process and no configuration.

So `a_target_that_cannot_be_seized_is_an_error_not_an_empty_window` now asserts
the thing that actually regressed: `attach_window` answers `Err`, not the empty
window every caller reads as a target that has exited. With `seize_refusal`'s
unit test already covering the wording, the refusal is pinned at both ends —
the branch that chooses it, and what it says.

It only runs where the file compiles, so a Linux shard is still what executes it;
I type-checked it for `x86_64-unknown-linux-gnu` the same way as the rest.

## The changelog

Not pruned — removed, because the policy is unambiguous and I had missed it.
`AGENTS.md`: `CHANGELOG.md` is a release artifact rather than a development log,
a feature or fix Pull Request must not add entries under `[Unreleased]`, that
section stays empty between releases, and only release preparation may add, move
or rewrite bullets.

This branch had accumulated 47 of them, one per change over its life, which is
the exact anti-pattern that paragraph names. Every one describes work `main` has
since released in `[0.26.6]`, written properly by the release task — the monitor
module is in `main` bar the three files this PR adds, and the compiler fixes
among them map onto released bullets one for one. Keeping them would have the
next release report the same work twice, in two voices.

Verified rather than judged: `git diff origin/main -- CHANGELOG.md` is now empty,
so this Pull Request does not touch the file at all. Nothing is lost — the text
is in the branch's history, and the release task reconciles from merged Pull
Requests, not from bullets left behind in a branch.

## One line I did keep

`docs/internals/architecture.md` still carries a one-line change from the rebase:
main's tree listing says `shared_{count_guard,helper,mixed_string}.rs (3 files)`
and there are four — `shared_state.rs` is the fourth. It is unrelated to this PR
and I would normally drop it to keep the diff honest, but taking main's side
would mean knowingly shipping a line that is wrong. Say the word if you would
rather it went to its own change.

nahime0 (changes requested):
Both must-fixes from the last round are closed, and I could confirm them on a real Linux host this time (x86_64, `yama/ptrace_scope=1`, no root — the target was made traceable with `PR_SET_PTRACER_ANY` before its exec, which is enough to reach the seize under scope 1). Running the Linux path also turned up one new defect in the `--live` transport and one regression from the last attach fix. Both need to land before this merges.

## Must-fix

**1. `--live` goes blank after ~8 s of CPU: the snapshot is not cumulative.**

The description says "snapshots are cumulative, like the endpoint's answer", and `run_live` subtracts one snapshot from the previous on that basis. But `sampled_answer()` folds the ring, and the ring holds the last `RING_SLOTS` samples only — `let available = taken.min(RING_SLOTS)` in `crates/elephc-probe/src/lib.rs`. Once the head wraps, the "cumulative" count stops growing and every later window's delta is zero or negative, which the filter drops.

Measured with a CPU-bound program running 14 s, `--duration 2`:

| window | cumulative samples | rows in the table |
|---|---|---|
| 1–3 | 2043 → 6162 | `descend` 100% |
| 4 | 8192 | `descend` 100% |
| 5 | 8196 | `descend` 75% ▼ |
| 6 | 8197 | none |

The program was running flat out the whole time. The CLI fixtures run 6 s at ~1 kHz and never reach 8192, which is why neither test sees it.

The single-reader argument in the description cuts the other way here: the polled channel has exactly one reader, so a reset-on-read for the `POLLED` case loses nothing to "any reader that disconnected". Either the probe answers "samples since the last answer" and resets the head for that channel, or it keeps a cumulative folded counter beside the ring. Whichever it is, please add a test that crosses the ring's capacity (a long enough fixture, or a ring sized down under test).

**2. `--attach --live`: a target that ends is now reported as a refusal.**

`b522f37c6` makes `attach_window` return `Err` whenever `targets` is empty. But `thread_ids` coming back empty (the process has been reaped) and `bias_of` returning `None` (a zombie has no `maps`) are both "the target is gone", not "the kernel refused". With a target that exits after 5 s and `--duration 2`, the view renders three correct windows and then ends with:

```
elephc monitor: cannot attach to the target: it has no threads this tool can read
```

The "ends when the target exits" row in the description was measured before that commit, and the follow-up says the refused-seize branch could not be re-run here. Please return `Ok(empty)` when no seize was actually refused (`refusal` is `None` or `ESRCH`), and `Err` only on `EPERM` and the like. `run_once` has the same race between `attach_image` and the window.

## Confirmed on Linux

- `monitor hot.php --live --duration 1` compiles with the probe, produces three windows naming `descend`, keeps the program's own stderr line, leaks no `elephc-probe:`/`elephc-instr:` rows.
- Attaching to a non-descendant under scope 1: exit 1, the `ptrace_scope` message, target untouched.
- One-shot attach: 296 samples, `inner_hot_loop` 87%, target left in `R` with `TracerPid 0` afterwards. A string/array-heavy program has 100% of its samples rooted at `{main}`, so the frame-pointer walk holds through the runtime.
- Stripped binary and absent pid refused with the right sentence.
- `monitor::` 71 on the bin target (the self-seize test included), both live CLI tests, `fmt`, no clippy warnings in `src/monitor`.

## Smaller

- `attach_image` appends the yama hint to every `image_for` error, so an absent pid reads `cannot read /proc/999999/exe: No such file or directory (the kernel's yama/ptrace_scope is 1, …)` and the stripped-binary message ends the same way. That is a hint naming a cause that did not occur.
- The no-yama arm of `seize_refusal` has a run of spaces inside the literal (`src/monitor/ptrace.rs`, "Run as root,                      grant CAP_SYS_PTRACE").
- A program that finishes inside the first window prints no table and no message, exit 0 — the loop sleeps first and asks second, and the final `elephc-probe:` dump is filtered. Saying so once would do.
- `MONITOR_USAGE` and `cli-reference.md` still say worker children are merged "in both modes"; the launched `--live` now asks only the root over the socketpair.

Unrelated, for a separate issue: a `while (microtime(true) < $deadline)` loop with `$deadline = microtime(true) + 0.3` and a function call in the body never terminates when built with `--with-monitoring`. It reproduces with `main`'s compiler, so it is not this branch.

Guikingone:
Thank you for running it on a real host — both of these needed a Linux box and
one of them is mine.

## 1. The snapshot was not cumulative

Confirmed exactly as you describe. `folded_profile` folds the whole live ring, so
the answer means "the last `RING_SLOTS` samples", and the moment the head wraps
that stops growing while `run_live` keeps subtracting on the documented promise
that it is cumulative.

Taken the second way you offered: the fold now walks TICKETS from a watermark
rather than every slot, and carries what it took into a process-wide
accumulator. What is cumulative is not the ring — it is what has been carried out
of it.

Two things are deliberate. A slot caught mid-write is still skipped and the
watermark still advances past it: one sample lost rather than the same slot
counted again on the next read. And carrying is behind a flag rather than
unconditional, because a fold that carries is not idempotent, and the exit dump
and the unit tests both fold more than once expecting the same answer twice —
the dump reads once with the timer disarmed, where "everything still in the ring"
is exactly right.

`a_lapped_ring_still_answers_a_growing_total` stages the lap rather than running
it: a fixture that takes 8192 samples is eight seconds of CPU in a unit test, and
what is under test is the arithmetic across the lap, not the sampler.

**The test caught me first.** My initial version asserted only that the total
grew after the lap, and it PASSED against the unfixed fold — because the
accumulator alone makes the total grow, by counting the samples still in the ring
twice. So it now also asks with nothing new taken and requires the total not to
move. Each half is sentinel-checked: drop the ticket range and the second
assertion fails on the double count, drop the accumulator and the after-lap
assertion fails on the flat total.

## 2. `--attach --live`: a target that ends read as a refusal

Mine, and you are right about the shape of it — I generalised "not one thread
seized" into "the kernel refused" when the ordinary endings look the same from
there: a reaped process has an empty `/proc/<pid>/task` and a zombie has no
`maps`, and neither is the kernel saying no.

`window_refusal` now makes that distinction: nothing held, or `ESRCH` from the
seize itself (the thread went between the `/proc` read and the syscall), answers
`Ok(empty)` — the empty window attach has always used. `EPERM` and the
unexpected stay an error, since an operator shown an empty table learns nothing
from either. Pinned by `a_target_that_went_away_is_an_empty_window_not_a_refusal`.

I should have seen this when I wrote it: the previous round's fix was about not
folding two answers into one value, and I closed it by folding two OTHER answers
into one.

## Smaller

- **The hint on every image error.** Same class again, so it got the same
  treatment rather than a patch: `image_for` returns `ImageError { reason,
  denied }`, and `attach_image` appends the `ptrace_scope` sentence only when the
  kernel actually refused a read. An absent pid and a stripped binary now end
  where their own sentence ends.
- **The run of spaces** in the no-yama arm: a lost line continuation, fixed.
- **A program shorter than one window** now says so once, on stderr — the loop
  sleeps first and asks second, so silence and a zero exit read as a broken
  profiler rather than as a program that finished.
- **`MONITOR_USAGE` and `cli-reference.md`** no longer claim both modes merge
  worker children: `--attach` discovers and merges the tree, a launched `--live`
  asks the program it started over the channel it handed it, and its answer is
  that process's own.

Nothing touched on the `waitpid` timeout or the PAC bits. The
`--with-monitoring` hang you saw in a `microtime` loop reproduces on `main`, so I
have left it for the separate issue rather than folding it in here.

## Rebased

Also rebased onto `origin/main` — five commits behind this time, one conflict.
Both `docs/` capability matrices had gained a Network column and curl
dimensions upstream, so I took main's shape whole and re-applied only the two
`--live` / `--attach` rows this branch changes.

Guikingone:
Taken here, since it is three lines and the test belongs beside the ones already
in this file — say the word if you would rather it went out as its own change,
it lifts out cleanly as a single commit.

**It is not a `--with-monitoring` defect.** The first thing I did was build your
shape WITHOUT the flag, and it hangs just the same. `--no-ir-opt` terminates, so
it is an EIR pass, and `--with-monitoring` only makes it likelier by inlining
more.

The IR says it plainly:

```
v21: I64 php=int = load_local slot[2] ; origin: licm     <- hoisted out of the loop
while.body:
  store_local v10 slot[2]                                <- which fills it
for.cond:
  v23 = icmp v22 v21 Slt                                 <- the inner for's bound
```

Slot 2 is `$rounds`, the parameter of an INLINED `spin`.
`immutable_local_loads` had proven that slot is written once and cleared
`READS_LOCAL` from its load — which is true in `spin`, where it is a parameter.
The inliner then copied that already-purified load into the caller's loop, where
the slot is written once per ITERATION. LICM took the purity at face value and
hoisted the read above the store, so `for ($i = 0; $i < $rounds; …)` ran against
whatever the frame happened to hold, and never came back.

Note the shape of the wrongness: it is not only a hang. Whenever that stale slot
holds a small number the loop terminates and `spin` simply computes the wrong
answer, silently.

The fix is in the inliner: a purity proof established in the callee's frame does
not travel, so a spliced `LoadLocal` gets its `READS_LOCAL` back.
`immutable_local_loads` runs again on the host and strips it a second time where
it is still true — being conservative costs a round, being wrong costs the
program.

Two things measurement corrected in my own work here, both worth saying because
the first would have shipped:

- I first "fixed" `immutable_local_loads`: its caller checks that the FUNCTION's
  entry block has no predecessor without checking that the single store is IN
  that block, which looked exactly like the hole. Both reproducers pass without
  that change — `store_dominates_loads` already refuses a store outside the
  entry, one function further down. I had judged the caller without reading the
  callee. Removed.
- The test asserts on the spliced instruction rather than by running such a
  program, because the wrong answer there is a HANG and a test that hangs
  reports nothing. Sentinel-checked: drop the three lines and it fails.

`cargo test -p elephc --lib` is 1623 green, and the codegen suite is running.
Your original shape now prints `done…` under the optimiser, and a
non-inlining variant of it agrees with `php -n` at 12.

src/monitor/ptrace.rs
```diff
@@ -0,0 +1,721 @@
...
+/// it is being profiled — which reads as the profiler being slow, not as the
+/// profiler having stopped it.
+pub(crate) fn sample_thread(pid: u32, tid: u32) -> Option<Vec<u64>> {
+    // Below this line the thread is stopped, or on its way to being; above it,
+    // nothing has happened to it.
+    if interrupt(tid).is_err() {
+        return None;
+    }
+    let stopped = wait_for_stop(tid);
+    let signal = stopped.as_ref().map_or(0, |status| pending_signal(*status));
```
nahime0: **[P1] Handle pending stops before requesting another interrupt**

An attached thread can already have a signal-delivery stop pending when this unconditional `PTRACE_INTERRUPT` runs. In the reproduction, the loop subsequently consumes repeated synthetic `PTRACE_EVENT_STOP` notifications while the target makes no progress until the window detaches.

Reproduced with an unstripped, CPU-bound C fixture that installs a `SIGUSR1` handler and calls `prctl(PR_SET_PTRACER, PR_SET_PTRACER_ANY, 0, 0, 0)`. Start `elephc monitor --attach <pid> --duration 2`, wait until `/proc/<pid>/status` names the monitor as `TracerPid`, let it sample for 100 ms, then send `SIGUSR1`. Read the target's state and CPU ticks every 300 ms:

- Without the signal: state `R`, ticks `44 -> 74 -> 104`.
- With the handled signal: state `t`, ticks `14 -> 14 -> 14`.
- At detach: state returns to `R`; after another 300 ms, ticks rise to `44`.

A normal handled signal therefore pauses the program for the remainder of the capture window. Please account for already-pending stops before generating another interrupt and add a Linux tracee regression that asserts continued execution while capture is active. This confirms the pending-stop concern discussed [on August 29](https://github.com/illegalstudio/elephc/pull/794#issuecomment-5461388112).

crates/elephc-probe/src/lib.rs
```diff
@@ -1850,16 +2080,35 @@ unsafe fn folded_profile() -> Option<String> {
...
+    // A slot caught mid-write is skipped, as it always was, and the watermark
+    // still advances past it: one sample lost rather than the same slot counted
+    // again on the next read. The reader that falls more than a full ring behind
+    // loses what was overwritten, which no fixed ring can avoid.
+    let oldest_live = taken.saturating_sub(RING_SLOTS as u64);
+    let from = if cumulative {
+        CARRIED_THROUGH.load(Ordering::Relaxed).max(oldest_live)
+    } else {
+        oldest_live
+    };
```
nahime0, line 2118: **[P2] Serialize the entire cumulative snapshot transaction**

`CARRIED_THROUGH` is read here before taking `CARRIED`'s mutex in `carry_forward`. Two readers can therefore choose and fold the same ticket interval, then each acquire the mutex and add that interval to the cumulative totals. Protecting only the final addition does not prevent double counting.

This is reachable through ordinary endpoint use: `endpoint::dispatch` starts a thread per connection, and `current_folded_profile()` now uses this cumulative path for every sampled response. An endpoint reader can also overlap the launched live control thread.

I reproduced it with the production probe code unchanged: populate all 8,192 slots with settled samples, set the head to 8,192, start two `current_folded_profile()` calls together behind a barrier, and never write another sample. The answers report 8,192 and 16,384 samples; a subsequent serial read still reports 16,384. No artificial delay or externally held accumulator lock is needed.

Please serialize watermark selection, folding, and accumulator/watermark publication as one operation, or provide an equivalent ownership protocol. Extend the ring regression to cover concurrent readers; its existing repeated-read check only covers serial calls.

src/monitor/ptrace.rs
```diff
@@ -0,0 +1,721 @@
...
+/// itself carry a `PTRACE_EVENT_*` in the high bits — `PTRACE_INTERRUPT` reports
+/// `SIGTRAP` with `PTRACE_EVENT_STOP`, and a group-stop under `SEIZE` reports the
+/// stopping signal with the same event. Neither was on its way to the program,
+/// and re-injecting either is how a profiler kills what it measures.
+fn pending_signal(status: libc::c_int) -> libc::c_int {
+    if !libc::WIFSTOPPED(status) || (status >> 16) != 0 {
+        return 0;
+    }
+    let signal = libc::WSTOPSIG(status);
+    if signal == libc::SIGTRAP { 0 } else { signal }
```
nahime0: **[P2] Preserve genuine SIGTRAP delivery stops**

The high-bit check above already excludes `PTRACE_EVENT_STOP` and the other synthetic events. This additional `SIGTRAP` special case also discards a real signal-delivery stop, so `PTRACE_CONT`/`PTRACE_DETACH` receive zero and suppress a signal intended for the program.

Reproduced with a target that installs a `SIGTRAP` handler incrementing a counter. Sending one `SIGTRAP` without the monitor prints `traps=1`; sending it after the monitor has seized the target prints `traps=0`, including after detach and normal target exit.

Please preserve a real `SIGTRAP` while continuing to suppress the sampler's synthetic stop. Add both the missing `pending_signal(stopped(SIGTRAP, 0)) == SIGTRAP` case and a process-level handler test. The synthetic `stopped(SIGTRAP, PTRACE_EVENT_STOP)` case must remain zero. See the signal-delivery and event-stop distinctions in [ptrace(2)](https://man7.org/linux/man-pages/man2/ptrace.2.html).

src/monitor/local.rs
```diff
@@ -157,8 +293,69 @@ pub(crate) fn run_live(cmd: &MonitorCommand, root: u32, mut child: Option<&mut p
...
+        // and the exit dump it wrote to its own stderr is filtered out of a live
+        // view on purpose. Silence and a zero exit reads as a broken profiler
+        // rather than as a program that finished, so say which it was.
+        eprintln!(
+            "elephc monitor: the program finished before the first window; there was nothing \
+             to sample. Use a longer-running input, or a shorter --duration."
+        );
     }
-    0
+    LiveOutcome { code: 0, leave_target_running: lost_channel }
```
nahime0: **[P2] Preserve attach failures in the live outcome**

The `Err(reason)` branch above prints the refusal and breaks, but does not preserve an error outcome. When that happens before the first successful window, this block then claims that the program finished, and the unconditional `code: 0` reports success.

Reproduced under `yama/ptrace_scope=1` with an unstripped sibling target that does not grant tracing permission:

```text
elephc monitor --attach <pid> --live --duration 2
```

The command prints the correct `cannot attach ... yama/ptrace_scope is 1` diagnostic, immediately followed by `the program finished before the first window`, and exits 0. The target is still in state `R` with `TracerPid 0`.

Please retain an explicit failure outcome, return a nonzero exit status, and emit the short-lived-program message only when target termination was actually observed. Cover the live refusal at the caller level, since the existing `attach_window` refusal tests cannot detect this propagation error. The same unconditional message can also contradict the lost-channel diagnostic when a launched target remains alive.

nahime0 (changes requested):
Reviewed `0f21338087f33370781a749e57fd746a7c9e9f2f` on Linux x86_64, without root, with `yama/ptrace_scope=1`. The four additional reproductions below were rerun on September 7.

The two must-fixes from my September 4 review are resolved: a 14-second launched live run now passes 12,000 cumulative samples while continuing to name the hot function, and a naturally exiting attached target produces three windows without a false refusal. The inliner fix also makes the original `microtime` loop terminate with and without `--with-monitoring`.

Four additional issues remain, detailed in the inline comments:

- **P1:** a handled signal can leave the attached target stopped until the window detaches.
- **P2:** concurrent snapshot readers count the same ring tickets twice.
- **P2:** a genuine `SIGTRAP` delivery is discarded.
- **P2:** a refused `--attach --live` exits successfully and also claims that the still-running target finished.

Please fix these and add focused regressions for the observed behavior. The pending-stop issue is the previously discussed uncertainty that a real Linux tracee now reproduces; it is separate from the deferred D-state `waitpid` timeout and PAC limitations.

Validation on this SHA: build succeeded; 75 monitor tests, 16 inliner tests, both live CLI tests, and the ring-wrap regression passed; `git diff --check` was clean. GitHub currently reports 145 successful checks and one skipped check. The new concurrent-reader test fails, and the process-level reproductions confirm the other three defects.

src/monitor/ptrace.rs
Guikingone: Fixed in 3a5a3943177e5d3609c1a89e10df701dc94b1d2d.

`stop_for_sample` first consumes an available status with `waitpid(__WALL | WNOHANG)`. It requests `PTRACE_INTERRUPT` only when no status is pending. Sampling and detachment both use this path.

The new `handled_signal_keeps_running_during_capture` regression uses a real CPU-bound C tracee. It waits for the SIGUSR1 delivery stop, samples for 40 more ticks, and requires CPU ticks to increase **before detachment**. It also checks handler delivery during capture, for a signal pending at detach, and after detach. With the old code restored, it fails with `target stopped making progress: 0 -> 0`.

These process tests passed on native Linux ARM64, including a rerun as UID 65534. That is new ARM64 execution evidence; I have not repeated your Linux x86_64 reproduction locally. The deferred D-state wait timeout and PAC limitations remain unchanged.

crates/elephc-probe/src/lib.rs
Guikingone: Fixed in 3a5a3943177e5d3609c1a89e10df701dc94b1d2d. The watermark now lives inside `Carried`, under the same mutex as the totals. The guard spans ticket selection, folding, and publication; `carry_forward` receives the already-locked accumulator. Rendering uses a cloned result after releasing the guard.

`concurrent_readers_carry_each_ring_ticket_once` fills all 8,192 slots, releases four readers behind a barrier, then performs another serial read without writing samples. Every answer must contain exactly 8,192 samples and the corresponding allocation weight. Restoring the old implementation makes it fail with 16,384 samples and 49,152 allocation units instead of 24,576.

The existing ring-wrap regression and all 63 probe tests pass on Linux ARM64. This covers concurrent readers as well as repeated serial reads.

src/monitor/ptrace.rs
Guikingone: Fixed in 3a5a3943177e5d3609c1a89e10df701dc94b1d2d. The SIGTRAP special case is removed; the stopped-status and event-bit checks remain. The unit test now requires `pending_signal(stopped(SIGTRAP, 0)) == SIGTRAP`, while the synthetic `PTRACE_EVENT_STOP` case still requires zero.

`genuine_sigtrap_reaches_handler_during_capture` exercises a real handler during sampling, with a signal pending at detach, and after detachment. It passed on native Linux ARM64, also as an unprivileged user.

I checked this independently of the pending-stop fix: restoring only the SIGTRAP suppression makes the process regression fail with `tracee did not deliver the signal`.

src/monitor/local.rs
Guikingone: Fixed in 3a5a3943177e5d3609c1a89e10df701dc94b1d2d. `run_live` retains a failure outcome and returns 1 for a capture refusal. The short-lived-program message is gated on observed termination; an empty attached window now checks the process independently, including zombie state on Linux. Losing the control channel while the child is alive also returns failure and leaves the child running.

The new caller-level subprocess tests capture stderr and the returned exit code. They cover a refused self-attach, a broken channel with a live child, and a child that actually finished before the first window. Restoring the old caller makes the first two fail: exit 0 and the contradictory “program finished” message both return. Self-attach exercises a real refused seize without depending on yama configuration; it does not reproduce your exact scope-1 sibling setup.

Validation for this commit: full macOS compiler build without warnings; 70 macOS and 81 Linux ARM64 monitor tests through isolated harnesses compiling the repository's actual monitor modules. The Linux probe run has a pre-existing musl `suseconds_t` deprecation warning.

Greptile's newly reported control-thread startup failure remains a separate open issue: these changes do not close the unserved socket when thread creation fails.

nahime0:
Follow-ups from the latest Grok review (deferred so this PR can merge; I'll take them next):

- #914 — `kind_of` drops `_method_*` frames as non-PHP (was MUST)
- #915 — `FD_CLOEXEC` on control channel after handshake
- #916 — `--live` visibility into prefork workers
- #917 — bound `waitpid` for uninterruptible tracees
- #918 — strip aarch64 PAC bits on walked LR
- #919 — harden `--attach --live` against pid reuse

Prior review threads look addressed in `3a5a39431` / `fb3838bdd`; these residuals are tracked above.

nahime0 (approved):
lgtm

PR #796, opened 2026-08-27
fix(codegen|types): run an inherited private constructor for the selected class

### LINKED ISSUE #795: new static() silently skips a private inherited constructor when a subclass is selected
## Summary

`new static()` skips the constructor entirely — no diagnostic, no crash — when the
runtime-selected class is a **descendant** of the lexical class and the constructor it
inherits is **private**. The object is allocated, its property defaults are applied, and the
constructor body never runs.

That combination is the ordinary singleton / static-factory shape, so the failure lands on
code people actually write:

```php
<?php
class Connection {
    private function __construct() { echo "connecting "; }
    public static function make(): static { return new static(); }
    public function ok(): string { return "ok"; }
}
class PooledConnection extends Connection {}

echo PooledConnection::make()->ok(), "\n";
```

PHP 8.5.6:

```
connecting ok
```

elephc:

```
ok
```

Verified on `main` `71b14b8018` (v0.26.5-155-g71b14b8018), macOS ARM64, against PHP 8.5.6.

Three conditions must hold together; changing any one of them makes the constructor run:

| Case | PHP | elephc |
|---|---|---|
| `new static()`, descendant selected, **private** inherited ctor | `P v` | `v` ❌ |
| `new static()`, descendant selected, private ctor chained to a private grandparent ctor | `B P v` | `v` ❌ (whole chain skipped) |
| `new static()`, descendant selected, **public** inherited ctor | `P v` | `P v` ✔ |
| `new static()`, descendant selected, **protected** inherited ctor | `P v` | `P v` ✔ |
| `new static()`, **no** subclass (selected class declares the ctor) | `P v` | `P v` ✔ |
| `new self()`, descendant exists, private ctor | `P v` | `P v` ✔ |

So it is specifically: **`new static()` + the called class is a descendant + the constructor
is private**. With a public or protected constructor the child inherits the method normally
and everything works, which is why this hides so well.

Constructor arguments do not change the outcome — `new static(7)` with
`private function __construct(int $a)` also runs nothing.

## Suspected cause

Private methods are deliberately not inherited into a child's method map —
`ClassState::inherit_methods` (`src/types/checker/schema/classes/state.rs:498-501`) skips
them, which is correct for visibility, since `C` genuinely does not inherit a private `P::m()`.

But the **constructor** is not just a visible method: PHP still runs an ancestor's private
constructor when instantiating a descendant, provided the *call site* is inside the class that
declares it. `new static()` inside `Connection::make()` is exactly that scope.

`new static()` lowers to `Op::DynamicObjectNew`
(`src/ir_lower/expr/mod.rs:14316`, `lower_new_scoped_object`), and each runtime candidate
resolves its constructor from the visible map alone —
`dynamic_new_candidate` (`src/codegen/lower_inst/objects.rs:1986-2010`):

```rust
let constructor_impl = if let Some(constructor) = class_info.methods.get(&constructor_key) {
    ...
} else if arg_count == 0 {
    None                       // <-- no constructor, silently
} else {
    return Ok(None);
};
```

For `PooledConnection`, `methods` has no `__construct` (the private one was not inherited),
`arg_count == 0`, so `constructor_impl` is `None` and `emit_dynamic_new_candidate`
(`objects.rs:2185-2200`) allocates the object and skips the call. With arguments it takes the
`return Ok(None)` arm instead, dropping the candidate — the same root cause, different
symptom.

Resolving `__construct` through the ancestor chain when the visible map has no entry should
fix it. PHP's accessibility rule has to come with it, though: a private constructor is usable
only from the class that declares it, so

```php
class G { private function __construct() {} }
class P extends G { public static function make(): static { return new static(); } }
class C extends P {}
C::make();
```

must stay an error — PHP 8.5.6 says
`Error: Call to private G::__construct() from scope P`, and elephc should too rather than
start calling it.

## Why this one matters

There is no warning and no crash. The object comes back looking fine and every field the
constructor would have set is silently left at its declared default — a null connection
handle, an unseeded id, an empty collection. `private __construct` plus a `static` factory is
the standard way to write a singleton or a named constructor in PHP, and adding one subclass
to such a class is enough to trigger it.

Guikingone:
## Summary

Fixes #795. `new static()` allocated the object and ran **no constructor at all** whenever the
runtime-selected class was a descendant and the constructor it inherited was private — the
ordinary singleton / static-factory shape, failing silently:

```php
<?php
class Connection {
    private function __construct() { echo "connecting "; }
    public static function make(): static { return new static(); }
    public function ok(): string { return "ok"; }
}
class PooledConnection extends Connection {}
echo PooledConnection::make()->ok(), "\n";
```

| | before | after | php 8.5.6 |
|---|---|---|---|
| | `ok` | `connecting ok` | `connecting ok` |

## Why it happened

A private method is not inherited, so `ClassState::inherit_methods` deliberately drops it and a
descendant's `ClassInfo::methods` carries no `__construct`. Every consumer then read that as
"this class has no constructor".

PHP draws the line differently for constructors, and its own reflection shows both halves:

```php
class Owner { private function __construct() {} }
class Child extends Owner {}
method_exists('Child', '__construct');                                        // false
get_class_methods('Child');                                                   // []
(new ReflectionClass('Child'))->getConstructor()->getDeclaringClass()->getName();  // "Owner"
(new ReflectionClass('Child'))->hasMethod('__construct');                      // true
```

So the descendant does not *inherit* the private method as a callable member, yet the class still
*has* a constructor, resolved through the chain to its declaring class. This keeps `methods` as it
is and resolves `__construct` through the ancestor chain at each point of use, which models both
halves. Putting the entry into `methods` instead would make `method_exists()` answer `true`, and
would also break `prune_keeps_private_override_vtable_hole_on_descendant`, which asserts the
current shape for a good reason.

## The change

`types::constructor_owner()` walks to the nearest ancestor whose map still owns `__construct`, with
a cycle guard. It returns the class itself whenever that class has the entry, so an inherited
public or protected constructor resolves exactly as before and the common path is unchanged.

Three callers:

- `types/checker/inference/objects/constructors.rs` — owns both the visibility error and the arity
  check, so it fixes the diagnostics and the argument case;
- `codegen/lower_inst/objects/fixed_new.rs` — `new Child()` named directly;
- `codegen/lower_inst/objects/dynamic_factory.rs` — the `new static()` candidate ladder.

## Three further symptoms, same root cause, fixed here too

**Naming the descendant directly also skipped the constructor**, not only `new static()`:

```php
class Owner { private function __construct() { echo "built "; } 
              public static function makeChild(): Owner { return new OwnedChild(); } }
class OwnedChild extends Owner {}
Owner::makeChild();     // before: nothing   after: "built"   php: "built"
```

**Where php raises an `Error`, elephc silently succeeded** — constructing an object php refuses to
build. `new Owner()` was already rejected correctly; only the descendant slipped through:

```php
class Owner { private function __construct() {} }
class Child extends Owner {}
new Child();
// before: constructed
// after:  Cannot access private constructor: Owner::__construct
// php:    Error: Call to private Owner::__construct() from global scope
```

The diagnostic now names the class that **declares** the constructor rather than the one being
instantiated, matching php — including when it comes from further up the chain
(`Root::__construct` for a `Leaf` whose lexical scope is `Middle`).

**With arguments the symptom flipped to a spurious compile error**:

```php
class Tagged { private function __construct(private int $tag) {}
               public static function make(int $tag): static { return new static($tag); } }
class SubTagged extends Tagged {}
SubTagged::make(7);
// before: error: Constructor 'SubTagged::__construct' expects 0 arguments, got 1
// after:  7          php: 7
```

## Testing

Focused per `AGENTS.md`; the full sharded matrix is CI's job.

- 5 new codegen tests in `tests/codegen/oop/relative_types.rs`: the plain case, the
  `parent::__construct` chain, constructor arguments, the fixed `new Child()` path, and the
  realistic singleton shape with dispatch through the base type.
- 3 new error tests in `tests/error_tests/classes_traits.rs`: descendant from global scope,
  `new static()` from a non-declaring scope, and the grandparent case naming its declaring class.

```
cargo build                                       warning-free
cargo test --test error_tests                     1440 passed
cargo test -p elephc --lib                        1488 passed
cargo test --test codegen_tests -- oop::           593 passed
cargo test --test codegen_tests -- objects::       246 passed
cargo test --test codegen_tests -- optimizer::     327 passed
cargo test --test codegen_tests -- eval_constructors        11 passed
cargo test --test codegen_tests -- static_class_features    18 passed
```

Behaviour verified against PHP 8.5.6 on macOS ARM64 across 13 shapes, including the ones that must
**not** change: protected constructors, public constructors, a class with no subclass, `new self()`,
and `new Owner()` from global scope (already an error before).

### DIFF
--- a/src/codegen/lower_inst/objects/dynamic_factory.rs
+++ b/src/codegen/lower_inst/objects/dynamic_factory.rs
@@ -98,7 +98,13 @@ pub(super) fn dynamic_new_candidate(
     }
     let constructor_key = php_symbol_key("__construct");
     let mut padding_thunk = None;
-    let constructor_impl = if let Some(constructor) = class_info.methods.get(&constructor_key) {
+    // Same reason as the fixed-class path: a runtime-selected descendant of a class with a
+    // private constructor carries no `__construct` entry of its own, and PHP still runs the
+    // ancestor's. The owner is the candidate itself for every other constructor.
+    let constructor_owner = crate::types::constructor_owner(&ctx.module.class_infos, class_name);
+    let constructor_impl = if let Some((owner_name, constructor)) = constructor_owner
+        .and_then(|(name, info)| info.methods.get(&constructor_key).map(|sig| (name, sig)))
+    {
         // A site that passes FEWER arguments than the constructor declares is still a candidate
         // when every omitted parameter has a default and `ir_lower` emitted the padding thunk for
         // this (class, argc) pair. `new $c(…)` cannot be padded by the checker — it does not know
@@ -126,11 +132,9 @@ pub(super) fn dynamic_new_candidate(
                 padding_thunk = Some(thunk);
             }
         }
-        let impl_class = class_info
-            .method_impl_classes
-            .get(&constructor_key)
-            .cloned()
-            .unwrap_or_else(|| class_name.to_string());
+        let impl_class = constructor_owner
+            .and_then(|(_, info)| info.method_impl_classes.get(&constructor_key).cloned())
+            .unwrap_or_else(|| owner_name.to_string());
         if !class_method_already_emitted(ctx, &impl_class, &constructor_key, false) {
             return Ok(None);
         }
--- a/src/codegen/lower_inst/objects/fixed_new.rs
+++ b/src/codegen/lower_inst/objects/fixed_new.rs
@@ -57,7 +57,14 @@ pub(in crate::codegen::lower_inst) fn lower_object_new(ctx: &mut FunctionContext
             )));
         }
         let property_defaults = collect_property_defaults(class_info, inst)?;
-        let constructor_impl = if let Some(constructor) = class_info.methods.get(&constructor_key) {
+        // A private constructor is not inherited, so a descendant of a class that declares one
+        // carries no `__construct` entry of its own. PHP still instantiates it through that
+        // ancestor, so resolve the owner; it is the class itself for every other constructor.
+        let constructor_owner =
+            crate::types::constructor_owner(&ctx.module.class_infos, &class_name);
+        let constructor_impl = if let Some((owner_name, constructor)) = constructor_owner
+            .and_then(|(name, info)| info.methods.get(&constructor_key).map(|sig| (name, sig)))
+        {
             if constructor.params.len() != inst.operands.len() {
                 return Err(CodegenIrError::unsupported(format!(
                     "constructor call to {}::__construct with {} args for {} params",
@@ -66,11 +73,9 @@ pub(in crate::codegen::lower_inst) fn lower_object_new(ctx: &mut FunctionContext
                     constructor.params.len()
                 )));
             }
-            let impl_class = class_info
-                .method_impl_classes
-                .get(&constructor_key)
-                .cloned()
-                .unwrap_or_else(|| class_name.clone());
+            let impl_class = constructor_owner
+                .and_then(|(_, info)| info.method_impl_classes.get(&constructor_key).cloned())
+                .unwrap_or_else(|| owner_name.to_string());
             if !class_method_already_emitted(ctx, &impl_class, &constructor_key, false) {
                 return Err(CodegenIrError::unsupported(format!(
                     "constructor call to {}::__construct without an emitted EIR method body",
--- a/src/types/checker/inference/objects/constructors.rs
+++ b/src/types/checker/inference/objects/constructors.rs
@@ -106,13 +106,24 @@ impl Checker {
                     &format!("Cannot instantiate abstract class: {}", class_name),
                 ));
             }
-            if let Some(sig) = class_info.methods.get("__construct") {
-                if let Some(visibility) = class_info.method_visibilities.get("__construct") {
-                    let declaring_class = class_info
+            // A private constructor is not inherited, so a descendant's own method map has
+            // no `__construct` entry. PHP still instantiates the descendant through the
+            // ancestor that declares it, and reports that declaring class when the call site
+            // may not reach it, so resolve the owner before reading the constructor contract.
+            let constructor_owner = crate::types::constructor_owner(&self.classes, class_name.as_str())
+                .and_then(|(owner_name, owner_info)| {
+                    owner_info
+                        .methods
+                        .get("__construct")
+                        .map(|sig| (sig, owner_name, owner_info))
+                });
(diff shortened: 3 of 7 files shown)


Guikingone:
Thanks — the defect is real and I have filed it as #797, but the diagnosis and the merge blocker
do not hold. Measured on `main` `71b14b8018`, without this PR applied:

```php
<?php
class Tagged {
    public function __construct(private int $tag = 5) {}   // PUBLIC
    public static function make(): static { return new static(); }
    public function tag(): int { return $this->tag; }
}
class SubTagged extends Tagged {}
echo SubTagged::make()->tag();
```
```
php:            5
main (no PR):   error: unsupported EIR backend feature: dynamic object construction ...
this PR:        error: unsupported EIR backend feature: dynamic object construction ...
```

The constructor is **public**, so no owner resolution is involved and the descendant inherits
`__construct` into its own method map exactly as before. It still fails, identically, with and
without this PR. The same holds for partially-omitted defaults and for a variadic constructor.

The mechanism is not "thunk emission still performs a descendant-local lookup". It is that **no
thunk is ever emitted for a `new static()` site at all**:
`lower_dynamic_constructor_thunks` (`src/ir_lower/program/spl_discovery.rs:363`) collects its
arities by scanning for `Op::DynamicObjectNewMixed` only, and `new static(...)` lowers to
`Op::DynamicObjectNew` (`lower_new_scoped_object`). A module whose only dynamic construction is
`new static()` collects no arities, emits no thunks, and every candidate is then dropped by the
`!emitted` guard. `new self()` and a directly named `new Child()` take the fixed path and are
unaffected, which is why they work with the very same constructors.

So this is a pre-existing gap in `new static()` argument padding, orthogonal to what this PR
changes — which constructor is *resolved*, not how omitted arguments are *padded*. This PR neither
introduces it nor widens it. Details, the affected/unaffected matrix, and the suspected fix are in
#797.

One thing your summary got exactly right that I want to keep: the four sites really do have to
agree, and `lower_dynamic_constructor_thunk` is a fourth reader of the same method map. It bails
on `class_info.methods.get("__construct")` like the other three did. It is unreachable for
`new static()` today for the reason above, so changing it here would be untestable — but when
#797 wires the thunk driver to `Op::DynamicObjectNew`, that lookup has to be resolved through the
owner in the same move, or the private-constructor case will fail again at the newly reachable
site. I have noted it on #797.

nahime0 (approved):

PR #798, opened 2026-08-28
feat(monitor): `--attach` reads a Linux process instead of refusing it

Guikingone:
> **Stacked on #794.** Its three commits are the first three here, so this diff
> currently carries them; review the last commit alone, or wait for #794 to land
> and the diff narrows to it by itself. (The base could not be set to #794's
> branch: it lives on the fork, and an upstream PR cannot base on it.)

## What this is

`--attach` was the last thing in `monitor` that a platform could refuse. #794
removed the refusal for `--live` by observing that it *launches* its target, so
it can hand it a socketpair and ask. `--attach` cannot: it is handed a pid
already running under someone else's control, so reading it from the outside is
not a shortcut, it is the whole job. macOS shells out to `/usr/bin/sample`.
Linux has nothing to shell out to, so elephc does it itself — seize each thread,
stop it, read `pc` and `fp`, walk the frame chain, resume.

The syscalls and the loop that drives them live alone in `ptrace.rs`, because
they are the one part no test on a macOS host can run. Everything that turns
what they read into a profile — the ELF symbol table, the naming, the folding —
was separated for that reason and is exercised on every host.

## Two defects that only a real program showed

Neither announced itself, and both were in code that already had tests.

**`function_symbols` returned the linker's order; `symbolize` binary-searches.**
The parameter is literally named `sorted`, and nothing had ever called both. An
unsorted binary search does not fail — it answers `None` for nearly every
address. A 12-second spin loop came back profiled as 251 samples of `<native>`,
which reads as *this program is not PHP* rather than as *the names were never
looked up*. The producer now guarantees the order its only consumer requires,
and the test asserts it where it is produced: one that sorted first and then
searched would pass against the bug.

**`display_stack` published the mangled spelling.** The fix is ordered, not
cosmetic. The prefix that says whether a frame is PHP or a runtime helper —
`fn_`, `__rt_` — is exactly what demangling removes, so classifying after it
leaves every PHP function classified native, and a profile claims a PHP program
spends none of its time in PHP. Classify on the mangled name, display the source
one.

## Verified against a running process

In a Linux container (`--cap-add=SYS_PTRACE`), attached to a live 12-second
program:

```
{main} -> outer_driver -> inner_hot_loop     74.4%  self 13.4%
    memory release       23.6%
    heap allocation      13.8%
    Mixed cell boxing     6.5%
```

and afterwards the target is `state=R` with its utime climbing 396 → 596 — a
sampler that gives its target back, which is the property that matters most when
the thing you attached to is production. Both fixes are mutation-verified:
removing the sort turns the order assertion red, and classifying after
demangling turns every PHP frame native.

## The refusals name what to do

Reading from the outside asks two things of a target that asking it does not, so
neither is a bare `EPERM`:

- a stripped binary says to rebuild with `--keep-symbols` (elephc strips by
  default, because nothing *inside* a program reads its symbols — a sampler
  reads them from the outside);
- a kernel that forbids tracing names the `yama/ptrace_scope` setting that did
  it.

## Docs

`--attach` was called macOS-only in four places. One of them also still said
`--live` reads a process from the outside, which stopped being true one commit
earlier.

## Suites

macOS: `-p elephc --bins` 1678, `--lib` 1488, `codegen::cli` 37; clippy and
`fmt` clean. Linux (container): `monitor::` 61, clippy clean on `src/monitor`.


Guikingone:
Pushed a second commit: the Linux path now measures the **worker tree**, not just
the pid it was handed.

I had shipped `pids.first()` with a comment explaining why one pid was enough.
It was not an explanation — it was the limitation restated. `--attach` is
documented to merge a target's worker children so a `--web` prefork server is
measured across all of them, and the macOS path does exactly that; mine would
have reported a prefork parent's idle accept loop as the whole program.

The obstacle was real but small: workers share the *image* (they are forks of one
executable, so one symbol table answers for all of them) but not the *bias* —
each is mapped wherever its own `/proc` says. So `Image` carries the executable
path and its first load address, and every process derives its own bias. Reusing
the root's would not fail; it would resolve each worker's addresses against the
wrong offset and name the wrong functions, which is worse than a short table
because it is confidently wrong. A process whose bias cannot be read is dropped
rather than resolved against a neighbour's.

Every process is sampled on every tick rather than one per window each — a worker
given a third of the window would otherwise contribute a third of the samples,
and its share of the table would be a third of the truth.

Verified against a real tree in a Linux container: a program that re-execs itself
as a worker, parent and child spinning in differently named functions, mapped at
genuinely different bases (`ffff8daf0000` vs `ffffb8fe0000` — so the per-pid bias
is load-bearing, not defensive). Attaching to the parent alone:

```
samples: 672 · 2 processes
child_work    40.3%   self 4.6%
parent_work   36.8%   self 3.7%
```

and both processes are still running afterwards.

---

One finding out of scope for this PR, recorded here because it cost me an hour
and will cost the next person the same: **`getenv()` of an unset variable returns
`string(0) ""` rather than `bool(false)`.** A set variable is fine; only absence
lies. That makes `getenv('X') !== false` — the idiom for "is this set" — always
true. I hit it writing the fixture above: the parent silently ran the child's
branch, which first looked like the profiler finding no children. Worth its own
issue.

Guikingone:
Two more, both found by exercising the thing rather than by reading it — and the
second one had been reporting a *wrong answer*, not an error.

**A wait that fails must not leave the thread stopped.** `resume`'s own docblock
claimed it runs on every path out of a stop. It did not: `interrupt_and_wait` did
two things and reported one failure for both, so a wait that failed returned
before any resume with the thread already stopped. One `EINTR` was enough, and a
profiler is exactly the process that has signals about. The thread came back when
the window ended and the detach released it, so the symptom is a target that
stalls for as long as it is profiled — which reads as *the profiler is slow*,
not *the profiler stopped my program*. The invariant is structural now:
`interrupt` is the request alone, and past the line where it succeeds the resume
sits below every path that can fail. The `EINTR` path is reasoned, not
reproduced; what is measured is the invariant it threatens — after a window,
every thread of every process is back in `R`, none in `t`.

**`--attach --live` drew one window and said the target was gone.** It was still
running, with eight seconds to go. `PTRACE_DETACH` requires a *stopped* tracee
and restarts it as it detaches; asked of a running one it fails with `ESRCH` and
changes nothing. The window ended by resuming every thread and then detaching, so
no detach ever took — silently, because nothing reads that result. Seizes happen
per window and a thread already traced cannot be seized again, so window two
seized nothing; and an empty window is precisely how `--attach` learns its target
is gone, having no channel to ask. A failure nobody was told about became a
confident wrong answer.

Measured before and after rather than by mutation, since no test on a macOS host
reaches ptrace:

| | before | after |
|---|---|---|
| live view of an 8s target | exits at 2s, 1 window | full 8s, 4 windows + cumulative |
| `TracerPid` after window 1 | still set | cleared |
| window 2 | seized 0 of 1 | seizes 1 of 1 |
| ending | claimed the target was gone | ends when it actually exits |

The one-shot and worker-tree captures are unchanged, and no thread is left
stopped in either.

Guikingone:
Folded into #794 — the two were the same branch plus six commits, so keeping
both open only showed the same work twice. Everything here is now in #794, with
nothing dropped: `b3e286677d` through `227fef77d7`.

Nothing in this PR is lost, including the two review rounds recorded above:
the worker-tree fix, the resume invariant, and the detach that never took.

PR #802, opened 2026-08-28
feat(env): `getenv()` answers the whole environment, and the superglobals carry it

### LINKED ISSUE #806: eval: getenv() of an unset name still answers "", not false
Follow-up from #801 (AOT `getenv` of an unset name answers `false`). Not a merge blocker for that PR. Also not #802 (`getenv()` with no args, `local_only`, `$_ENV` / `$_SERVER`).

#801 restored PHP's missing-vs-empty split for **AOT**. Magician `eval()` still collapses both to `""`, so `getenv($name) !== false` is true for every name — the same silent polarity #801 fixed on the compiled path.

```php
<?php
var_dump(getenv('DEFINITELY_NOT_SET'));
var_dump(getenv('DEFINITELY_NOT_SET') !== false);
```

php / elephc AOT (after #801): `bool(false)` then `bool(false)`.
`eval()` today: `string(0) ""` then `bool(true)`.

Same after `putenv("NAME")` (unset): AOT `false`, Magician `""`.

Path: `crates/elephc-magician/src/interpreter/builtins/network_env/getenv.rs` (`unwrap_or_default()` / “empty string when it is unset”). Out of #801's files. `examples/eval/main.php` still teaches `=== ""` for a cleared name.

Measure against `php -n` the same table #801 used: missing → `false`, empty → `""`, set → the string, `putenv("NAME")` → `false`.

Guikingone:
> **Stacked on #801.** Its commit is the first here; review the last one alone, or
> wait for #801 to land and the diff narrows to it by itself.

PHP's CLI SAPI hands a script three things elephc did not have:

| | php | before | after |
|---|---|---|---|
| `getenv()` (no argument) | `array<string,string>` | **compile error** | the environment |
| `getenv($n, true)` | works | **compile error** | accepted |
| `count($_ENV)` | 59 | `0` | 59 |
| `count($_SERVER)` | 68 | `0` | 68 |
| `$_SERVER['argv']` | present | absent | present |

The superglobals existed but were seeded **empty**, which reads to any program
that looks like *not set* rather than like a gap.

Measured against `php -n` throughout, on macOS and in a Linux container: the
array and its count, a value containing `=` (the **first** one separates —
splitting on the last renames every variable whose value holds one), `false` for
a name that is not set, `local_only`, `$_ENV == getenv()`, `$_SERVER`'s nine own
keys, and the five request superglobals staying empty as they are in PHP.

## Decisions taken from measurement rather than from the manual

**`local_only` is accepted and ignored**, because that is what it means here: in
the CLI SAPI `getenv($n)` and `getenv($n, true)` agree for a shell variable, a
`putenv` one and `PATH`, and `getenv() == getenv(null, true)`. There is no
environment separate from the process's for the flag to select.

**The four path-shaped `$_SERVER` keys are `$argv[0]`.** PHP names the script it
was handed; a compiled program has no script at run time, and the thing that
*was* invoked is the closest true answer rather than a fabricated path.

**`putenv` does not reach `$_ENV`/`$_SERVER`, only `getenv()`.** They are
snapshots taken before the program ran. PHP has the same asymmetry — checked,
because it is easy to get wrong in either direction — and it is asserted.

**Seeding stays pay-for-use**: only the names the source spells are emitted, which
is also what PHP's `auto_globals_jit` does, and for the same reason.

## The environment is read live

The first design captured the `envp` handed to `main`, which avoids a platform
split. It was wrong in a way only `putenv` shows: libc may reallocate the entry
vector when a variable is added, so the startup pointer goes stale — and
elephc's own one-name `getenv($n)` goes through libc and saw the addition, so the
two halves of one builtin disagreed. Reading it live costs the split back
(`environ` on Linux, `_NSGetEnviron()` on macOS), which is the same shape as
`errno` elsewhere in this runtime. The startup capture is removed rather than
left behind looking load-bearing.

## Three defects of mine on the way, all one shape

A convention assumed instead of read.

- `__rt_hash_set`'s x86_64 registers were **guessed**, and wrong — that would have
  corrupted the hash silently on Linux.
- `__rt_hash_new` was called with **no arguments**, so it took whatever was in the
  capacity register and asked for `garbage * 64` bytes. Its own docblock warns
  about exactly that.
- The long one: the result was declared `AssocArray` while the value is a hash
  **boxed** in a Mixed cell, so `count()` read the cell's tag and answered `5` for
  a 65-entry environment. `getdate` declares `Mixed` for precisely this reason.
  The type is the representation, and losing `array<string,string>` is the price
  of the box.

## Two existing tests pinned the old behaviour

One pinned a diagnostic **PHP does not have**: `unset($_SERVER); $_SERVER = 5;` is
legal PHP and elephc refused it. Its guard moves to `$_GET`, whose type still
makes the refusal observable, and the property it actually protects — `unset` must
not abandon global storage — is now measured directly, as a heap *difference*.

That measurement turned up a **pre-existing leak**, out of scope here: a boxed
assoc array from a builtin is never released. Confirmed against `getdate()`,
which leaks its 13 blocks the same way (`allocs=15 frees=2`) and predates all of
this. Same shape likely in `stat()`, `localtime()`, `parse_url()`,
`ob_get_status()`. Worth its own issue.

## Suites

`elephc` lib 1488 · bins 1661 · `error_tests` 1437 · codegen misc 41 ·
`constants_and_system` 134; `fmt` and clippy clean. Verified on Linux in a
container, including the `putenv` case. A `--web` build still skips the seeding
entirely: it references no `__rt_getenv_all` and keeps its request prelude.

Fixes #806


Guikingone:
Rebased onto #801's docs fix, and regenerated for this PR's own signature change — `getenv` now documents `(string $name = null, bool $local_only = false)` in the public builtin index and its page, which is what the catalog entry here declares.

Same gate as #801, one level up: changing a builtin's parameters changes its generated documentation, and the CI gate regenerates and compares rather than trusting the committed page. Folded into the feature commit rather than added as a follow-up, so that commit is self-consistent.

Guikingone:
Swept the prose docs for claims this makes false, and found four.

Three were plainly stale — `docs/php/namespaces.md` described `$_SERVER` and `$_ENV` as seeded **empty** in CLI builds (that empty seeding is exactly what this closes), and `docs/php/system-and-io.md` carried the one-parameter signature. `docs/internals/the-runtime.md` gains `__rt_getenv_all` and now says what `__rt_getenv` returns for a name that is not set.

The fourth was worse than stale, because it gave a **reason** rather than a rule. `docs/compiling/cli-reference.md` said an empty environment variable is treated as unset by the opcache override *"because `getenv()` cannot distinguish the two"*. It can, as of #801. I checked the code rather than assuming a regression: the behaviour is unchanged, because that override reads its value through a `(string)` cast and an unset name and an empty one both arrive as `""` after it. So the explanation needed correcting, not the rule — a doc that argues against its own code is worse than one that is merely behind.

All four pages are hand-written; the generator was re-run and left them alone, and no generated page drifted.

Guikingone:
Both findings addressed in `86f6cf2b1c`.

**The x86_64 alignment is a real defect, and it was backwards from what the
comment claimed.** A `call` leaves `rsp` at 8 mod 16 on entry; the five pushes
bring it back to 0. The `sub rsp, 8` then took it to *8* — misaligning
`__rt_hash_new`, `__rt_str_persist` and `__rt_hash_set` — under a comment reading
"keep rsp 16-byte aligned for the nested calls".

Removed rather than widened to 16: nothing used those eight bytes, so the padding
had no purpose beyond the alignment it was breaking. Prologue and epilogue are
five pushes and five pops, and the arithmetic is now written down so the next
reader can check it instead of trusting the comment.

Not observed as a crash — aarch64 is unaffected and the x86_64 shard is CI's, not
this machine's — so I am reporting it as a convention violation that leaves the
path at the mercy of any callee using aligned moves, not as something I saw fail.

**On the documentation**: `docs/php/system-and-io.md` and `docs/php/namespaces.md`
were updated earlier on this branch (`3a81ee50ad`), along with the generated
builtin pages. What was genuinely missing was the example, and you were right to
ask — `examples/system-info/main.php` already covered `getenv()` and stopped at
the one-argument form. It now shows the three forms, the false-versus-empty
distinction, `$_ENV`/`$_SERVER`, and the asymmetry where a later `putenv()`
reaches `getenv()` but not the snapshots. Its environment section is byte-identical
to `php -n`, which is how I know the asymmetry is PHP's and not ours.

nahime0:
I reviewed this PR against the current state of `main`. First, it needs to be **rebased onto `main`**: #801 and several subsequent changes have landed in the meantime, so #802 is now `CONFLICTING` and is no longer correctly aligned with the base and contracts introduced by #801. After the rebase, the conflicts should be resolved, the generated documentation refreshed, and the three-target verification rerun.

In addition to the rebase, I reproduced three blocking issues:

1. `getenv(null, true)` is valid PHP and should return the complete environment, but the backend treats it as a single-variable lookup and compilation fails with `getenv name for PHP type Void`. The “whole environment” mode is currently selected only when the instruction physically has zero operands. The named form `getenv(local_only: true)` should be covered as well.
2. The shared contract now accepts 0–2 arguments, but the Magician/eval backend still accepts exactly one. For example, `eval('echo is_array(getenv()) ? "array" : "not-array";');` terminates with `Fatal error: eval() runtime failed`; `getenv($name, true)` is also rejected under eval.
3. Every call to `getenv()` leaks the complete environment hash. With `--heap-debug`, three calls under a minimal environment report `allocs=18 frees=3 live_blocks=15 live_bytes=25032`: the `Mixed` box is released, while the hash, keys, and values remain live. The ownership path needs to be completed, with a regression test under `tests/codegen/runtime_gc/`.

There are also two stale docblocks: the arity test still says that calling `getenv()` without arguments is an error, and the `readline` module preamble says `Union(Str, Bool)` while the implementation now uses `Union(Str, False)`.

The published CI is green and the direct AOT tests pass, but they do not cover the three paths above. My current assessment is **changes requested**: first rebase onto `main`, then fix the three blockers and add the corresponding regression tests.

nahime0:
Follow-ups from the latest Grok re-review (`d1b0610`; deferred for merge once CI is green):

- #935 — Magician/eval seed `$_ENV` / `$_SERVER`
- #936 — `examples/eval` still teaches unset getenv as `\"\"`
- #937 — checker: type `getenv(local_only: true)` via `argument_plan`
- #938 — verify boxed-assoc decref for getdate/stat/…
- #939 — CI macos codegen `download-artifact` 403 flake
- #940 — `consumer_window_enables_timing` macos flake (re-check vs `049eeea`)

#806 remains open: Magician polarity looks fixed on this tip; close it when this PR lands (add `Fixes #806` if you want auto-close). Noted on the issue.

PR #803, opened 2026-08-28
fix: prevent loop inlining leaks and guard Mixed-array sorting

Guikingone:
## Summary

Fix a per-iteration ownership leak when the EIR inliner moves a callee's owning local stores into a caller loop, and support scalar Mixed-array sorting with an explicit fatal for unsupported values.

```php
function sized(): int { $rows = ["a" => 1, "b" => 2]; return count($rows); }
$total = 0;
for ($i = 0; $i < 50; $i++) { $total += sized(); }
```

The local store was lowered outside a loop in the original callee, so it has no release-of-previous. Inlining it into the caller loop overwrote the transplanted slot on each iteration without releasing the previous value.

## Inliner and ownership

- Refuse loop call sites whose callee stores release-requiring refcounted values into locals, including owned or maybe-owned arrays, boxed Mixed values, and strings such as `getcwd()` results.
- Keep calls outside loops eligible for inlining, preserving the existing cleanup behavior for ordinary calls and exception unwinding.
- Recompute loop information after each inline splice and add IR and heap regression coverage for array and string locals.
- Retain the `getdate()`/`localtime()` fresh-hash heap regression. Their boxing ownership fix is already on `main` via #802, not a new runtime ownership fix in this PR. The original description incorrectly identified `getenv()` as another instance of that leak.

## Mixed-array sorting (AOT)

- Accept `sort()` and `rsort()` on indexed arrays with runtime-typed elements (`Array(Mixed)`). Scalars (`bool`, `int`, `float`, and `string`) and `null` are compared through `__rt_php_compare`.
- Before sorting, `__rt_mixed_sort_require_scalars` checks all elements. Nested arrays, objects, resources, and boxed callables terminate execution with:

  ```text
  Fatal error: sorting Mixed arrays containing non-scalar values is not supported
  ```

- This is a deliberate restriction that replaces silently incorrect non-scalar ordering. Full PHP container ordering is not implemented, and Magician parity is outside this PR's scope.
- Document the restriction in the array guide and generated `sort()`/`rsort()` references through the shared catalog's documentation metadata.

## Other changes already in the branch

- Anchor Linux DWARF abbreviation and line-table offsets to this compilation unit's sections so linked libc debug sections do not redirect PHP source lookups; retain the existing Mach-O behavior and add a focused regression test.
- Document the Mixed comparator helpers and keep emitted assembly comments aligned.

## Validation

- CI passed on runtime-change head `5c7cbce91` across `macos-aarch64`, `linux-aarch64`, and `linux-x86_64`.
- The follow-up after that head changes only documentation, documentation metadata, and generated output. No runtime, test, CI, changelog, or version changes.
- Documentation checks passed: canonical `curl`-enabled exporter build, full docs regeneration (repeat run produces no further changes), 10 builtin contract documentation pipeline tests, 23 PHP comparison generator tests, builtin/site/boundary audits, and `git diff --check`.


Guikingone:
Second commit: a **different** unbounded leak, found while confirming the first one had not fixed it.

`__rt_mixed_from_value` increfs the child it boxes — right for a SHARED payload, one reference too many for a hash whose only reference is the creator's. `getdate`, `localtime` and `stat` hand their freshly built hash straight to the box and keep nothing, so the count starts at two and no release reaches zero.

| | before | after |
|---|---|---|
| `$x = getdate(); unset($x);` | `allocs=14 frees=1` | `14 / 14` |
| `$x = ["a"=>1]; unset($x);` (control) | `3 / 3` | `3 / 3` |
| `getdate()` × 20 in a loop | `301 / 41` | `301 / 301` |
| `localtime(0, true)` × 5 | `61 / 11` | `61 / 61` |

**Reading the release path found nothing, and could not** — it is correct at every step. The EIR emits the release; it lowers to `__rt_decref_mixed`; the tag ladder in `__rt_mixed_free_deep` drops tag 5 into `__rt_decref_any`; that dispatcher handles heap kind 3. The debugger answered it: two `__rt_decref_mixed` calls taking the cell to zero, then one `__rt_decref_hash`, and the hash sitting at refcount 2 throughout.

One near-miss worth flagging for review: `parse_url` returns a Mixed-boxed assoc array and is clean, which almost became *"so the leak is not general to boxed hashes"*. It does not use `__rt_hash_new` at all and builds its result another way — a false sibling, and the wrong thing to reason from.

Mutation-verified: keeping the creator's reference restores the leak exactly (460 live blocks over the test's forty calls). The test asserts `localtime` alongside `getdate` so a fix reaching only one cannot pass.

Guikingone:
The one red job on this PR — `PDO core live suites` — is **not caused by it**, and
the check that settles that is worth writing down: the same workflow is red on
`main` (`343ed111`), on the same test.

It is also not flaky. The value it reads back differs **every run**, which is the
signature of freed memory rather than of a slow database:

| run | `left` | expected |
|---|---|---|
| `main` @343ed111 | four replacement chars, then `8yT` | `00000` |
| this branch @43c8254a | a NUL byte, then `0000` | `00000` |

The cause is in `crates/elephc-pdo/src/lib.rs`:

```rust
fn store_cstr(cell: &'static Mutex<CString>, s: &str) -> *const c_char {
    let mut guard = lock_recover(cell);
    *guard = cstr;      // frees the previous CString
    guard.as_ptr()      // a pointer INTO it...
}                       // ...and the lock is released here
```

The pointer escapes the lock. `sqlstate_cell()` is a single `static`, and
`pg_round_trip` and `my_round_trip` run concurrently in one binary
(`cargo test -p elephc-pdo --lib`, no `--test-threads=1`), so one can replace the
`CString` between the other's store and its read. Twenty-three
`static C: OnceLock<Mutex<CString>>` cells share the pattern.

The fix that belongs here is already in the same file, twenty lines below:
`store_bytes` uses `COLDATA_CELL`, a **thread-local** `RefCell<Vec<u8>>`. A
per-thread FFI buffer removes the race without changing the C API.

I have deliberately not made that change in this PR. It is a different subsystem,
it touches twenty-three sites, and verifying it needs live PostgreSQL **and**
MySQL — which I cannot run here, so I would be shipping an unverified fix into
code I do not own. Reporting it seemed better than guessing at it.

Guikingone:
Applied in `a776dd7ebb`, with one honest caveat.

You are right that the exact `Ownership::Owned` test is narrower than the hazard:
a heap-valued `MaybeOwned` store carries the same risk, and the guard should read
`may_require_release()`.

**I could not build a case that leaks through the narrower check**, and would
rather say so than claim the change closes something I watched fail. Three shapes,
each thirty calls inside a loop, measured with heap debug:

| shape | heap |
|---|---|
| `$a = [...]; $b = $a;` | `allocs=121 frees=121` |
| `$a = getdate(); $b = $a;` | `allocs=451 frees=451` |
| alias behind a branch, then reassign | `allocs=91 frees=91` |

All clean before the change. The reason looks structural rather than lucky: a
callee that stores a `MaybeOwned` heap value has, in every shape I could write,
also stored an `Owned` one first — and that store already refuses the callee.

Widened anyway, for two reasons that do not depend on finding the case. Both
tests only fire on `Heap` values, so scalar callees inline exactly as before and
the cost is not measurable. And the two directions are not symmetric: being wrong
this way loses an inline, being wrong the other way loses memory on every
iteration — which is the arbitration `inline.rs` had already made for the
by-value-parameter guard sitting immediately above it.

nahime0:
I found three correctness issues and one repository-policy issue that should be addressed before merging:

1. **[P1] The inliner guard still misses owned strings** (`src/ir_passes/inline.rs:1103`)

   `callee_stores_a_container_local()` only rejects releasable `IrType::Heap(_)` values. `IrType::Str` is also refcounted/lifetime-tracked, so an owned string local lowered at function entry can still be spliced into a caller loop and overwritten without releasing the previous iteration's value.

   I reproduced this with a callee that assigns `getcwd()` to a local and returns `strlen($path)`, called 20 times from a loop under `--heap-debug`:

   - IR optimization enabled: `allocs=63 frees=44 live_blocks=19`
   - `--no-ir-opt`: `allocs=84 frees=84 live_blocks=0`

   The guard should cover every releasable refcounted storage type, not only `Heap(_)`, and this needs a string-local heap-debug regression test.

2. **[P1] Mixed `sort()` silently misorders nested arrays** (`src/codegen_support/runtime/arrays/php_compare_slots.rs:52`)

   The new Mixed sorting path delegates to `__rt_php_compare`, but that helper explicitly documents that arrays, objects, resources, and callables only rank above scalars and compare equal to one another (`src/codegen_support/runtime/compare/php_compare.rs:29-33`). Enabling the path for arbitrary Mixed values therefore exposes incorrect PHP behavior.

   Reproducer:

   ```php
   <?php
   $decoded = json_decode("[[2],[1],[]]", true);
   $values = [];
   foreach ($decoded as $value) {
       $values[] = $value;
   }
   sort($values);
   echo json_encode($values), "\n";
   ```

   PHP 8 prints `[[],[1],[2]]`; this branch prints `[[2],[1],[]]`. Container comparison needs to be implemented before accepting those Mixed tags, or the unsupported tags need an explicit diagnostic.

3. **[P2] The fresh-hash ownership fix is local and leaves equivalent producers leaking** (`src/codegen/lower_inst/builtins/system.rs:207`)

   The new helper correctly balances `getdate()`/`localtime()`, but duplicates ownership-transfer logic instead of routing through the existing owned-value boxer (`src/codegen_support/value_boxing.rs:206`). Other fresh tag-5 producers still call `__rt_mixed_from_value` without releasing the creator's reference:

   - `hrtime()` (`src/codegen_support/runtime/system/hrtime.rs:91`): 20 calls leave `live_blocks=20`.
   - `ob_get_status()` (`src/codegen/lower_inst/builtins/output_buffering.rs:580`): 20 calls leave `live_blocks=20`.

   This transfer should be centralized and covered with heap-debug tests for all fresh associative-array producers.

4. **[P3] The new emitter does not satisfy the repository's mandatory documentation/comment policy** (`src/codegen_support/runtime/arrays/php_compare_slots.rs:30`)

   `emit_one()` has no `///` function docblock, and several new `emitter.instruction(...)` calls (starting at line 38) have no aligned inline assembly comment. `scripts/check_asm_comments.py` currently passes because it checks alignment only when a comment is already present; it does not detect missing comments.

For validation, I ran the focused ownership tests (8/8), the new Mixed-sort test, both focused DWARF/debug-info tests, `cargo build --bin elephc`, and `git diff --check`; those all pass. The three reproductions above are additional cases not covered by the PR tests.

nahime0:
Follow-up for the remaining Mixed-boxed fresh-hash sites this PR did not touch: #807 (`ob_get_status`, hrtime array form, `json_decode`). Not a merge blocker. getdate/localtime are in scope here after `2abe9e9f`; getenv($name) is not that leak.

nahime0:
Optional follow-ups from the Grok merge-gate (docs SHOULD already landed in `8bd363df0`):

- #941 — real PHP Mixed `sort`/`rsort` container ordering (replace fatal)
- #942 — inliner: emit `Release` in loops instead of refusing
- #943 — Linux DWARF `addr2line` binary test

Already tracked elsewhere: #923, #924, #925, #938.

PR #880, opened 2026-09-02
feat: add monitoring contract and curl telemetry

nahime0:
## Summary

Adds first-class monitoring support for the curl bridge and introduces a typed gate that requires new bridges and bridge-backed I/O operations to declare their monitoring behavior explicitly.

This PR now covers the full curl monitoring contract: operation counts, measured network wait, inclusive and exclusive attribution, trace propagation, report and export integration, dormant-path behavior, generated builtin documentation, and contributor guidance.

## Contract and quality gates

- Added the shared `elephc-monitoring-contract` crate.
- Added mandatory `MonitoringPolicy` metadata to every bridge and typed runtime function descriptor.
- Added CI audits for:
  - bridges without a reviewed monitoring policy
  - bridge-backed runtime functions without a policy
  - I/O operations without typed event monitoring
  - network effects not classified as `IoKind::Network`
  - blocking I/O without `WaitPolicy::Measured`
  - infrastructure exemptions without a source-level justification
- Added `BLOCKING_IO` and `NETWORK_IO` EIR effects.
- Kept callback-capable curl drivers conservative for PHP-visible effects such as throws, output, and heap allocation.
- Documented the monitoring contract in `CONTRIBUTING.md`, including operation semantics, callback-time exclusion, active-window ownership, and dormant-path requirements.

## Curl monitoring

- Added network operation and wait telemetry for:
  - `curl_exec()`
  - multi-handle transfers
  - `curl_multi_select()`
  - `curl_upkeep()` wait time
- Easy transfers count once per `curl_exec()`.
- Multi transfers count once per easy-handle attachment. Reading a completion message no longer clears the count while the handle remains attached; remove and re-add starts a new counted attachment.
- `curl_upkeep()` records measured maintenance wait without incrementing transfer operations.
- `curl_exec()` subtracts PHP callback execution from measured network wait, so user CPU in write, header, read, progress, debug, or transfer-info callbacks is not reported as driver wait.
- Nested transfers retain separate callback timing scopes, and inactive nested transfers cannot leak timing into an active caller.
- Dormant multi execution no longer clones the attachment list. Trace headers are refreshed in place while holding the established easy-then-multi lock order, with no steady-state per-drive allocation.

## Trace propagation

- Added automatic W3C `traceparent` propagation during active captures.
- Explicit user headers win case-insensitively.
- Removing a user header allows automatic propagation to resume on the next transfer.
- Injected headers are not copied as user headers by `curl_copy_handle()`.
- Stale injected headers are removed when a reused handle runs outside the capture that introduced them.
- Added a real libcurl regression test for slist replacement, removal, bookkeeping, and user precedence through the production pre-transfer path.
- Documented why replacing and freeing the previous slist is safe for an attached multi handle before `curl_multi_perform()` starts walking it.

## Attribution, reports, and exports

- Added inclusive and exclusive exact counters:
  - `incl_network`
  - `excl_network`
  - `incl_network_wait`
  - `excl_network_wait`
- Network events now propagate through nested calls, recursion, coroutine state, and exception resynchronization with the same attribution model used by database counters.
- Added network metrics to:
  - exact monitoring tables
  - sampled probe summaries
  - interactive HTML reports
  - distributed trace stitching
  - Graphviz output
  - OTLP span attributes
  - Prometheus gauges
  - performance assertions via `network` and `network_wait_ms`
- Prometheus now exports:
  - `elephc_network_operations_per_request`
  - `elephc_network_wait_seconds_per_request`
- Run totals and assertion metrics use exclusive values to avoid double-counting nested frames, while graphs expose both inclusive and exclusive values.

## Probe and PDO behavior

- Database and network telemetry use separate runtime slots, so HTTP work cannot affect query budgets or N+1 diagnostics.
- Event consumers retain ownership of their capture-window gate. Bridge-side exact timing stays dormant outside an active exact or remote-probe window.
- The shared helper fixes PDO wait timing in remote-probe windows: those windows previously recorded database operations with zero driver wait, and now record the measured wait duration. Exact PDO capture semantics remain unchanged.
- `probe_io_summary` intentionally continues into allocation and network summaries even when there are no database rows, so non-database activity is not hidden.
- Updated the probe shared-region layout comment after widening the event table.

## Compatibility

- Exact logs written with the earlier `network_ops` and `network_wait` fields still parse.
- Saved JSON graphs using the earlier field names deserialize through aliases and restore the direct value as the minimum valid inclusive value.
- Added regression tests for both legacy formats.

## Documentation

- Updated curl, profiling, CLI, and contributor documentation.
- Regenerated the builtin registry and internal pages after restoring conservative curl runtime effects.
- Validated generated docs, site compatibility, and the typed EIR boundary audit.

## Verification

- `cargo build -p elephc --features curl`
- `cargo check -p elephc-curl`
- `cargo test -p elephc-monitoring-contract`
- `cargo test -p elephc-curl timing_tests`
- Real libcurl tests:
  - `perform_refreshes_the_real_traceparent_slist`
  - `multi_perform_runs_a_transfer_and_queues_its_completion`
- `cargo test -p elephc-instr network_metrics`
- `cargo test -p elephc-probe io_events_are_counted_exactly_per_route`
- `cargo test -p elephc-probe the_first_ask_opens_a_fresh_event_window`
- `cargo test --bin elephc every_bridge_declares_a_monitoring_policy`
- `cargo test --bin elephc every_runtime_lowering_has_a_logical_signature`
- `cargo test --bin elephc curl_runtime_effects_mark_network_boundaries`
- Focused parsing, saved-profile compatibility, assertion, Prometheus, HTML, trace, and stitching tests
- `cargo build --example gen_builtins --features curl`
- `python3 scripts/docs/extract_builtins.py --render --force`
- `python3 scripts/docs/audit_builtins.py`
- `python3 scripts/docs/elephc_builtins/validate_site_compat.py`
- `python3 scripts/audit_builtin_eir_boundary.py --enforce-target-architecture`
- `git diff --check`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -161,9 +161,20 @@ by EIR lowering and `src/codegen/`.
 Optional, heavyweight functionality that is naturally expressed with Rust
 libraries lives in a workspace crate under `crates/elephc-<name>/`, compiled as a
 `staticlib` and linked into the user program on demand. Every such bridge is
-described by one entry in the `BRIDGES` table in `src/linker.rs`; `link()` and the
+described by one entry in the `BRIDGES` table in `src/linker/bridges.rs`; `link()` and the
 discovery/auto-build helpers are fully table-driven, so adding a bridge is a
 single table entry rather than new linker logic.
+
+Every bridge entry must declare a reviewed `MonitoringPolicy`. Use generic
+timing when ordinary function timing is sufficient, typed I/O policy when the
+bridge performs database or network work, and an infrastructure policy with a
+nonempty reason only for monitor/runtime plumbing. The bridge-catalog gate
+rejects an unspecified policy. Bridge-backed `RuntimeFnId` values must also
+declare their operation policy, and any runtime operation carrying
+`BLOCKING_IO` or `NETWORK_IO` effects must use evented monitoring. I/O bridges
+share the dependency-neutral `elephc-monitoring-contract` helper and distinct
+database/network runtime slots, which must remain dormant outside an active
+capture.
 
 A bridge is linked when its `lib_name` appears in `extra_link_libs`. That set is
 populated automatically by feature detection: the type checker records a needed
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -609,7 +609,10 @@ cannot contaminate each other. The event consumer owns its capture-window gate,
 and must publish an active-window callback when that state is not represented
 by the exact-capture word. Bridge-side timing must avoid clock reads outside
 either active window, while trace propagation stays scoped to an exact trace
-context. Steady-state dormant paths must not allocate per operation.
+context. Count only actual external operations, not maintenance calls such as
+connection upkeep. Measured wait must exclude time spent in user callbacks so
+network-bound diagnostics describe the driver boundary rather than PHP work.
+Steady-state dormant paths must not allocate per operation.
 
 Set `whole_archive: true` only when the staticlib has link-time side effects that
 must survive (e.g. a provider registration) or owns the program entry point (like
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -406,6 +406,13 @@ enum), implement it under `src/codegen/lower_inst/runtime_functions/` or
 `runtime_calls.rs`, and keep every supported ABI in the same path. If it needs a
 runtime routine, add it under `src/codegen_support/runtime/<category>/`.
 
+Every bridge-backed `RuntimeFnId` must also return a reviewed
+`MonitoringPolicy` from `RuntimeFnId::monitoring_policy()`. Operations marked
+with `BLOCKING_IO` or `NETWORK_IO` effects must use an evented I/O policy. The
+registry test rejects an unspecified decision, so a new runtime boundary cannot
+land until its generic timing, typed events, wait measurement and trace-context
+behavior have been reviewed explicitly.
+
 For a builtin that is naturally expressed as reusable EIR operations, use
 `BuiltinLowering::Eir(lower)`. The hook receives only `BuiltinLoweringContext` and a
 `NormalizedBuiltinCall`; it may emit typed EIR values/runtime calls, but must not
@@ -569,7 +576,7 @@ Every supported target must build and link the crate: `macos-aarch64`,
 artifacts are libraries cross-compiled from macOS. A bridge that only works on
 one target is not acceptable (see the supported-target policy in `CLAUDE.md`).
 
-### 3. Register the bridge in `BRIDGES` (`src/linker.rs`)
+### 3. Register the bridge in `BRIDGES` (`src/linker/bridges.rs`)
 
 Add one `BridgeStaticlib` entry. This is the only linker change required —
 discovery, on-demand build, search paths, whole-archiving, and macOS frameworks
@@ -584,8 +591,24 @@ BridgeStaticlib {
     whole_archive: false,               // true if link-time side effects / owns entry
     macos_frameworks: &[],              // transitive native deps' frameworks
     needs_libdl: true,                  // Rust runtime/unwinder needs -ldl on Linux
+    php_extension: None,                // canonical PHP extension, when distinct
+    monitoring: MonitoringPolicy::GenericTiming,
 },
 ```
+
+The `monitoring` field is mandatory. Use `GenericTiming` when function timing
+fully describes the bridge, `Io { ... }` when it must publish a distinct I/O
+operation or wait dimension, and `Infrastructure { reason }` only for monitor or
+runtime plumbing that is not a PHP-visible work boundary. The bridge-catalog
+test rejects `Unspecified` and empty infrastructure reasons.
+
+I/O bridges should depend on `elephc-monitoring-contract` and call its
+`EventHooks` helper through category-specific runtime slots. Keep database and
+network slots separate so query budgets, N+1 analysis and network budgets
+cannot contaminate each other. The event consumer owns its capture-window gate,
+while bridge-side timing and trace propagation must stay dormant outside an
+active exact capture, including avoiding clock reads and steady-state
+per-operation allocations.
 
 Set `whole_archive: true` only when the staticlib has link-time side effects that
 must survive (e.g. a provider registration) or owns the program entry point (like
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -606,9 +606,10 @@ I/O bridges should depend on `elephc-monitoring-contract` and call its
 `EventHooks` helper through category-specific runtime slots. Keep database and
 network slots separate so query budgets, N+1 analysis and network budgets
 cannot contaminate each other. The event consumer owns its capture-window gate,
-while bridge-side timing and trace propagation must stay dormant outside an
-active exact capture, including avoiding clock reads and steady-state
-per-operation allocations.
+and must publish an active-window callback when that state is not represented
+by the exact-capture word. Bridge-side timing must avoid clock reads outside
+either active window, while trace propagation stays scoped to an exact trace
+context. Steady-state dormant paths must not allocate per operation.
 
 Set `whole_archive: true` only when the staticlib has link-time side effects that
 must survive (e.g. a provider registration) or owns the program entry point (like
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -674,6 +674,7 @@ dependencies = [
  "elephc-curl",
  "elephc-image",
  "elephc-magician",
+ "elephc-monitoring-contract",
  "elephc-pdo",
  "elephc-phar",
  "elephc-probe",
@@ -733,6 +734,7 @@ dependencies = [
 name = "elephc-curl"
 version = "0.1.0"
 dependencies = [
+ "elephc-monitoring-contract",
  "libc",
  "serde_json",
 ]
@@ -775,11 +777,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "elephc-monitoring-contract"
+version = "0.1.0"
+
+[[package]]
 name = "elephc-pdo"
 version = "0.1.0"
 dependencies = [
  "bytes",
  "chrono",
+ "elephc-monitoring-contract",
  "futures-util",
  "libloading",
  "libpq",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,8 +28,8 @@ categories = ["compilers", "command-line-utilities"]
 # see crates/elephc-curl/build.rs for how its own gated tests opt into a real
 # link.
 [workspace]
-members = [".", "crates/elephc-builtin-contract", "crates/elephc-tls", "crates/elephc-pdo", "crates/elephc-crypto", "crates/elephc-bcmath", "crates/elephc-iconv", "crates/elephc-phar", "crates/elephc-tz", "crates/elephc-probe", "crates/elephc-instr", "crates/elephc-image", "crates/elephc-web", "crates/elephc-magician", "crates/elephc-curl"]
-default-members = [".", "crates/elephc-builtin-contract", "crates/elephc-tls", "crates/elephc-pdo", "crates/elephc-crypto", "crates/elephc-bcmath", "crates/elephc-iconv", "crates/elephc-phar", "crates/elephc-tz", "crates/elephc-probe", "crates/elephc-instr", "crates/elephc-image", "crates/elephc-web", "crates/elephc-magician", "crates/elephc-curl"]
+members = [".", "crates/elephc-builtin-contract", "crates/elephc-monitoring-contract", "crates/elephc-tls", "crates/elephc-pdo", "crates/elephc-crypto", "crates/elephc-bcmath", "crates/elephc-iconv", "crates/elephc-phar", "crates/elephc-tz", "crates/elephc-probe", "crates/elephc-instr", "crates/elephc-image", "crates/elephc-web", "crates/elephc-magician", "crates/elephc-curl"]
+default-members = [".", "crates/elephc-builtin-contract", "crates/elephc-monitoring-contract", "crates/elephc-tls", "crates/elephc-pdo", "crates/elephc-crypto", "crates/elephc-bcmath", "crates/elephc-iconv", "crates/elephc-phar", "crates/elephc-tz", "crates/elephc-probe", "crates/elephc-instr", "crates/elephc-image", "crates/elephc-web", "crates/elephc-magician", "crates/elephc-curl"]
 resolver = "2"
 
 [[bin]]
@@ -45,6 +45,7 @@ path = "tools/gen_builtins.rs"
 
 [dependencies]
 elephc-builtin-contract = { path = "crates/elephc-builtin-contract" }
+elephc-monitoring-contract = { path = "crates/elephc-monitoring-contract" }
 inventory = "0.3"
 libc = "0.2"
 # Grows the stack on demand inside the recursive-descent parser so deeply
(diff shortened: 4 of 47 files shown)


Guikingone (commented):
## Review — monitoring contract and curl telemetry (`2b7cdda35`)

The typed gate is the good idea here, and it is built the right way round: `MonitoringPolicy` is a
required field on `BridgeStaticlib` and on `RuntimeFnDescriptor`, so a new bridge cannot compile
without naming a decision, and `Unspecified` is then rejected by the audit rather than being a
usable default. The registry audit going further — `NETWORK_IO` must be classified as
`IoKind::Network`, `BLOCKING_IO` must be `WaitPolicy::Measured` — turns the effect bits and the
policy into each other's check, and `Infrastructure { reason }` requiring a non-empty string means
the escape hatch has to be argued for in the source. Keeping network events in their own runtime
slots rather than folding them into the DB ones is also right, for the reason stated next to the
slot: an HTTP call must not move a query budget or an N+1 diagnosis.

I checked the wire-level claims against a fake HTTP server that logs every request header, and the
load-bearing ones hold. During an active capture the `traceparent` is injected; a user header wins
with any letter case (`TrAcEpArEnT: …` suppressed it); removing that user header lets injection
resume on the next transfer on the same handle; a dormant run injects nothing at all;
`curl_copy_handle()` carries the user's own header and does not duplicate the injected one;
`curl_multi_select()` reports blocked duration and no operation; easy transfers count once per
`curl_exec()`. Old logs and baselines still parse — `instr_field()` yields 0 for an absent key and
the new `GraphNode` fields carry `#[serde(default)]`.

Two things I would fix before this lands, three worth a decision, and some smaller items.

### Method

About 25 differential programs against a local logging HTTP server, php 8.5.6 (`php -n`) where php
has an opinion, `elephc monitor` on `--with-curl --with-monitoring` builds, and `--emit-ir` for the
effect metadata. Grades: **✅ proven** — I ran it and the output is in the finding; **📄 source** —
read verbatim in the code under review, quoted, not executed. An adversarial jury of three agents
from two model families was asked to refute these rather than grade them; it cost one of my
findings its severity and gained a better one, both recorded below.

---

## Should-fix

### 1 — a multi transfer is counted more than once per handle ✅ proven

Three easy handles on one multi, driven by the loop from php.net's `curl_multi_info_read` page,
i.e. `curl_multi_info_read()` **inside** the loop:

```php
$running = 1;
while ($running > 0 && $loops < 500) {
    curl_multi_exec($mh, $running);
    if ($running > 0) { curl_multi_select($mh, 1.0); }
    $info = curl_multi_info_read($mh);
    while ($info !== false) { $messages++; $info = curl_multi_info_read($mh); }
    $loops++;
}
```

```
server received:  3 requests
elephc monitor:   curl_multi_exec  calls 8  …  network 5      (three runs, same answer)
```

`docs/php/curl.md` says "Multi transfers count once per attached easy handle". Three handles are
counted five times, and the new gate turns that into a false build failure:

```
$ elephc monitor ./e2c --assert 'network:*<=3'
assertions - 0 passed, 1 failed
  [FAIL] network:*<=3 (actual 5)          # the program made exactly 3 requests
```

Moving `curl_multi_info_read()` **after** the loop gives the correct `network 3`, and a **single**
handle in the interleaved loop also gives `network 1` — so the trigger is precise: a handle that
completes and is popped by `info_read` while other handles keep the loop running.
`elephc_curl_multi_info_read` does `entry.monitored_easy.remove(&easy_id)` and the next
`elephc_curl_multi_perform` re-inserts that still-attached id and counts it again.

The field's rationale — "Completion or removal clears the id so a later transfer is counted
independently" — does not hold for completion: libcurl leaves a finished handle attached in its
done state and `curl_multi_add_handle` on it returns `CURLM_ADDED_ALREADY`, so the only route to a
new transfer is remove-then-add, and `elephc_curl_multi_remove` already clears the id.

**Fix:** drop the `remove` in `info_read`; keep the one in `multi_remove`.

### 2 — PHP callback time is reported as outgoing network wait ✅ proven

`hooks.timed(|| easy::perform(curl))` brackets the whole libcurl transfer, and libcurl runs the PHP
`CURLOPT_WRITEFUNCTION` / `HEADERFUNCTION` / `PROGRESSFUNCTION` callbacks inside it. So user CPU
spent in a callback is billed as network wait — and counted twice, because the callback's own
frames are attributed normally as well.

Same program, same total work; only the placement of the PHP loop differs.

```php
// A: the work runs inside the write callback
curl_setopt($ch, CURLOPT_WRITEFUNCTION, function ($handle, string $data) {
    burn(20000000); return strlen($data);
});
curl_exec($ch);
```
```
curl_exec  …  network 1  network-wait 487.7 ms
burn       …  self 484.1 ms
• the run is network-bound: 99% of it (487.7 ms) is outgoing network wait - curl_exec blocks longest
```

```php
// B: the identical loop, moved after the transfer
curl_exec($ch);
burn(20000000);
```
```
burn       …  self 452.1 ms
curl_exec  …  network 1  network-wait 1.6 ms
(no network-bound recommendation)
```

The transfer is to `127.0.0.1` and takes about 1.6 ms in both. In A the report says 99% of the run
is outgoing network wait and points at `curl_exec`; the answer is `burn`. Guzzle installs a write
callback on every request, so this is the ordinary shape, not a contrived one.

**Fix:** keep the callback out of the measured span — stop and restart the wait clock across the
callback boundary the way the bridge already tracks `callback_threw`, or take the transfer's
duration from libcurl's own timing rather than wall time around `perform()`.

---

## Worth a decision

### 3 — `curl_upkeep()` is counted as a network operation ✅ proven

```php
curl_exec($ch);
curl_upkeep($ch); curl_upkeep($ch); curl_upkeep($ch);
```
```
server received:  1 request
elephc monitor:   curl_exec     calls 1  …  network 1  network-wait 1.8 ms
                  curl_upkeep   calls 3  …  network 3  network-wait 625 ns
```

`curl_easy_upkeep` performs no transfer — at most an HTTP/2 PING on an idle connection — and the
documented counting rules cover only `curl_exec()`, multi handles and `curl_multi_select()`. A
client that calls it on a pool will have its "outgoing network operations" dominated by things that
fetched nothing. Keep the timed wait; either drop the operation count or document it.

### 4 — `curl_exec` and `curl_multi_exec` no longer declare that they can throw ✅ proven (metadata)

`--emit-ir`:

```
runtime.__elephc_curl_easy_perform ; effects: reads_heap, writes_heap, reads_process,
                                     writes_process, may_warn, blocking_io, network_io
runtime.__elephc_curl_multi_exec   ; effects: … may_warn, network_io
```

Before this PR these used the catch-all arm (16 bits); the explicit sets carry 7 and 6. Dropped:
`MAY_THROW`, `OUTPUT`, `ALLOC_HEAP`, `ALLOC_CONCAT`, `MAY_FATAL`, `READS_GLOBAL`, `READS_FS`,
`WRITES_FS`, `MAY_DEOPT`.

Three of those look wrong for these two operations. They run user PHP callbacks and your own bridge
comment says so — "The runtime re-raises the pending throwable as soon as this returns" — that is
`MAY_THROW`. Without `CURLOPT_RETURNTRANSFER`, `curl_exec` writes the response body to stdout: that
is `OUTPUT`. With it, it returns a freshly allocated PHP string: that is `ALLOC_HEAP`. For
`curl_easy_upkeep` and `curl_multi_select` the omission is correct — non-rethrowing forwarders, no
callbacks.

`MAY_THROW` is not decorative. `dead_store.rs` documents what it protects:

> a `may_throw` instruction inside a `try` must conservatively make every eligible slot live at
> that point. Otherwise a later store on the normal path can incorrectly make the value observed by
> the catch handler look dead.

**Why it does not bite today.** A juror argued this should be worse than I graded it:
`optimize/exception_flow.rs` falls back to `Effect::from_eir(descriptor.effects).may_throw`, and an
unreachable catch clause is then deleted by `optimize/control/dce/tries.rs`. It does not fire, and
the reason is the prelude: `curl_exec` is a PHP function (`src/curl_prelude.rs:1120`) wrapping
`__elephc_curl_easy_perform`, so the AST analysis summarises *its* body — which reads
`$handle->__elephc_handle`, a declared-slot property read carrying `MAY_THROW` — rather than the
runtime descriptor. The minimal shape confirms it:

```php
try { curl_exec($ch); } catch (Throwable $e) { echo "caught: ", get_class($e), "\n"; }
// php and elephc both: caught: RuntimeException
```

So: provably wrong metadata on two live passes, currently shielded by an indirection that nothing
in this PR guarantees will stay. The narrowing is worth keeping — it is a real improvement over
"assume everything" — it just needs `MAY_THROW | OUTPUT | ALLOC_HEAP` back on the two operations
that invoke user callbacks.

### 5 — network has no inclusive counter ✅ proven, and downgraded by the jury

Every node except the curl builtin reports `network 0`, so a per-function budget cannot fail:

```
fetch_in_function   …  network 0     ← this function performs one HTTP request
curl_exec           …  calls 2  …  network 2  network-wait 2.9 ms

  [PASS] network:fetch_in_function<=0 (actual 0)
  [FAIL] network:*<=1 (actual 2)
```

**I first filed this as a should-fix and as a contrast with the DB dimension. The jury refuted the
contrast, and it was right.** `PDO::exec` and `PDOStatement::execute` are prelude PHP functions with
their own frames too, `queries` reads `io_exclusive`, and your own docs show the DB recommendation
naming a prelude frame — "*the run is I/O-bound … PDO::exec blocks longest*"
(`docs/beyond-php/profiling.md:850-855`). `--assert queries:fetch_db<=0` passes just the same, and
both metrics are documented as "…it issues itself". This is not a regression.

What genuinely differs is narrower: DB carries an `incl_io` / `incl_wait` pair that the HTML graph
and its tooltips use, and network has no inclusive counterpart — `note_network` credits
`stack.last()` with no propagation into enclosing frames, so the new HTML dimension is
exclusive-only where every other one offers both. Adding `incl_network` / `incl_network_wait` would
make it behave like the rest of the table.

---

## Smaller items

- **The PDO refactor does change when driver wait is measured.** 📄 source. `timed()` used to gate
  on `asked()` alone; the shared helper also consults `elephc_monitor_event_active_fn`, which
  `emit_probe_init` fills with `elephc_probe_event_active`. Exact captures are unaffected (`active`
  short-circuits); what changes is the **remote-probe** window, where DB wait was reported as
  `wait_ns=0` and now is measured. That is the improvement the second commit is named after — but
  the description says the refactor is "without changing its existing metrics", and a reviewer
  reading that will not go looking.
- **`--prometheus` did not get the new dimension.** 📄 source. `prometheus_text` still emits only
  `elephc_requests_total`, `elephc_request_duration_seconds` and `elephc_queries_per_request`,
  while OTLP gained `elephc.network_operations` and `elephc.network_wait_ns`. The docs do not
  promise it, so this is a gap rather than a false claim.
- **The multi path allocates on every drive call, dormant or not.** 📄 source.
  `elephc_curl_multi_perform` now does `entry.attached.clone()` and takes a second lock on every
  `curl_multi_exec()` so the traceparent refresh can run; before, it read `entry.multi` under one
  lock. `docs/php/curl.md` says automatic propagation "performs no allocation or clock read in
  steady-state dormant execution".
- **`probe_io_summary` lost its early return.** 📄 source. With no DB rows it used to return an
  empty string; it now falls through to the allocation and network summaries, so a capture with
  allocations but no database activity prints a block it previously did not. Almost certainly
  intended, but it is a behaviour change absent from the description.
- **`Database —` became `Database -`.** 📄 source. Nothing needs it, no test pinned it, and the
  `allocations —` header printed directly beneath in the same summary keeps its em dash.
- **The traceparent tests exercise a test-only twin of the production path.** 📄 source.
  `effective_http_headers` is `#[cfg(test)]` and rebuilds the header list itself; the three tests in
  `monitoring_header_tests` never call `refresh_monitor_traceparent`. Precedence is genuinely
  covered — both go through `automatic_traceparent` — but the slist rebuild, the free of the
  previous list, the `applied_traceparent` bookkeeping and the removal path have no test, and those
  are the parts that touch libcurl-owned memory.
- **A region comment in the probe is now stale.** 📄 source. Widening `EVENT_WORDS` from 2 to 4
  moved the control word and the replay table, which the comment claiming each new area "is
  appended LAST so adding it moves no existing offset" says cannot happen.

---

## Jury additions I examined and did not adopt

- **"`refresh_monitor_traceparent` frees a `curl_slist` libcurl may still be using."** It rebuilds
  and frees the previous list on a handle that can be attached to a running multi, and libcurl does
  not copy that list. But the `setopt` swapping libcurl's pointer happens *before* the free, and
  libcurl only walks the list synchronously inside `perform`. I also drove two in-flight transfers
  against a two-second endpoint and rotated `ELEPHC_TRACEPARENT` on every one of 100 drive calls:
  no crash, no corrupted header. Worth a source comment saying why it is safe; not a defect.
- **"`EVENT_WORDS` 2 → 4 is unguarded by `CONTROL_MAGIC`, so a newer monitor misreads an older
  process."** Nothing in `src/` computes `event_word` offsets: `event_report()` renders its lines
  inside the target process and the monitor parses `elephc-probe-*` text. There is no cross-process
  reader of that layout.
- **"The catch clause of `try { curl_exec(…) } catch { }` can be deleted outright."** See finding 4.

## Not this PR

- **`putenv('NAME')` with no `=` does not unset.** php then answers `getenv('NAME') === false`;
  elephc keeps the previous value. Measured with no curl and no monitoring in the program. ✅ proven.
- **`$running = null; curl_multi_exec($mh, $running);` fails to compile** with
  `EIR backend error: unsupported EIR backend feature: icmp for PHP type Void`, positionless, on a
  program php runs. The by-reference out-parameter left null; nothing to do with monitoring. I did
  not build the merge base, so the attribution is ❓ unverified.

## Mergeability

`MERGEABLE / CLEAN`, 145 checks pass and one skipped. The branch is 24 commits behind `origin/main`
as I write this and the merge is clean — worth a merge-up before landing, since #879's control-flow
normalization is among those 24 and this PR touches EIR effects.

## Verification and caveats

macOS ARM64 only; the Linux and x86_64 shards are CI's word, not mine. Two documented claims I did
not test and am not asserting either way: the dormant-execution cost claim beyond the multi-path
allocation noted above, and "removed when a reused handle runs outside the capture that introduced
it" — a CLI capture spans the whole process and `putenv` cannot clear the variable to simulate the
transition. Automatic propagation in `--web` mode is likewise untested here:
`elephc_instr_trace_begin` is reached only through the web bridge, so in a CLI run
`ELEPHC_TRACEPARENT` is whatever the environment already held, which is how I drove the injection
tests above.

nahime0:
Addressed the monitoring findings in `5d6351372`.

- Multi operations now remain counted for the lifetime of an attachment. `curl_multi_info_read()` no longer makes a completed but still-attached handle count again, and the real libcurl regression covers completion, another drive call, and removal.
- Easy-transfer wait subtracts time spent in PHP callbacks. Timing scopes are nested, dormant when monitoring is inactive, and covered independently.
- `curl_upkeep()` keeps measured maintenance wait but no longer increments transfer operations.
- `CurlEasyPerform` and `CurlMultiExec` restore conservative callback-capable effects, including throw, output, and allocation behavior, while retaining their network and blocking distinctions.
- Exact network operation and wait metrics now have inclusive and exclusive attribution, including recursion and exception resynchronization. Totals and budgets continue to use exclusive values.
- Prometheus exports mean network operations and network-wait seconds per request.
- The dormant multi path no longer clones `attached`; trace headers are refreshed in place with the established lock order.
- A native test now exercises the production libcurl slist path for injection, removal, bookkeeping, and user-header precedence.
- The probe region comment and attached-handle slist safety comment are updated.
- The sampled summary behavior and PDO remote-probe wait change are now explicit in the PR description. Summary punctuation is consistent.
- Legacy exact logs and saved JSON profiles remain supported and have regression tests.
- `CONTRIBUTING.md`, curl docs, profiling docs, CLI docs, generated builtin pages, and the builtin registry are updated.

The focused Rust tests, native libcurl regressions, policy gates, generated-doc workflow, site compatibility validator, EIR boundary audit, build, formatting, and `git diff --check` all pass locally.

nahime0:
Merging, can't find blocking issues

PR #907, opened 2026-09-06
fix: make bare putenv names unset variables

### LINKED ISSUE #891: putenv('NAME') without '=' does not unset the variable
## Context

Noted by Guillaume on #880 as **not this PR** (proven with no curl/monitoring in the program).

## Bug

PHP: `putenv('NAME')` with no `=` unsets; then `getenv('NAME') === false`.

elephc: keeps the previous value.

## Repro shape

```php
putenv('FOO=bar');
putenv('FOO'); // should unset
var_dump(getenv('FOO')); // PHP: false; elephc: still "bar"
```

Related: #880 (out of scope there). Distinct from #806 (eval `getenv` of an unset name answering `""`).

be-student:
## Summary

- treat a `putenv("NAME")` argument without `=` as PHP's unset form
- emit `unsetenv()` on both AArch64 and x86_64 while retaining the persistent `putenv()` buffer for assignments
- document the set/remove behavior and add a compatibility regression

## Validation

- `cargo test --test codegen_tests test_putenv -- --nocapture` (Rust 1.95.0; 2 passed)
- `cargo build` (Rust 1.95.0; no warnings)
- generated `linux-x86_64` assembly compiled with `clang -target x86_64-linux-gnu`
- `./scripts/check_asm_comments.py src/codegen/lower_inst/builtins/system.rs`
- `git diff --check`

The regression failed before the implementation with `set:failed:present` and now produces `set:unset:missing`.

Closes #891.

## AI assistance

AI tools were used to inspect the lowering paths, draft the implementation and tests, and prepare this description. I reviewed the changes and validated the behavior with the checks above.


nahime0:
lgtm, merge

Issue #685, opened 2026-08-08
new $class(...$assoc) with a string-keyed spread fails EIR validation


nahime0:
## Summary

`new $class(...$args)` fails to compile when `$args` is a **string-keyed** array held in a
variable. The same call with a literal string-keyed array, with named arguments written out,
or with a statically known class name all work.

## Repro

```php
<?php
class P {
    public function __construct($a = 'x', $b = 'y') { echo "$a/$b|"; }
}
$c = "P";
$named = ["b" => 8];
new $c(...$named);
```

PHP 8.4: `x/8|`

elephc:

```
EIR lowering error: EIR validation failed: OperandTypeMismatch {
    inst: InstId(18), operand: ValueId(8),
    expected: "Heap(Array)", actual: Heap(Hash)
}
```

Reproduced on pristine `main` (`a7a0e090c`) in a clean worktree, so this is not specific to
any feature branch.

## Perimeter

| Call | Result |
| --- | --- |
| `$c = "P"; $n = ["b" => 8]; new $c(...$n);` | **EIR validation failure** |
| `$c = "P"; new $c(b: 8);` | `x/8` — correct |
| `$n = ["b" => 8]; new P(...$n);` | `x/8` — correct |
| `$c = "P"; $n = [4, 5]; new $c(...$n);` | `4/5` — correct |

So it needs all three together: a dynamic class name, a spread, and a **string-keyed**
spread operand held in a variable.

## Cause

`lower_new_dynamic` routes a call carrying a spread through
`lower_untyped_descriptor_invoker_arg_container`. That helper returns a **hash** container
when the arguments are named and an **indexed** container otherwise
(`src/ir_lower/expr/descriptor_calls.rs`). A spread of a variable is not "named arguments"
as far as that check is concerned, so the indexed container is chosen and the string-keyed
source is spread into an indexed array — which is where the `Heap(Array)` vs `Heap(Hash)`
mismatch is raised.

The literal case works because the spread is expanded into real named arguments before
lowering; a variable operand cannot be expanded statically, so it survives as a spread.

## Suggested direction

The container choice should follow the **runtime shape of the spread operand**, not just the
syntactic presence of named arguments: a spread whose operand is a hash has to build the hash
container, so the keys reach the constructor as named arguments. A spread whose operand type
is not statically known needs a runtime decision or an explicit diagnostic — silently
choosing the indexed container would drop the keys and bind the values positionally, which is
a wrong answer rather than an error.

[...]

PR #913, opened 2026-09-07
feat(xml): complete ext/xml and ext/xmlwriter in AOT and eval() on catalog libxml2

### LINKED ISSUE #930: ci: xml iOS compile/link (catalog libxml2)
## Context
Follow-up from Grok review of #913. SHOULD; deferred for merge.

## Problem
Catalog + lock include `ios-arm64` / `ios-sim-arm64` for libxml2, but CI never builds those artifacts (`managed-native-smoke` is macOS+Linux only). Curl already has an iOS compile/link job.

## Expected
Add `ci: xml-ios-compile-link` (and/or iOS managed-native smoke for libxml2), matching curl.

## Related
- Parent: #913

### LINKED ISSUE #931: docs: sync AGENTS.md for --with-xml / second catalog-native bridge
## Context
Follow-up from Grok review of #913. SHOULD; deferred for merge.

## Problem
`AGENTS.md` still says curl is the only managed-native bridge and omits `--with-xml`.

## Expected
Document the second catalog-native bridge (`libxml2` / `--with-xml`) alongside curl.

## Related
- Parent: #913

nahime0:
Implements everything `docs/php/compatibility.md` attributes to the `xml` and `xmlwriter` PHP modules: the 22 `xml_*` functions with `XMLParser`, the 42 `xmlwriter_*` functions with `XMLWriter`, and the 28 `XML_*` constants, on both the native compiler and Magician `eval()`. The engine is libxml2 2.15.3 itself, built from the pinned upstream tarball by the curated native catalog, so DOM, SimpleXML, XMLReader and XPath can later share the same package. Behavior is pinned against PHP 8.5.10 / libxml2 2.15.3 with php-src `ext/xml` (`xml.c`, `compat.c`), `ext/xmlwriter` and `ext/libxml` as the reference.

## Commits

1. **feat(native): add the libxml2 2.15.3 catalog package** — `libxml2` joins the curated catalog (GNOME 2.15.3 release, static, platform iconv, no python/zlib/icu/modules) together with `.tar.xz` source support (format derived from the pinned URL, xz inflated by the pure-Rust `lzma-rs` into a size-bounded temporary, then the shared tar entry loop and bounds). The recipe compiles the Elephc-owned `libxml2_shim.c` against the freshly built headers and archives it as `libelephc_libxml2_shim.a`: versioned `elephc_libxml2_v1_*` entry points expose php-src's `compat.c` view of the parser context (input cursor, positions, errNo, subset/content state, entity lookups, external-resource refusal) so the Rust side never declares libxml2 struct layouts. `examples/xml` carries the manifest and lock CI materializes from.
2. **feat(linker): require managed libxml2 for the xml bridge** — a planned `elephc_xml` link requires the catalog package `libxml2` (like `elephc_curl` → `curl`), the resolver names the surface (`xml support requires managed native package libxml2`), `BridgeStaticlib` gains `apple_libraries` (system libraries linked by name on Apple targets only: libxml2's encoding handlers need the SDK's libiconv there, glibc has iconv in libc), and the release packaging probe declares libxml2 before `--with-xml`.
3. **feat(xml): add the elephc-xml bridge crate** — `staticlib` + `rlib` whose parser drives libxml2's push parser through a SAX table reproducing `compat.c` (qualified names, default-handler tag text, entity routing, per-event position snapshots, the external-entity stop) and whose writer is libxml2's `xmlwriter` over an `xmlOutputBufferCreateIO` sink like php-src's memory writer; libxml2's opaque public API is declared without linking (the curl pattern), behind a panic-free `elephc_xml_*` C ABI. Every SAX callback runs under `catch_unwind` (a panic halts the parser with `XML_ERROR_NO_MEMORY` instead of aborting at the `extern "C"` boundary). `build.rs` links the artifact and compiles the libxml2-calling unit tests in when `ELEPHC_XML_LIBXML2_LIB_DIR` is set; the parser corpus fixtures are reproducible (`cargo test -p elephc-xml --lib -- regen --ignored` regenerates the PHP probe scripts from the replay tables and re-checks the recorded outputs with `php`). Registered as the `elephc_xml` bridge (`--with-xml`) in the workspace, nextest archive, CI/nightly/release packing (including an iOS compile-and-link job for xml mirroring curl's) and the Linux test scripts.
4. **feat(contracts): catalog the xml and xmlwriter surface** — `catalog_xml.rs` (64 functions), the two classes on the prelude route, 28 constants, the bridge requirement, eval-support and count gates.
5. **feat(builtins): add the xml registry builtins** — `xml_parse_into_struct()` (write-only by-reference outputs need a checker hook) and the nine `xml_set_*_handler()` setters, whose check hook types an unannotated handler closure's parameters from the SAX event and specializes statically named handlers (`'on_element'`, `[$obj, 'method']`, first-class callables) the same way. Lowering is one call to the prelude twin (`__elephc_xml_*`) through the new `BuiltinLoweringContext::emit_user_call` / `store_operand_local` primitives (the struct helper is told whether `$index` was requested, like php-src's `parser->info`); `BuiltinArgumentLowering::XmlHandlerSetter` carries the closure hints into codegen.
6. **feat(xml): add the xml prelude** — `XMLParser` (final, php-src handler binding rules, `xml_set_object()`, recursion guard, `parse_into_struct` accumulation) and `XMLWriter` (memory / URI / stream modes, `php://output` through `STDOUT`), plus the procedural wrappers, built as Rust AST from a PHP form kept as a node-by-node oracle. Injected by usage detection or forced by `--with-xml`, in the pipeline and in the test harnesses (including the examples-corpus lowering harness).
7. **feat(magician): add the xml eval area** — 64 homes forward to the compiled prelude through the native-function bridge; the ten registry builtins are reached through the `XMLParser` object's methods, with `xml_parse_into_struct()` writing back through the captured caller lvalues. `extension_loaded('xml' | 'xmlwriter')` answers false when the host never linked the bridge, and so does `function_exists()` for the 54 prelude functions (the ten registry builtins exist on both backends regardless).
8. **test(xml)** — codegen fixtures (event stream with positions, error codes, incremental parsing, namespaces, handler routing and exceptions, `xml_set_object()`, `xml_parse_into_struct()`, XMLWriter output/indent/DTD/flush semantics, the same surface inside `eval()`, and one fixture per documented divergence from PHP with PHP's output recorded alongside), error tests, the harness gate (`skip_without_xml_native`, loud failure under `ELEPHC_TEST_REQUIRE_XML_NATIVE=1`) and the managed-archive link provider shared with curl.
9. **docs(xml)** — `docs/php/xml.md` (enabling the managed package, guide, measured "Runtime limits", "Differences from PHP"), the xml example, README / CLI / linking / native-dependencies references, AGENTS.md (`--with-xml` as the second bridge needing a managed package), ROADMAP entry, `.plans/xml.md`, docs tooling.
10. **docs(builtins)** — regenerated builtin pages with the canonical `--features curl` exporter.

## Enabling

`elephc native add libxml2` once per project; then any program using the surface links the bridge and the package automatically, or `--with-xml` forces both. A compile that plans the bridge without the package fails closed with the `native add libxml2` recovery; there is no fallback to a system `libxml2`.

## Documented divergences from PHP

- External DTD subsets and external parsed entities are never fetched: the bridge refuses every external resource load (php-src would go through its stream layer).
- The PHP 8.4/8.5 deprecations on `xml_set_object()`, `xml_parser_free()` and non-callable string handlers are not emitted.
- `xml_parse_into_struct()` throws `Error` (like `xml_parse()`) when called from a handler; `$values` / `$index` are typed `mixed`.
- `xmlwriter_open_uri()` throws `ValueError` instead of returning `false` (its return type is `XMLWriter`); `XMLWriter::openUri()` returns `false` with the stream layer's warning.
- Handlers installed from inside `eval()` must be compiled callables (an eval closure cannot be invoked by the compiled parser).
- `XMLParser::__clone()` / `XMLWriter::__clone()` are `final`: a subclass overriding them is a compile error, where PHP compiles the override and still refuses the clone at run time.

## Review pass

A six-lens adversarial review (FFI/memory, php-src fidelity, compiler, eval, catalog, link/CI/docs; every finding re-checked by a refuter and a reproducer) found 41 confirmed issues; all of them are fixed here and pinned by tests:

- `xml_parse_into_struct()` outputs in static, global or by-reference-captured variables (the compiler used to panic / drop `$index`); outputs that are properties or array elements are rejected with a diagnostic.
- Handler setters: first-class callables with fewer parameters, `array $attributes` on a literal closure, a compile error for variadic or over-required closures and for a function name whose spelling differs from its declaration (dynamic callables resolve case-sensitively); the detection walk now covers enum members and variadic types.
- Binary-safe `XMLWriter` output (UTF-16 / UCS-4 documents crossed the C ABI with their NUL bytes stripped), complete `catch_unwind` coverage of the ABI, error codes wrapped to a C `int`, `-bundle` link modifiers so a test build never bakes libxml2 into the shipped bridge archive.
- Prelude: an external-entity-ref handler that throws stops the parser with error 21 like php-src, scalar handler names are coerced like PHP's coercive parse, `xml_parser_free()` from inside a handler answers `false`.
- Magician: by-reference outputs in every dynamic-callable shape (`$f(...)`, first-class callables, `call_user_func_array` with references), PHP's type names in `TypeError`s, a clear catchable `Error` for handlers or objects declared inside `eval()`, the ten registry builtins callable without the bridge (TypeError instead of undefined function), `get_loaded_extensions()` listing xml/xmlwriter.
- Catalog: libxml2 and the shim compiled at `-O2` with `-ffile-prefix-map` (the artifact was `-O0` and carried the builder's staging path), in-process xz test fixtures, resolver label test; nightly/release verify jobs install the zlib/bzip2 dev packages the probe link needs; docs corrected (catalog lists, `xmlwriter_open_uri()` description, the differences list).

## Review follow-up

The consolidated review (Codex + Kimi K3 + GLM 5.3) raised three P2 findings; all three are fixed here and pinned by tests:

- `XMLParser` and `XMLWriter` (and `XMLWriter` subclasses) are uncloneable like PHP's: `clone` throws `Error: Trying to clone an uncloneable object of class ...` in compiled code and inside `eval()`, and the rejected copy hands its handle back before the exception, so the original keeps working.
- The error code visible from inside a SAX handler is snapshotted per event, like positions: handlers for the well-formed prefix of a chunk see `0` even when the same chunk fails later (`ROOT:0 OK:0`, final `76`), identically in one chunk or split across calls; an explicit stop still overrides it.
- `xml_set_*_handler()` calls with named arguments, including ones that unpack the parser (`...[$p]`, `...['parser' => $p]`, `...$args` from a list, `...$named` from a string-keyed array), get the same handler-closure lowering as positional calls (static spreads flattened first, dynamic list unpacks through the shared spread plan with the hints applied per slot, dynamic string-keyed unpacks through the plan's normalized by-key reads; source evaluation order kept), so a bare `array $attributes` on such a call receives the attribute map; the checker's eager argument pass now maps a named callback argument to its parameter slot before deciding to skip it, so a handler written before `parser:` with a type-sensitive body compiles (this also fixes named `callback:` arguments of the array callback builtins at check time).

- Serialization follows php-src: `XMLWriter` serializes as an empty object and unserializes to a fresh, unopened writer (its handle and stream state never travel, so a restored copy cannot own the original's handle); `XMLParser` refuses `unserialize()` too (`Unserialization of 'XMLParser' is not allowed`), not only `serialize()`.
- A handler that arrives inside a run-time unpack (`xml_set_element_handler($p, ...$handlers)`) is the builtins' general unknown-length-unpack diagnostic at compile time; documented under "Differences from PHP" and pinned by an error test.

## Pre-existing issues found (not fixed here)

Filed as #924 (invocation leak), #925 (invoker array marshalling), #923 (array read-copy leaks), #946 (a failed `clone` leaks the shallow copy) and #947 (`clone` through a base-typed variable uses the static class); follow-ups on this PR's own code as #926 (xz block memory bound) and #927 (hide `__elephc_*` from user code). Also seen and not addressed here, both independent of xml: `array_map(callback: ..., array: ...)` with named arguments fails EIR validation, and unpacking a function call's string-keyed array before named arguments (`f(...named(), key: ...)`) fails EIR validation because the spread is classified by its static `array` type (same family as #685).

**Closure and dynamic-callable invocations leak one small heap block per string argument, and per argument bound to a `mixed` or untyped parameter** (32–40 bytes each) in the compiled runtime, on `main` too (`--heap-debug` on a loop of `$f("x$i", "y")`: 2000 live blocks after 1000 calls; ints into `int` parameters, objects into class-typed parameters and direct named-function calls are clean; an array handed to a `mixed` parameter leaks its whole box). The arguments go through the descriptor-invoker container (`src/ir_lower/expr/descriptor_calls.rs`, `emit_invoker_arg_mixed` / `release_invoker_arg_preserving_result` in `src/codegen/lower_inst/callables.rs`). For SAX parsing this means a start+end handler pair exhausts the default 8 MB heap between 85,000 and 90,000 elements (`Fatal error: heap memory exhausted`), and a `mixed $attributes` handler between 6,000 and 7,000; `--heap-size` raises the ceiling (64 MB: 300,000 elements). Measured and documented under "Runtime limits" in `docs/php/xml.md`; tracked in #924.

**Array read-copy leaks in plain PHP** (no callables involved): `$p = isset($this->prop[$k]) ? $this->prop[$k] : []; $p[] = $x; $this->prop[$k] = $p;` leaks the previous list on every store (quadratic: 6.5 MB after 1000 iterations in a probe) while the same code with `$this->prop[$k] ?? []` is clean, and a nested append `$this->prop[$k][] = $x` leaks the whole previous array each time. The xml prelude's `xml_parse_into_struct()` index builder hit the first shape and now uses `??`, so the struct memory is the document's own footprint (about 1.6 KB per element; 4,000 elements fit the default heap). Tracked in #923.

Dynamic callable argument marshalling in the runtime invoker: a handler declared with a bare `array $attributes` and invoked through a runtime descriptor receives garbage (`[7]`), and a local grown element by element from `[]` is boxed with the indexed-array tag (a `mixed` callee then sees `count() == 0`). The prelude works around the second one (handler attributes are handed over through a fresh copy) and the docs tell users to declare dynamically resolved handler parameters as `mixed`. Tracked in #925.

## Verification

Every commit compiles the whole workspace on its own (`cargo check --workspace` per commit, warning-free). Locally green at HEAD on linux-x86_64 against the catalog-built artifact, rebased on current `main`: `elephc-xml` (106 native tests, plus the two `--ignored` fixture regenerators, which pass against `php` 8.5.10), `elephc-builtin-contract` (24), `elephc --lib` (1671, including the examples corpus and the xz/catalog/recipe tests), `elephc --bin` (1921, including the linker/link-planning/workflow-pinning tests), `builtin_parity_tests`, `error_tests` (1518), `verify_release_artifact`, Magician lib (1192), `codegen_tests xml::` + the `--with-xml` CLI diagnostic + the eval builtin parity probe (88, real links through the managed archives), the EIR boundary audit, the docs pipeline (extract/audit/validate/unittests, 444 non-registry routes, generated pages byte-identical), `git diff --check` over the branch, warning-free `cargo build`. `examples/xml/main.php` compiled through the production resolver matches PHP 8.5.10 byte for byte. Not verified locally and left to CI: macOS/iOS (`-liconv`, the Apple libxml2 build, the xml iOS compile-and-link job), linux-aarch64, and the full `codegen_tests` suite.

Closes #930, closes #931.


nahime0:
Follow-ups from the Grok merge-gate review (deferred so this PR can merge):

- #928 — descriptor-invoker leak for 2+ arg closure/dynamic SAX handlers
- #929 — dynamic invoker assoc-array boxing for handler attributes
- #930 — CI xml iOS compile/link (catalog libxml2)
- #931 — sync `AGENTS.md` for `--with-xml` / second catalog-native bridge
- #932 — docs: heap/array caveats + soften complete/100% claims
- #933 — codegen pins for documented divergences
- #934 — optional hardenings (`catch_unwind` SAX, hide `__elephc_*`, xz block-size)

Guikingone:
Review of **11826c34ff7914a6438cf2a455d939a654afd943**, against base `a5c29be9605d6bb144ab97ab8357a74e4ff4cb4e`.

Consolidated review by Codex, with Kimi K3 (`kimi-k3:cloud`) and GLM 5.3 (`glm-5.3:cloud`) through Ollama in dedicated Codex sessions. Independent initial inspections were followed by targeted verification and consolidation. Findings below were rechecked against the source and reproduced by the consolidating reviewer using this PR's macOS CI compiler artifact and managed libxml2 2.15.3.

**Three confirmed issues to fix before merging:**

### 1. [P2] Reject cloning XMLParser and XMLWriter before copying their native handles

Locations: [`src/xml_prelude/build/writer.rs:20–41`](https://github.com/illegalstudio/elephc/blob/11826c34ff7914a6438cf2a455d939a654afd943/src/xml_prelude/build/writer.rs#L20-L41), and [`src/xml_prelude/build/parser.rs:80–88`](https://github.com/illegalstudio/elephc/blob/11826c34ff7914a6438cf2a455d939a654afd943/src/xml_prelude/build/parser.rs#L80-L88).

These new classes own an integer bridge handle and free it in their destructor, but neither prevents cloning. The generic `ObjectCloneShallow` path copies the integer unchanged, giving two PHP objects ownership of the same parser/writer. Mutating the copy changes the original; destroying or reopening the copy removes the registry entry and leaves the original unusable. The existing AOT and eval clone exclusion paths do not exclude these classes.

```php
$w = new XMLWriter();
$w->openMemory();
try {
    $copy = clone $w;
    $copy->writeElement("from_copy");
    echo "original:", $w->outputMemory(), "\n";
    unset($copy);
    var_dump($w->writeElement("after_copy_destroyed"));
} catch (Throwable $e) {
    echo get_class($e), ": ", $e->getMessage(), "\n";
}
```

Observed Elephc output:

```text
original:<from_copy/>
bool(false)
```

PHP rejects the clone with `Error: Trying to clone an uncloneable object of class XMLWriter`. The same problem is reproduced for `XMLParser`: clone it, destroy the copy, then `xml_parse($original, "<ok/>", true)` returns `0`. Cloning a compiled XMLWriter inside `eval()` also succeeds and shares its output. PHP 8.5.10 explicitly disables the clone handler for [XMLParser](https://github.com/php/php-src/blob/php-8.5.10/ext/xml/xml.c#L207-L218) and [XMLWriter](https://github.com/php/php-src/blob/php-8.5.10/ext/xmlwriter/php_xmlwriter.c).

Please add an uncloneable policy for both classes, including XMLWriter subclasses and eval, with regression tests. Reject **before** a shallow copy becomes an owner: merely adding a throwing hook after copying the handle needs care so exception cleanup cannot free the original's handle. The bridge registry makes a second free a no-op, so this is an ownership/behavior defect, **not a demonstrated C double-free**.

### 2. [P2] Snapshot the error code exposed during queued SAX callbacks

Location: [`crates/elephc-xml/src/parser/mod.rs:324–331`](https://github.com/illegalstudio/elephc/blob/11826c34ff7914a6438cf2a455d939a654afd943/crates/elephc-xml/src/parser/mod.rs#L324-L331).

`feed()` parses the entire chunk before the PHP prelude drains the queued events. Positions are snapshotted per event, but `error_code()` reads the final live libxml2 state. Consequently, handlers for valid opening tags observe errors that occur later in the same chunk. Code that checks parser errors inside a handler behaves differently depending on how the same XML document is chunked.

```php
$p = xml_parser_create();
xml_set_element_handler($p, function ($p, $name, $attrs) {
    echo $name, ":", xml_get_error_code($p), "\n";
}, null);
var_dump(xml_parse($p, "<root><ok/></wrong>", true));
echo "final:", xml_get_error_code($p), "\n";
```

| Observation | Elephc | PHP |
|---|---|---|
| Root start callback | `ROOT:76` | `ROOT:0` |
| Child start callback | `OK:76` | `OK:0` |
| Parse return / final error | `0` / `76` | `0` / `76` |

Feeding `<root><ok/>` non-finally, followed by `</wrong>` finally, produces `ROOT:0` and `OK:0` on **both** implementations. Please capture the handler-visible error state alongside each queued event, expose it while delivering that event, and preserve the final/explicit-stop error outside delivery. Add both one-chunk and split-chunk regressions.

### 3. [P2] Preserve XML handler parameter hints for named arguments

Location: [`src/ir_lower/expr/array_builtin_args.rs:130–141`](https://github.com/illegalstudio/elephc/blob/11826c34ff7914a6438cf2a455d939a654afd943/src/ir_lower/expr/array_builtin_args.rs#L130-L141).

The new `XmlHandlerSetter` lowering applies only when there are no named arguments or spreads. A named call falls through to generic argument lowering, bypassing `handler_closure_params()` and the XML event hints. The checker still accepts and specializes that closure. For a bare `array` annotation, the bypass restores the incorrect argument representation that the positional special lowering was introduced to avoid.

```php
$p = xml_parser_create();
xml_set_element_handler(
    parser: $p,
    start_handler: function ($parser, $name, array $attributes) {
        var_dump($attributes);
    },
    end_handler: null
);
xml_parse($p, '<root id="x"/>', true);
```

Observed Elephc output:

```text
array(1) {
  [0]=>
  int(0)
}
```

PHP, and the otherwise identical **positional** Elephc call, both return the correct attribute map:

```text
array(1) {
  ["ID"]=>
  string(1) "x"
}
```

This is the PR's missing named-call application of its own workaround, not a request to solve the general invoker issue #925 here. Please apply the handler hints using the shared argument plan for named/spread calls while retaining source evaluation order, and add a regression that observes the whole attribute map. GLM identified the missing path; its initial example that directly indexed `"ID"` did **not** reproduce the defect, but the whole-array probe above does.

### Validation and scope

- Used the `nextest-archive-macos-aarch64` compiler/bridge artifact from [CI run 34242043145](https://github.com/illegalstudio/elephc/actions/runs/34242043145), plus the pinned managed libxml2 2.15.3 package.
- Reproduced writer cloning, parser cloning, cloning through eval, the one-chunk/split-chunk error-state discrepancy, and malformed attributes with a named handler call. Executed PHP comparisons with local PHP **8.5.6**; this is not a claim that the full PHP 8.5.10 differential corpus was rerun. The pinned 8.5.10 source was also checked for the clone prohibition.
- Reviewed XML bridge/FFI, preludes, compiler integration, eval dispatch, native package/linking and CI changes. This was a targeted review of a large PR, not an exhaustive proof of the entire compatibility surface.
- Kimi confirmed findings 1–2 and rejected the handler-clear hypothesis. GLM identified finding 3 and independently confirmed the source paths for findings 1–2; all three final findings have consolidator-run reproductions.
- Existing CI snapshot: 148 successful checks and one skipped. No full local suite or new Linux/iOS execution was performed. No source changes were made.
- Documented divergences and already tracked #923–#927 are excluded from the findings. An alias-output probe matched PHP. The named-handler finding is limited to the demonstrated whole-array observation; the initial direct-key-access probe matched PHP.

nahime0:
Thanks for the consolidated review. All three findings are fixed on the new head, rebased on current `main`, each with regressions pinned against PHP 8.5.10:

**1. Uncloneable `XMLParser` / `XMLWriter`** — both prelude classes now declare a `final public function __clone(): void` that detaches the bridge state from the copy *before* throwing (`__elephc_handle = 0`, and for the writer also the URI-mode/stream fields), then throws `Error: Trying to clone an uncloneable object of class <runtime class>`. Because both the compiler's `lower_clone` and Magician's `eval_object_clone_result` shallow-copy first and then run `__clone()` on the copy, the same hook covers compiled code, `eval()` (host-created and eval-created objects) and `XMLWriter` subclasses; the copy's destructor already skips handle 0, so it can never free or flush the original's handle. Fixtures: `tests/codegen/xml/clone.rs` (writer, parser, subclass naming the runtime class, inside `eval()`; every original is used after the failed clone). The hook is `final` on purpose: a subclass overriding it would silently re-enable the shared-handle clone, so that override is a compile error (`tests/error_tests/xml.rs`, documented under "Differences from PHP").

**2. Error code visible from handlers** — `Queued` events now carry the libxml2 `errNo` snapshot taken in the SAX callback next to the position snapshot; `Parser::error_code()` answers the delivered event's snapshot during delivery, the live value otherwise, and an explicit `stop()` code wins from then on (mirroring `frozen_pos`). Your program prints `ROOT:0 OK:0 / int(0) / final:76` in one chunk and the same across two chunks; `xml_parse()`'s return is untouched. Unit tests in `parser/tests/probes.rs` (one chunk, split, non-fatal namespace error visible from the offending tag on, stop precedence) and codegen fixtures in `tests/codegen/xml/parser.rs` (including a character-data handler reading the code after an earlier non-fatal error, and the ext-entity stop reading 0 inside the handler and 21 afterwards).

**3. Named-argument handler setters** — `XmlHandlerSetter` gained a named-argument branch that builds the shared call plan, lowers the sources in source order with the event hints applied by parameter slot, and assembles the operands from the plan (spreads still take the generic path, like the positional case). Your `parser:/start_handler:/end_handler:` program now prints `["ID"=>"x"]`. While pinning it we found the checker's eager argument pass skipped contextual callback slots by *source* position, so a handler written before `parser:` with a type-sensitive body (`strlen($data)`) was rejected at check time; it now maps a named argument to its parameter slot first (this also covers `array_map(callback: ..., array: ...)` at check time). Fixtures in `tests/codegen/xml/handlers.rs` (attribute map, signature-independent order, handler-first, positional parser + named handlers, named function names / first-class callables) and an error test for the surplus-parameter diagnostic on a named call.

Two pre-existing compiler behaviours surfaced while writing the clone fixtures and are not addressed here: a failed `clone` (any class whose `__clone()` throws) leaks the shallow copy, and `clone` through a base-typed variable builds the copy with the static class (`get_class` reports the base). I can file both.

nahime0:
@Guikingone ready for a second review

Guikingone:
Follow-up review of **800885e1f1fb3a4d81039e802ca2b45202e4cd72**, against base `640380cf8856379228db77e48c142d7613ed4e48`. Compared with the [previous consolidated review](https://github.com/illegalstudio/elephc/pull/913#issuecomment-5590456596) at `11826c34ff7914a6438cf2a455d939a654afd943`; `git range-diff` was used to distinguish the follow-up changes from the rebase.

Codex review, with independent Kimi K3 (`kimi-k3:cloud`) and GLM 5.3 (`glm-5.3:cloud`) passes through Ollama in dedicated Codex sessions. After checking the executable probe evidence, both reviewers agree with the residual P2 below; no additional confirmed findings are retained.

**One P2 remains: the handler-typing fix still bypasses calls containing argument unpacking.**

### Status of the previous findings

| Previous finding | Result on this head |
|---|---|
| Cloning XMLParser/XMLWriter shares the native handle | Fixed for the tested AOT and eval cases: cloning throws and the original survives, including XMLWriter subclasses. |
| SAX handlers see an error from later in the same chunk | Fixed: one-chunk and split-chunk probes now both report `ROOT:0`, `OK:0`, then final error `76`. Explicit external-entity stop remains `21` after a subsequent parse attempt. |
| Named handler closures receive malformed attributes | Direct named calls are fixed. Calls mixing unpacking and named handlers still reproduce the malformed attribute map; the finding remains partially open. |

### [P2] Apply the XML handler hints when another argument is unpacked

Location: [`src/ir_lower/expr/array_builtin_args.rs:139–149`](https://github.com/illegalstudio/elephc/blob/800885e1f1fb3a4d81039e802ca2b45202e4cd72/src/ir_lower/expr/array_builtin_args.rs#L139-L149); the new helper also explicitly falls back for spreads at [`builtin_special_args.rs:229–233`](https://github.com/illegalstudio/elephc/blob/800885e1f1fb3a4d81039e802ca2b45202e4cd72/src/ir_lower/expr/builtin_special_args.rs#L229-L233).

Both specialized `XmlHandlerSetter` branches require that **no** argument is a spread. Consequently, unpacking just the parser causes the ordinary, statically named closure argument to bypass `handler_closure_params()` and the SAX parameter hints. The call is accepted and compiles, but the callback receives the wrong array representation.

```php
<?php
$p = xml_parser_create();
xml_set_element_handler(
    ...[$p],
    start_handler: function ($parser, $name, array $attributes) {
        var_dump($attributes);
    },
    end_handler: null
);
xml_parse($p, '<root id="x"/>', true);
```

Observed Elephc output:

```text
array(1) {
  [0]=>
  int(0)
}
```

PHP output:

```text
array(1) {
  ["ID"]=>
  string(1) "x"
}
```

The same failure reproduces with `$args = [$p];` followed by `...$args`. Replacing the unpack with `parser: $p` produces the correct map on this head. This is a residual case of the previous finding, not a duplicate request to fix the general dynamic-invoker issue #925: the closure is still directly present in the call and its handler parameter is statically known.

Please carry the XML closure hints through the shared named/spread argument plan, preserving single evaluation, source order, and spread validation. Removing only the dispatch guard is insufficient because the helper itself falls back for spread plans. Add regressions for both literal and variable parser unpacking, observing the **whole** attribute map rather than only indexing one key.

### Validation and limits

- Used the compiler, XML bridge, Magician, phar and crypto archives from this head's [macOS CI run 34282257057](https://github.com/illegalstudio/elephc/actions/runs/34282257057), with managed libxml2 **2.15.3**. Auxiliary tz/iconv/tls/bcmath archives were reused from the preceding review after verifying their crate sources are unchanged between the two heads.
- Executed **14 PHP-to-native probes**: 12 matched PHP, including the four added clone regression programs, parser/writer/subclass lifetime controls, callback error-state probes and direct named-handler control. The two spread variants reproduced the P2 above.
- Local differential oracle: PHP **8.5.6**. This is not a rerun of the full pinned PHP 8.5.10 corpus.
- Targeted review of the changed compiler/prelude/parser paths and their adjacent lifetime, error-state and argument-planning behavior; existing documented divergences and #923–#927 were excluded.
- CI snapshot: 146 successful checks, one skipped; `git diff --check` passed. No full local suite, workspace rebuild, or new Linux/iOS execution. Source files and the original dirty checkout were preserved.

nahime0:
Thanks for the follow-up. The residual P2 is fixed on the new head: the `XmlHandlerSetter` lowering now takes every argument shape through one entry point (`lower_xml_handler_setter_call`). Static spreads (`...[$p]`, `...['parser' => $p]`) are flattened before the positional/named paths, and a dynamic unpack next to named handlers (`...$args, start_handler: ...`) goes through the shared spread plan with a per-source hook (`lower_named_args_with_spread_plan_hinted`) that applies the SAX hints by parameter slot; the positional prefix is still evaluated once into a temp, sources lower once in source order, and the spread bounds guard is unchanged. Your literal and variable unpack programs now print `["ID" => "x"]`, and a function-call unpack (`...make()`) is evaluated exactly once.

Regressions: `tests/codegen/xml/handlers.rs::test_xml_unpacked_parser_with_named_handlers` observes the whole attribute map for the literal, assoc-literal and variable unpacks (the last with reversed named handlers and a named character-data handler on the same unpack) plus the single-evaluation case; `tests/error_tests/xml.rs` pins the surplus-parameter diagnostic after an unpack. Positional arguments after an unpack remain a compile error on both sides (PHP's fatal), and a handler that lives *inside* the unpacked array (`xml_set_element_handler($p, ...$tail)`) is not a literal the hints can type, so it stays on the generic path (today an arity diagnostic; noted as a separate divergence of dynamic spreads on builtins, not addressed here).

nahime0:
Follow-up hardening on the same theme, on the new head:

- The string-keyed dynamic unpack (`$a = ['parser' => $p]; xml_set_element_handler(...$a, start_handler: ...)`) now gets the handler hints too: the shared spread plan declines assoc dynamic unpacks, so the setter lowers the plan's normalized form (each parameter read from the array by key, in signature order) with the per-slot typing. Pinned in `tests/codegen/xml/handlers.rs` with the whole attribute map.
- `XMLWriter` serialization matches php-src: `serialize()` writes `O:9:"XMLWriter":0:{}` and `unserialize()` yields a fresh, unopened writer (handle and stream state never travel, so a restored copy cannot alias the original's handle); `XMLParser` now refuses `unserialize()` as well as `serialize()`. Pinned in `tests/codegen/xml/clone.rs`.
- A handler that lives inside a run-time unpack (`xml_set_element_handler($p, ...$handlers)`) stays a compile-time diagnostic (the builtins' general unknown-length-unpack rule); it is now listed under "Differences from PHP" and pinned by an error test.

The two general clone behaviours from the previous round are filed as #946 (a failed clone leaks the shallow copy) and #947 (clone through a base-typed variable uses the static class).

nahime0:
New follow-ups from the latest Grok re-verify (`f1dfc4235`; merge still OK when CI is green):

- #948 — assoc unpack from a call filling 2+ setter slots (re-eval / order)
- #949 — `XMLWriter::__serialize` drops subclass-declared properties vs PHP

Still open but already tracked / documented: handler-inside-unpack divergence; clone bugs #946/#947. #930/#931 look done in-tree on this tip.

PR #954, opened 2026-09-10
feat: ctx-register runtime mode (--rt-ctx) — per-context state foundation for sandbox threads

Guikingone:
## What this is

The foundational refactor of the **sandbox-threads plan**: the runtime's mutable per-process state (heap bump offset, free list, small bins, concat scratch) moves from global `.comm` symbols into one `_rt_ctx` struct addressed through a **reserved context register** — `x28` (AArch64) / `r14` (x86_64). A second execution context (the M1 spawned thread) will only need a different register value, not a second copy of the runtime.

Selected per build via `--rt-ctx`; the legacy symbol-addressed runtime remains the default, and both modes coexist in the same compiler (separate runtime cache entries via feature bit 12).

## Commits

1. **feat: spike ctx-register runtime mode (--rt-ctx)** — layout (`_rt_ctx`: imm12-window scalars first, 64 KiB concat buffer last), `__rt_ctx_init` in the main prologue, heap-family routing through x28, x28 removed from the linear-scan pool, e2e tests.
2. **fix: remediate the Kimi K3 review of the spike** — external review of the spike commit; every blocking finding addressed (details below).
3. **feat: ctx-relative concat helpers** — the dual-mode accessors, landed ahead of the family migration.
4. **feat: migrate the concat family to ctx-relative state (M0)** — the full concat family (contract + ~350 sites across 77 files) in one change, per the partial-routing hazard below.

## Key design points

- **Reserved register**: x28/r14 — callee-saved, saved whole by `__rt_fiber_switch`, excluded from both register-allocator pools (AArch64 pool: 8→7). x18 rejected (Apple-reserved); TLS rejected (measured cost, TLV emitter complexity).
- **Measured cost: in the noise.** Allocation-heavy bench (4M allocations + 200k hash inserts, 5 alternating runs): legacy 1.36–1.65s, ctx 1.29–1.61s — ctx equal-to-slightly-faster (imm-offset loads replace adrp+add pairs on hot alloc/free paths).
- **Partial routing is incorrect routing** — mechanically enforced. The spike proved it empirically (a half-routed free path exhausted an 8 MB heap that legacy served fine, because frees landed on a never-read global free list). Now ctx builds **omit** the legacy heap and concat symbols from the data section entirely, so any unrouted helper fails the LINK with an undefined-symbol error instead of corrupting state at runtime.
- **Foreign entries re-publish.** The fiber trap generalizes: a host calling into compiled code preserves its OWN callee-saved ctx register value (host data, not zero). Every entry that reaches compiled PHP from foreign context re-publishes: main prologue (full init), `__rt_fiber_entry` (zeroed fiber stacks), cdylib/staticlib export wrappers, `elephc_init` (library lifecycle), extern FFI callback trampolines.
- **x86_64 is a first-class target**: the x86_64 arms of the whole heap and concat families route through r14; ~120 hand-written runtime scratch uses of r14 were migrated to `rbx` (unused by the runtime, preserved by contract everywhere); the x86_64 fiber wrapper moved its descriptor scratch to r15.
- **Scratch audit**: a dedicated test scans the ENTIRE ctx-mode runtime text on both architectures and fails on any ctx-register reference outside sanctioned shapes (publish, ctx-relative access, fiber-switch save/restore) — a hand-written helper starting to use x28/r14 as scratch cannot slip in silently.

## Tests

- Unit: ctx layout/register conventions, helper emission on both targets, regalloc pool exclusion, cache-key collision, full-runtime generation gates (ctx and legacy), the scratch audit.
- e2e (`--rt-ctx` binaries compile, link, run): allocation-driven program, **heap recycling** (600k iterations ≈ 60 MB through the 8 MB heap — only passes through free-list reuse), **legacy-vs-ctx golden cross-check** (identical output for the same program), `` preserved across `__rt_ctx_init` (the helper's clobber contract is pinned: no argument register touched), generators + Fibers survive the zeroed-fiber-stack trap.
- Suites green after the concat migration: strings (409), json (447), serialize (75), unserialize, var_dump, preg, date, and the rt_ctx suite (5/5).
- Known pre-existing local issue fixed along the way: a fiber scenario test overflowed the 2 MiB default test-thread stack while compiling its 128-local fixture in-process (CI sets RUST_MIN_STACK=32 MiB; a bare local `cargo test` does not) — it now runs on a dedicated large stack.

## Docs

- `docs/internals/runtime-ctx-register.md` — decision note: register choice, measured results, the two traps (partial routing, foreign entries), current state, and the remaining M0 list.
- `docs/compiling/cli-reference.md` — `--rt-ctx` entry, dual-architecture validated.

## What's next (M0 remainder)

- `_rt_ctx` as a pool (array + free list) instead of a single instance; `__rt_ctx_init`/`__rt_ctx_destroy` exported for the M1 thread-pool bridge.
- Exception/fiber/stack-guard state families still on legacy globals (inventoried next).
- Linux x86_64 execution validation via the Docker test script (emission is locked by the asm-level tests; the e2e run is the remaining CI-side check).

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -94,6 +94,7 @@ How elephc works under the hood — from lexing to code generation and runtime s
 - [The Code Generator](internals/the-codegen.md) — checked AST to EIR, then target assembly
 - [The EIR Design](internals/the-ir.md) — PHP-shaped intermediate representation used by codegen and `--emit-ir`
 - [The Runtime](internals/the-runtime.md) — hand-written assembly routines
+- [Runtime Context Register (spike)](internals/runtime-ctx-register.md) — reserved x28/r14 for per-context runtime state, `--rt-ctx`, measured cost, fiber trap, M0 path
 - [Eval Runtime Architecture](internals/eval-runtime.md) — literal AOT planning, scope synchronization, Magician fallback, and bridge ABI
 - [Memory Model](internals/memory-model.md) — stack frames, heap, reference counting
 - [Architecture](internals/architecture.md) — module map, calling conventions
--- a/docs/compiling/cli-reference.md
+++ b/docs/compiling/cli-reference.md
@@ -430,7 +430,7 @@ Elephc CLI process, not a signed iOS application bundle.
 | `--no-ir-opt` | — | — | `ELEPHC_IR_OPT=off` | Shorthand for `--ir-opt=off`. |
 | `--regalloc=linear\|stack` | `linear`, `stack` | `linear` | `ELEPHC_REGALLOC` | Register allocator: linear-scan, or stack-only fallback. |
 | `--null-repr=sentinel\|tagged` | `sentinel`, `tagged` | `tagged` | `ELEPHC_NULL_REPR` | Representation for null-capable scalar slots. |
-| `--rt-ctx` | — | off | — | Route per-context runtime state (heap bump offset, free list, small bins, concat scratch) through the reserved context register (`x28` on AArch64) instead of global symbols. Spike mode for the sandbox-threads plan; the runtime object is cached separately per mode. Currently AArch64-validated only — x86_64 selects it but its r14 reservation is not yet audited. |
+| `--rt-ctx` | — | off | — | Route per-context runtime state (heap bump offset, free list, small bins, concat scratch) through the reserved context register (`x28` on AArch64, `r14` on x86_64) instead of global symbols. Sandbox-threads spike mode; the runtime object is cached separately per mode. Validated on both architectures: legacy heap symbols are omitted from ctx builds, so any unrouted helper fails at link time. |
 
 See [Optimization and codegen controls](optimization.md).
 
--- a/docs/compiling/cli-reference.md
+++ b/docs/compiling/cli-reference.md
@@ -430,6 +430,7 @@ Elephc CLI process, not a signed iOS application bundle.
 | `--no-ir-opt` | — | — | `ELEPHC_IR_OPT=off` | Shorthand for `--ir-opt=off`. |
 | `--regalloc=linear\|stack` | `linear`, `stack` | `linear` | `ELEPHC_REGALLOC` | Register allocator: linear-scan, or stack-only fallback. |
 | `--null-repr=sentinel\|tagged` | `sentinel`, `tagged` | `tagged` | `ELEPHC_NULL_REPR` | Representation for null-capable scalar slots. |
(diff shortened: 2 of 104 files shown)


Guikingone:
## Round-2 review remediation

Every blocking finding from the second external review is addressed and pushed (`dafbf9589f` + `cf97f98077`):

**NB1 (rbx footgun)** — Every runtime helper touching rbx now preserves it: vsprintf + 3 zval helpers spill to dedicated frame slots (restore in shared epilogues), json_encode_str x86_64 spills beside its r15 cache, the 4 spl_dll helpers use push rbx + rax alignment pad, http.rs moved to dead caller-saved r8. Mechanical audit `x86_64_runtime_helpers_that_scratch_rbx_preserve_it` scans every emitted helper for balance, with a **negative control test** proving the audit flags violations. The self-contradictory doc sentence is rewritten. Also fixed the collision the wave-1 swap introduced in str_to_number (two distinct flags had merged into one register — the integer-significand flag now lives in a dedicated stack slot).

**NB2 (trampoline window)** — The ordering tests exposed a **real ABI hole**: the AArch64 callback trampoline published ctx over x28 without saving the foreign caller's value → C hosts received a clobbered callee-saved register. Fixed (spill x28 before publish, restore after invoker); both-arch tests pin save-before-publish / publish-before-ctx-access / restore-before-return. The exception path is documented in the test (longjmp past the trampoline means the host never resumes, or the process exits).

**NB3 (dangling-x6)** — Mechanical audit `legacy_aarch64_runtime_has_no_dangling_x6_stores` over the full legacy runtime; its **first run caught a real survivor** (php_uname's epilogue still loaded the concat offset through a stale x6 address) — fixed.

**NB4 (churn)** — str_to_number was the wave-1 mechanical r14→rbx swap; now explicitly documented in the round-2 commit message along with the collision fix.

**Test gaps landed:** kitchen-sink ctx gate (all features, both arches, zero legacy heap/concat symbol references and zero .comm declarations), e2e `--rt-ctx --heap-debug` workload through the ctx-routed validator (6/6 rt_ctx e2e green), lib suite 262 green.

**Still open (non-blocking per the review's own verdict):** compute/spill-heavy bench for the 8→7 pool question, orphan-matrix × 2-modes × 2-arches error paths, staticlib parity, Linux x86_64 execution shard — tracked as M0 follow-ups.

Guikingone:
Closing in favour of a fork-based PR: the work continues on `Guikingone:spike/runtime-ctx-register`, which is merged up to current `main` (this branch was 361 commits behind, so no CI job but `Classify` ever ran on it).

The replacement carries this branch's history plus a review pass that found the scratch audit was inert, the defects it was hiding, and finished the x86_64 migration. Link to follow.

Guikingone:
Replaced by #958 (from `Guikingone:spike/runtime-ctx-register`, merged up to current `main`).

PR #955, opened 2026-09-11
ci: push to main and the tap with the Illegal Bot App token

nahime0:
Two jobs push with the default workflow token today, which a branch ruleset on `main` would reject: the nightly repository-stats commit (`traffic.yml`) and the release version bump (`release.yml`). Both now mint a short-lived installation token for the **Illegal Bot** GitHub App, and their workflow token drops back to `contents: read`.

Two personal access tokens retire with the change:

- `TRAFFIC_TOKEN` — the traffic/clones API is covered by the App's `Administration: read`
- `TAP_TOKEN` — the Homebrew tap push uses a token minted **scoped to `homebrew-tap` alone**, so it cannot reach anything else in the organisation

### Notes

- `actions/create-github-app-token` is pinned to a full commit SHA (`fee1f7d`), matching how this repo already pins third-party actions that run with write access — this one receives the App private key.
- Unlike the workflow token, an App token **does** trigger workflows on push. Both commits carry `[skip ci]` so main keeps the behaviour it has today; the tagged commit the bump sits on is already verified green by `verify-ci`.
- `nightly.yml` and the release-publishing job still use `GITHUB_TOKEN`: they create tags and releases, which a *branch* ruleset does not gate. A future tag ruleset would need its own bypass configuration.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -67,9 +67,11 @@ jobs:
 
   prepare:
     name: Bump version on main
-    # Pushes the version-bump commit to main.
+    # Pushes the version-bump commit to main with an Illegal Bot installation
+    # token, not the workflow token: `main` is protected and the App is the
+    # ruleset's only bypass actor. The default token needs no write access.
     permissions:
-      contents: write
+      contents: read
     if: github.event.ref_type == 'tag' && startsWith(github.event.ref, 'v')
     runs-on: ubuntu-latest
     needs:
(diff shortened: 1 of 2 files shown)


nahime0:
The remaining review point — the privileged `actions/checkout` steps still referencing a mutable tag — is tracked in #956 rather than handled here.

Pinning it in these two jobs alone would leave the repository half-pinned: `actions/checkout` appears 39 times across the workflows. The inventory in that issue also puts the exposure in order — `dtolnay/rust-toolchain@stable` and `taiki-e/install-action@nextest` are moving refs on an individual maintainer's repository and run in every CI job, which is a weaker guarantee than a GitHub-owned action's immutable-by-convention version tag. That issue does the whole set, after adding the Dependabot configuration that keeps the pins from going stale.

PR #967, opened 2026-09-11
fix: support class constants in defined()

### LINKED ISSUE #752: defined('C::K') silently returns false for a declared class constant
## Summary

`defined('C::K')` returns `false` for a declared class constant; PHP returns `true`.

```php
<?php
class K { const KEY = 42; }
var_dump(defined('K::KEY'));
```

PHP: `bool(true)`. elephc: `bool(false)` — a silent wrong answer, unlike `constant('C::K')` which is at least a loud compile refusal ("constant() class constants are not supported").

Surfaced during the PR #714 reviews (round 2 tail note; reproduces identically on the pre-#714 merge base). Re-verified on current main. Either support the `Class::CONST` form in `defined()` (the checker knows every declared class constant) or make it a loud refusal like `constant()` — silent `false` is the worst of the three options.

### LINKED ISSUE #1005: defined('static::CONST') lacks late-static binding
## Summary

Inside an inherited method, PHP resolves `defined('static::CONST')` against the runtime called class. The AOT class-constant lookup added for #752 deliberately leaves `static::` unresolved, so these probes currently return `false`.

```php
<?php
class Base {
    const BASE_ONLY = 1;
    public static function probe(): void {
        var_dump(defined('static::BASE_ONLY'));
        var_dump(defined('static::CHILD_ONLY'));
    }
}
class Child extends Base { const CHILD_ONLY = 2; }
Base::probe();  // true, false
Child::probe(); // true, true
```

PHP 8.5 produces `true, false, true, true`. Correct AOT support needs runtime called-class dispatch (including inherited visibility and method/closure class scope), rather than another compile-time metadata boolean.

Follow-up from #752 and the review of PR #967: https://github.com/illegalstudio/elephc/pull/967#discussion_r3991303938

### LINKED ISSUE #1006: Trait-imported constants are unavailable to direct AOT scoped access
## Summary

PR #967 fixes the `defined()` facet: PHP-compatible probes now see direct and transitive trait constants through each using class, preserve visibility, and keep the trait receiver itself false.

The broader direct AOT access path remains unsupported:

```php
<?php
trait T { public const TC = 1; }
class UsesT { use T; }
echo UsesT::TC; // PHP: 1; elephc: Undefined class constant
```

Trait constants are collected in separate declaration metadata and composed into EIR class metadata after type checking, while `Checker::infer_scoped_constant_access()` still consults only class/interface/enum constant stores. Correct completion should unify checker-visible trait composition with direct scoped reads, typed/final/conflict validation, reflection, and eval metadata.

Originally opened from #752 and the review of PR #967: https://github.com/illegalstudio/elephc/pull/967#discussion_r3991303963

nahime0:
## Summary
- resolve literal `defined('Class::CONST')` names through class, interface, trait-imported, and enum metadata instead of the global-constant table alone
- preserve PHP class/member casing, inheritance, implemented interfaces, visibility, lexical `self::`/`parent::`, named arguments, and runtime-called-class `static::` behavior
- preserve the hidden called-class capture through class-scope closures and `Closure::bind`, `bindTo`, and `call` on every supported target
- compose trait constants in the frontend, accepting declarations with equivalent folded PHP values and rejecting incompatible trait/class compositions before canonical class metadata is built
- raise catchable PHP `Error`s for invalid relative scopes and mark `defined()` as potentially throwing so exception optimization preserves handlers
- document the behavior and demonstrate class and late-static probes in the classes example

## Verification
- PHP 8 cross-checks for late-static, named, trait-imported, equivalent/conflicting trait, visibility, namespace/casing, and invalid-scope behavior
- `cargo test --test codegen_tests test_defined_literal` (10 passed)
- `cargo test --test codegen_tests closure_bind` (22 passed)
- focused trait-constant codegen/eval/reflection and conflict-diagnostic tests
- focused closure call, existing late-static constant, diagnostics, and EIR smoke tests
- optimizer-on/off parity for late-static and trait-imported probes
- emitter coverage for macOS AArch64, iOS ARM64, iOS Simulator ARM64, Linux AArch64, and Linux x86_64 closure-bind descriptors
- `cargo build`
- generated builtin docs/parity/EIR-boundary audit workflow
- assembly-comment and diff hygiene checks

## Remaining limit
- eval-mode class-constant names and dynamic/first-class-callable `defined()` names remain outside this AOT lookup

Fixes #752
Fixes #1005
Fixes #1006

<div><a href="https://cursor.com/agents/bc-25eeb15c-e9b5-43ee-8020-5df743b3ce30?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-25eeb15c-e9b5-43ee-8020-5df743b3ce30&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


nahime0:
Follow-up for the residual Greptile 4/5 (named/scoped trait const equality before compare): #1007

PR #994, opened 2026-09-13
test(oop): cover the constructor_owner cases #796 left untested

### LINKED ISSUE #871: Tests missing for constructor_owner: method_exists, hiding, omitted defaults
## Summary

Leftover from #796. `tests/codegen/oop/relative_types.rs` and `tests/error_tests/classes_traits.rs` cover the #795 0-arg `new static()` success path. Untested, and able to be wrong:

- `method_exists('Child', '__construct')` / `get_class_methods('Child')` stay false (private ancestor ctor is not a callable member)
- public/protected inherited constructors unchanged
- Child replaces the constructor
- Child's own private ctor hides Owner (LSB `new static()` from Owner must be a compile error)
- omitted defaults on the owner path
- `test_singleton_subclass_runs_private_constructor_and_dispatches` uses `new self()` on the declaring class — that path already worked before #796

Guikingone:
Closes #871.

#796 introduced `types::constructor_owner` — the walk that finds the class whose `__construct` PHP runs when a descendant of a class with a non-public constructor is instantiated — and the existing coverage exercised only the 0-argument private-ancestor success path.

**Every gap the issue lists is a behaviour that is already correct on `main`.** What was missing is the pins that keep it so, which is what this PR adds; no production code changes.

## What is covered

A new `tests/codegen/oop/constructor_owner.rs` groups the member-visibility half, where private and protected deliberately diverge:

| Test | Pins |
|---|---|
| `test_protected_ancestor_constructor_is_inherited_and_runs` | a **protected** ancestor constructor IS inherited: it runs through the descendant, and `method_exists()` is `true` on both classes |
| `test_protected_ancestor_constructor_leaves_the_descendant_uninstantiable` | it is still not public, so `isInstantiable()` is `false` and the reflected declaring class is the ancestor |
| `test_private_ancestor_constructor_is_not_a_member_of_the_descendant` | a **private** one is NOT inherited: `method_exists()` is `false` on the descendant, `true` on the declaring class — the half that makes the owner walk necessary at all |
| `test_get_class_methods_omits_an_inherited_private_constructor` | `get_class_methods()` agrees, with an inherited public method alongside |
| `test_a_descendants_own_public_constructor_replaces_a_private_ancestors` | a descendant's own public constructor replaces a private ancestor's |

Two error tests in `classes_traits.rs` cover the direction the walk must **not** take — a descendant that declares its own **private** constructor hides an inherited public one, so `new static()` from the ancestor's scope and a named `new Child()` both name the *descendant's* constructor, as php-src does (`Call to private HideChild::__construct() from scope HideOwner`). Resolving to the ancestor would have *accepted* those calls, since the ancestor's constructor is public and the call site is its own scope.

The `get_class_methods` assertion is deliberately a contrast rather than a count of zero: the function is **scope-aware**, so from global scope a private constructor is omitted from its own declaring class too (php-src returns `0` for both classes when that is the only method). The inherited public method alongside it is what proves the list is populated at all.

## Verification

Every fixture's output was diffed against reference PHP 8.5 and matches byte for byte. `--test codegen_tests oop` (810) and `--test error_tests` (1520) pass.

## Overlap

The remaining items on the issue — omitted defaults on the owner path, a descendant replacing the constructor, and `method_exists`/`getMethods` staying false — are covered by the tests in #990 and #991, so they are not duplicated here.


nahime0 (approved):
only tests, merge and forget

PR #995, opened 2026-09-13
fix(eval): curl string arguments throw a TypeError instead of stringifying to ""

### LINKED ISSUE #872: eval: curl_escape/unescape/init stringify objects to empty instead of TypeError
## Context

Found while merge-gating #730 (`feat/curl` @ `65c3aa30567`).

AOT types `curl_escape` / `curl_unescape` / `curl_init` at compile time. Magician eval still goes through unguarded `cast_string` (`eval_curl_escape_result`, `eval_curl_unescape_result`, `eval_curl_init_result`). An object or array becomes `""` and the call succeeds. PHP throws a catchable `TypeError`.

`curl_setopt` string/slist/POSTFIELDS already TypeError on both backends. This is the same house-wide Magician string-cast gap as `urlencode` / `date`, visible here because curl has three unguarded entry points.

Not a merge blocker for #730: a correct PHP-shaped program does not pass objects there. It is still a silent-wrong-value hole for eval.

## Reproducer (eval)

```php
<?php
var_dump(curl_escape(curl_init(), new stdClass()));
```

Expected (PHP): `TypeError`.
Elephc eval: empty string, process continues.

Guikingone:
Closes #872.

## What was wrong

```php
<?php
var_dump(curl_escape(curl_init(), new stdClass()));
```

| | Result |
|---|---|
| php 8.5 | `TypeError: curl_escape(): Argument #2 ($string) must be of type string, stdClass given` |
| elephc eval | `string(0) ""` — the call **succeeded** |

`curl_escape()`, `curl_unescape()` and `curl_init()` took their string argument through an unguarded `RuntimeValueOps::cast_string()`. AOT rejects the same programs at compile time from the declared signature, so this is the eval half catching up to a check the compiled side already performs statically — a silent-wrong-value hole rather than a missing diagnostic.

## Fix

A shared `eval_curl_string_argument()` in the family's `handle.rs` does what php-src's typed parameter does: it rejects arrays and objects without `__toString()`, naming the type through `eval_curl_given_type_name` — the same helper every other guard in this family uses, and one whose answers (`get_class()` for an object, `gettype()`'s `"array"`) happen to match php-src exactly here, so these messages need no AOT-divergence caveat like the `$handle` ones above them.

**Coercive typing is preserved** for everything that really does convert. `int`, `float`, `bool` and `null` still reach `cast_string`, and an object *with* `__toString()` is converted through it — `curl_escape($ch, new Stringable())` is accepted by php-src and returns the escaped result, so rejecting every object would have traded one divergence for another.

Telling those two object cases apart needed a predicate the interpreter did not have. `eval_object_has_to_string()` in `method_binding` mirrors both dispatch branches (an eval-declared class's public, non-static, non-abstract, zero-parameter `__toString`; a runtime object's synthetic implementations plus the AOT dispatch metadata). It answers the question a **typed parameter** asks, which is not the one `eval_string_context_value` answers — a string context that cannot convert throws `Object of class X could not be converted to string`, while a typed parameter throws the `TypeError` above; php-src draws the same line. It is `curl`-feature-gated like `eval_throw_runtime_exception` beside it, until the sibling typed-string builtins the issue names (`urlencode`, `date`) get the same guard.

## Tests

Two fixtures in `tests/codegen/curl/eval.rs`:

- the four rejected shapes with their exact php-src messages, plus an `alive` marker proving the exception is caught by ordinary PHP `try`/`catch` inside `eval()` and the script keeps running
- a companion pinning that every value PHP coerces is still coerced: `5`, `1.5`, `true`, `"a b"`, a `__toString()` object, and `null`

Every expected string was measured against the real interpreter first.

`--test codegen_tests curl` (200) and `cargo test -p elephc-magician --features curl` (1205) pass. Both `cargo build` and `cargo check -p elephc-magician --features curl` are warning-free.


Guikingone:
> The new `curl_init()` failure path currently leaks the easy handle allocated before validation.

Good catch — real, and fixed in `01806fb5cd`.

The check sat between `ffi::easy_init()` and `open_curl_easy_handle()`, so a rejected `curl_init(new stdClass())` threw with the raw libcurl handle already allocated and never registered in the resource table: unreachable by `curl_close()` and by the table's own teardown, one leaked native handle per rejected call.

Nothing in the check needs the handle, so it now runs **before** the allocation and the window is gone structurally rather than being cleaned up after the fact.

Coverage note, stated plainly: a leaked *native* handle is not observable from PHP, so the added fixture cannot assert it the way a heap-debug test asserts a heap leak — the ordering in `eval_curl_init_result` is the real guarantee, and its docblock says so. What the fixture pins is that 200 rejections in a row leave the interpreter healthy and a later `curl_init()`/`curl_escape()`/`curl_close()` cycle still works, which is what would break first if the allocation came back ahead of the check.

nahime0 (approved):
lgtm

Issue #355, opened 2026-06-10
Generator::throw skips finally blocks during unwinding


nahime0:
## Summary
Throwing into a suspended generator does not execute the generator's `finally` block before control unwinds to the caller.

## Minimal repro
```php
<?php
function g() {
    try {
        yield 1;
    } finally {
        echo 'F';
    }
}

$g = g();
$g->rewind();
try {
    $g->throw(new Exception('x'));
} catch (Exception $e) {
    echo ':' . $e->getMessage();
}
```

## Steps to reproduce
1. Save as `main.php`.
2. Run `php main.php`.
3. Run `target/debug/elephc main.php` and then the generated binary.

## Expected behavior
PHP 8.4.19 prints:
```text
F:x
```

The `finally` block runs before the injected exception is observed by the caller.

## Actual behavior
Elephc prints:
```text
:x
```

The exception reaches the caller, but the generator's `finally` block is skipped.

## Environment
- elephc commit: `208c0c0f392d6bd1b49b98fd510e059ca91d9076`
- elephc version: `0.23.8`
- PHP: `8.4.19`
- OS: `Darwin 25.5.0 arm64`
- Reproduced at: `2026-06-10T07:06:53Z`

## Tests to add
- `Generator::throw()` into a generator suspended inside `try/finally`
- variant where `finally` mutates state before the caller catch runs

Issue #351, opened 2026-06-09
Missing PHP warnings/deprecations for string offsets and float array keys


nahime0:
## Summary
Elephc matches PHP's return values for several coercive string/array access cases, but misses PHP diagnostics on those same paths.

### Out-of-bounds positive string offset
```php
<?php
var_dump('abc'[99]);
```

### Out-of-bounds negative string offset
```php
<?php
var_dump('abc'[-4]);
```

### Float array key with precision loss
```php
<?php
echo [1 => 'x'][1.9];
```

## Steps to reproduce
1. Save one repro as `main.php`.
2. Run `php -d display_errors=1 main.php`.
3. Run `cargo run -- main.php && ./main`.

## Expected behavior
PHP 8.4 emits diagnostics in addition to the result:
- `var_dump('abc'[99])` warns: `Uninitialized string offset 99`
- `var_dump('abc'[-4])` warns: `Uninitialized string offset -4`
- `[1 => 'x'][1.9]` emits `Deprecated: Implicit conversion from float 1.9 to int loses precision`

## Actual behavior
Elephc produces the value/result only:
- `string(0) ""` for the string-offset cases, with no warning
- `x` for the float-key case, with no deprecation

## Notes
This looks like a PHP-diagnostic compatibility gap rather than a value-semantics bug. I am filing it separately from #293 because that issue covered out-of-bounds indexed-array access warnings, not string-offset warnings or float-key coercion deprecations.

## Environment
- elephc version: `0.23.7`
- Host OS: macOS / Darwin arm64
- PHP reference: `PHP 8.4.19 (cli)`

nahime0:
Fresh repro from the 2026-06-17 differential run on current `main` (`9a65c244f8b5b553f483fb326f9e8eace26611cb`), macOS arm64, PHP `8.4.20`, elephc `0.24.1`.

This still reproduces with multiple float keys in the same literal:

```php
<?php
$a = ['01' => 'a', '1' => 'b', 1.9 => 'c', -0.9 => 'd'];
foreach ($a as $k => $v) {
    echo gettype($k) . ':' . $k . '=' . $v . '|';
}
```

Observed today:

- PHP 8.4.20 prints two deprecations, then:
  ```text
  string:01=a|integer:1=c|integer:0=d|
  ```
- elephc prints just the value output with no diagnostics:
  ```text
  string:01=a|integer:1=c|integer:0=d|
  ```

[...]

PR #1003, opened 2026-09-14
test(generators): cover send() on a fresh generator, the #338 shape

### LINKED ISSUE #338: `Generator::send()` before first resume returns wrong value
## Description

Calling `Generator::send($value)` on a fresh generator does not match PHP. PHP starts the generator, injects the sent value into the first suspended `yield`, and continues execution. The elephc binary instead returns the initial yielded value.

## Minimal reproduction

```php
<?php
function gen() {
    $x = yield 10;
    echo "|$x|";
}
$g = gen();
echo $g->send(99);
```

## Steps to reproduce

1. Save as `main.php`.
2. Run `php main.php`.
3. Compile and run with elephc.

## Expected behavior

PHP 8.4 prints:

```text
|99|
```

## Actual behavior

The elephc-compiled binary prints:

```text
10
```

## Environment

- elephc 0.23.5
- PHP 8.4.19
- macOS arm64
- reproduced on 2026-06-05

## Notes

This suggests the initial `send()` path is behaving like `current()`/first `yield` retrieval instead of the PHP resume semantics.

Guikingone:
Closes #338.

> **Overlap with #440.** That PR's description also claims #338 ("Already fixed on main; added a regression test…"), so please close whichever of the two you prefer rather than merging both. Two things make me think this one is the live version: #440's diff contains **no** `send()` test at all — `gh pr diff 440 | grep '\$g->send('` returns nothing, so the test its body describes is not on that branch — and #440 is `CONFLICTING`, editing `src/codegen/runtime/data/fixed.rs` and `src/codegen_ir/lower_inst/hashes.rs`, paths that no longer exist (they are `src/codegen_support/runtime/` and `src/codegen/lower_inst/` on main). #440 still owns #351 and #355, which this PR does not touch.

## The bug is already gone

The reported behaviour does not reproduce on current `main`. Both of the issue's reproductions, run side by side:

```
$ php main.php        # PHP 8.5.10
|xB
$ ./main              # elephc, current main
|xB
```

Byte-identical, and so are the surrounding states — a fresh `send()` that runs the body to completion, a chain of fresh sends, `valid()`, `getReturn()`. elephc 0.24.1 printed `A`; the fix landed somewhere between then and 0.27.0 without this issue being closed.

## What was actually missing

The coverage that would have caught it — and the reason it was missing is visible in the file. **Every** existing `send()` fixture in `tests/codegen/generators/send_throw.rs` calls `rewind()` or `current()` before sending:

```php
$g = echoer();
$g->rewind();          // <- generator already running
echo $g->current();
$g->send(100);
```

So the implicit-start path is never taken by any of them. The one fixture whose name promises otherwise —
`test_generator_send_string_payload_to_fresh_yield_assignment_returns_next_yield` — means a fresh yield **assignment**, not a fresh generator: it calls `current()` on the line above the `send()`.

That is exactly the shape the issue reported, and it had no test.

## The tests

**`test_generator_send_on_a_fresh_generator_starts_it_and_returns_the_next_yield`** — the issue's own reproduction plus the richer one from its follow-up comment. Both assert the value *after* the first yield rather than the one the generator was suspended on, which is the whole distinction the bug got wrong.

**`test_generator_fresh_send_completes_and_chains`** — the states around it:

| case | expected |
|---|---|
| fresh `send()`, body runs to completion | returns `null`, `valid()` is `false` |
| fresh `send()` chained twice into `getReturn()` | each payload lands in the yield it was suspended on |

The first row is what separates "started and resumed" from "started and handed back the first yield": only a generator that really ran past its single `yield` can be done. A regression that returned the initial yield would leave it suspended and `valid()` would be `true`.

## Verification

Every expectation was measured against the host PHP 8.5.10 before it was written down; the PHP 8.4.20 output recorded on the issue agrees with all of it.

`cargo test --test codegen_tests generators::send_throw` — 12 passed (10 existing, 2 new).

No production code changes.


nahime0:
Merging this as #440 is old and must be reviewed against the new main baseline

nahime0 (approved):

PR #1014, opened 2026-09-14
security(deps): bump rustls to `0.23.45` for `RUSTSEC-2026-0285`

Guikingone:
Fixes the **RustSec advisory audit** job, which has been red on every open PR since 2026-09-14 — the day the advisory was published.

## What is failing

```
error: 1 vulnerability found!

Crate:    rustls
Version:  0.23.40
Title:    TLS 1.3 handshake messages incorrectly accepted across encryption level boundaries
Severity: 5.3 (medium)
Solution: Upgrade to >=0.23.45
```

[RUSTSEC-2026-0285](https://rustsec.org/advisories/RUSTSEC-2026-0285) / [GHSA-2mjx-qc3c-rqvc](https://github.com/rustls/rustls/security/advisories/GHSA-2mjx-qc3c-rqvc): rustls accepted TLS 1.3 handshake messages sent at the wrong encryption level when they followed a key-changing message in the same record — a plaintext `EncryptedExtensions` packed into the same record as the `ServerHello`, for example. RFC 8446 §5.1 requires the connection to be terminated with `unexpected_message`.

The handshake transcript stays authenticated, so a network-position attacker cannot alter or complete a handshake; the practical effect is that a peer can send in plaintext what should have been encrypted without rustls rejecting the connection. Functionally the same bug as Go's CVE-2025-61730.

## The fix

`Cargo.toml` already asks for `rustls = "0.23"`, so the fixed release was **already inside the declared range** — only `Cargo.lock` was pinning the old one. `cargo update -p rustls` is the whole change: no manifest edit, and deliberately **not** an `.cargo/audit.toml` entry, since that file is for advisories with no available fix and this one has one.

Four lines of substance in the lock:

| | from | to |
|---|---|---|
| `rustls` | 0.23.40 | 0.23.45 |
| `rustls-webpki` | 0.103.13 | 0.103.15 |

The two `windows-sys` edges that move `0.52.0` → `0.59.0` re-point onto a version the lockfile **already carried**, so no crate enters the graph.

rustls reaches the tree through `ureq`, `mysql`, `tokio-postgres-rustls`, `tokio-rustls` and elephc's own TLS bridge, all of which take `0.23`.

## After

`cargo audit` exits 0, with the same 8 allowed warnings as before — the unmaintained/unsound/yanked notices the gate deliberately does not fail on, because they have no fixed version to move to.

Verified with the same `cargo-audit 0.22.2` CI pins. `cargo build` for the whole workspace, plus `cargo test` for `-p elephc-tls` (14), `-p elephc-pdo` (131, including the TLS connector fixtures), `--lib` (1680) and `-p elephc-magician --lib` (1194).


Guikingone:
@nahime0 I'm merging this PR as it should unlock the CI for the others 🙂

nahime0:
👍 fine for me

Issue #1018, opened 2026-09-15
A property read passed straight to a may-alias callee leaks one block per call (the string half of ReleaseUnlessAliases)


Guikingone:
## Summary

A **property read passed directly as an argument** to a callee whose return-alias summary is
a MAY fact leaks one heap block per call. The block is the argument temporary: its release is
suppressed because the result *might* be the same payload, and nothing releases it when the
result turns out to be a different one.

This is the string half of the trade-off #619 solved for `Mixed`: `Op::ReleaseUnlessAliases`
already decides this per call at runtime, but `ir_lower` only emits it when **both** the
argument and the result are `Mixed`/`Union`. A plain `string` argument with a `string` result
falls back to the unconditional skip, which is the leak.

Found while revalidating #512 under the #551 umbrella. #512's reported symptom (corrupted
fields / worker crash) does **not** reproduce on current `main`; this does, and it is the
"borrowed string storage" the umbrella suspected.

## Environment

- `main` at `6e6f02611f`, macOS ARM64, `--gc-stats` (and `--heap-debug` for the block counts)
- Reproduces with `--ir-opt=on` (15/12 and 12/9 below) and with `--ir-opt=off`
  (19/16 and 16/13 — the same three blocks, against a different baseline)

## Minimal reproduction

```php
<?php
class Paste {
    public function __construct(public string $title, public string $body) {}
}

function fresh(string $s): string { return $s . "!"; }

for ($i = 0; $i < 3; $i++) {
    $a = new Paste("t", "a b c");
    echo strlen(fresh($a->body));
}
```

```
666GC: allocs=15 frees=12
```

Three iterations, three leaked blocks. `--heap-debug` agrees: `live_blocks=3 live_bytes=144`.

## What does and does not leak

Every row is the same loop, `$a = new Paste("t", "a b c")` per iteration, three iterations.

| Call | allocs/frees | |
|---|---|---|
| `strlen($a->body)` | 12/12 | clean |
| `strlen($a->body . "x")` | 12/12 | clean |
| `$x = $a->body; strlen(fresh($x))` | 18/18 | clean — a named local is not an owning temporary |
| `strlen(fixed($a->body))`, `fixed` returns `"X"` | 12/12 | clean — summary is `None`, so the release runs |
| `strlen(echoback($a->body))`, `echoback` returns `$s` | 12/12 | clean — proven alias, ownership flows through the result |
[...]

Issue #551, opened 2026-07-16
Ownership/leak follow-up: revalidate and resolve remaining lifetime bugs


nahime0:
## Summary

The leak campaign tracked through #516, #538, #540, and #500 is complete: the corresponding fixes are merged, the Ivory persistent-worker benchmark is heap-flat, and those issues are closed.

However, four older ownership/lifetime reports remain open. Some predate the recent EIR ownership work and may now be partially or fully fixed; one was independently reconfirmed after the #540 stack merged. This umbrella tracks revalidation, classification, and closure of the remaining family without assuming that all four share one implementation fix.

Current `main` at the time of this audit: `ebc39036db`.

## Open ownership/lifetime work

| Issue | Surface | Current status |
|---|---|---|
| [ ] #483 | Owned temporaries consumed inside larger expressions: call→property chains, nullsafe receivers, ternary arms, and Mixed property/index chains | Reported on an older `main`. Revalidate after the property-receiver and read-result cleanup merged through #543/#546. |
| [ ] #486 | Owned object/array temporaries passed to `mixed` parameters | Reported before the centralized Mixed boxing and call-argument cleanup work. Revalidate after #489/#541/#546. |
| [ ] #498 | Mixed boxes overwritten in untyped properties or PHP array elements without releasing the previous value | Still open; the minimal property/array forms and the larger aliased-array workload need fresh measurements on current `main`. |
| [ ] #512 | Returned object fields become corrupted, or the worker crashes, after an intervening heavy call | Reconfirmed 5/5 under `--web` on `84095e0ee`, after the #540 ownership fixes. Simplified CLI forms no longer reproduced, so the remaining trigger may involve borrowed string storage or the web allocation pattern rather than a pure missing-release bug. |

These issues should remain separate implementation units. This umbrella owns status and acceptance, not a single combined patch.

## Revalidation plan

For each issue:

1. Run the original minimal reproduction on current `main`, recording the exact commit and target.
2. Use `--heap-debug` or focused GC counters for leak reports; use the original worker-crash/corruption sequence for #512.
3. Cover `--ir-opt=on` and `--ir-opt=off` where EIR optimization can affect the ownership path.
4. If the repro is now clean because of a merged fix, add or identify durable regression coverage, post the evidence on the original issue, and close it.
5. If it still reproduces, minimize the current path before implementation and fix it in a dedicated PR.
6. Preserve borrowed/owned/alias/COW contracts and validate through the shared EIR/runtime path for every supported target.

[...]

Issue #515, opened 2026-07-09
--web has a fixed per-request allocation ceiling (~27KB body for a highlighted page) independent of --heap-size


mirchaemanuel:
## Summary

Under `elephc --web`, a request handler that allocates/emits more than
roughly 100 KB total, or a real page whose per-request working set (multiple
copies of a body through find → rebox → highlight → render → layout) exceeds
~27 KB of source body, kills the worker (empty response, usually with no
"heap memory exhausted" message in the log at all). `--heap-size` does not
help — the ceiling is identical at 8 MB and 64 MB, and gets *worse* at
256 MB (crashes even at ~50 KB in the minimal repro), consistent with
workers dying on startup at 256 MB noted elsewhere. This looks like a fixed
per-request allocation ceiling in the `--web` runtime, independent of the
configured heap size.

## Environment

- elephc v0.26.0 (release, commit `892ff162`), macOS ARM64
- PHP (`php -S`): unaffected at these sizes

## Minimal reproduction

Handler using `str_repeat` + `htmlspecialchars` + `implode` + `echo`:
responds fine up to roughly 50–80 KB of generated output, dies (empty
response) above that.

## Real-world bisection: syntax-highlighted paste page

A pastebin's syntax-highlighted view (`find()` → re-box → `highlight()`
[escape + tokenize] → `renderPaste()` → layout) makes several copies of the
paste body per request, so its effective ceiling is much lower than the
minimal repro's:

- 100-line PHP paste, 4159-byte body: served OK by the native view
- 110-line PHP paste, 4619-byte body: worker dies (empty response)
- Body-size bisection on a single worker: **26 KB OK, 28 KB crash**

`--heap-size` does not move this ceiling: identical at 8 MB and 64 MB, and
*worse* at 256 MB (crashes even at ~50 KB in the plain `str_repeat` repro —
consistent with the separately-observed fact that workers die on startup
altogether at a 256 MB heap size). The ceiling is also independent of how
the output string is built: reproduced identically with an older
`Highlighter` implementation that used a `.=` accumulator instead of
array+`implode`.

## Impact

Pastes above ~27 KB can be created and read via a raw/unprocessed view
(which serves up to ~50 KB), but the syntax-highlighted view crashes the
worker for the same content. An end-to-end regression fixture uses a 20 KB
paste specifically to stay under this ceiling.

## Expected

A single request's total allocation working set on the order of tens of KB
should not crash the worker regardless of `--heap-size`; if there is an
[...]

PR #1019, opened 2026-09-15
test(oop): cover the object-across-a-heavy-call shape, the #512 sequence

### LINKED ISSUE #512: Object returned by one class's method reads corrupted fields (or crashes) after an intervening heavy call to another class
## Summary

An object returned from a method of one class becomes unreliable — corrupted
fields, or a worker crash reading a string field — after a subsequent call to
a "heavy" method of a *different* class runs in between. This reproduces
deterministically (5/5) under `--web`, independent of what storage backend
produced the object (reproduces identically whether the returned object was
built from PDO or from a from-scratch filesystem-backed store), so it is not
a regression from a storage rewrite. It looks like the same escape-analysis/
arena family as #483/#484/#486/#498 (owned values not correctly promoted out
of the producing function's frame and later "recycled" by the next heavy
call in the same worker) but manifests as corruption/crash on read rather
than only as a leak — worth filing as its own issue given the distinct,
worker-crashing symptom, while cross-linking the family.

## Environment

- PHP 8.4: unaffected

## Minimal reproducing chain

```php
// inside a controller method; $this->db and $this->view/$this->hl are
// properties holding instances of OTHER classes
$p = $this->db->find($id, time());          // object constructed INSIDE Db::find()
// ... $p->title reads correctly here, right after the return ...
$hl = $this->hl->highlight($p->body, $p->lang);  // any "heavy" call interposed
echo $this->view->renderPaste($p, $hl);      // CRASH here, reading $p->title
```

Reproduces even with `View::renderPaste` reduced to reading a single field
(`$p->title`) and ignoring everything else. Reads of **int** fields (e.g.
`$p->createdAt`) don't crash but return a wrong value (e.g. `4294967328`
instead of the real timestamp) — so this isn't limited to string fields; the
entire object returned by the producing method becomes unreliable once a
"heavy" call (e.g. a syntax highlighter) runs in between.

## Workaround that works (but requires touching the caller)

Reconstructing a fresh object from the fields of the one returned by the
producer, immediately after receiving it and *before* calling the heavy
method in between, eliminates the crash (verified 5/5):

```php
$p = $this->db->find($id, time());
$fresh = new Paste($p->id, $p->title, $p->lang, $p->body, $p->createdAt, $p->expiresAt);
echo $this->view->renderPaste($fresh, $this->hl->highlight($fresh->body, $fresh->lang));
```

## Workarounds tried inside the producer (`Db::find()`) that do NOT work

- Defensive copy of every string field with `. ''` or `sprintf('%s', ...)`
  before constructing the returned object inside `find()`
- A static cache array that "pins" the object (avoids the crash but the
  field reads back empty — it doesn't repair the value, just changes the
  failure mode)
- Static string properties used as a cache (still crashes)
- Allocating and discarding ~10 KB of string right after constructing the
  object, to "advance" a hypothetical bump allocator (no effect)
- `--heap-size` from 8 MB up to 256 MB (no effect — not a heap-size problem)

The only fix that works requires the final (re)construction of the object to
happen in the **caller's** frame, not inside the producing method — consistent
with an escape-analysis/arena bug where values "returned" from a function are
not correctly promoted and remain tied to memory local to that function,
reusable by the next "heavy" call in the same worker.

## Cross-links

Same family as the following open leak/ownership issues — the common thread
is "values crossing a function-return boundary aren't given independent,
correctly-owned storage":

- #483 — owned heap temporaries in call->prop chains / `?->` / ternary arms never released
- #484 — `mixed_box` of an owned object leaks the producer's reference
- #486 — owned temporaries passed as mixed arguments never released by the caller
- #498 — Mixed value leaks when stored into an untyped property or array element

This issue's symptom (corruption/crash on *read*, not just a leaked
allocation) suggests it's not purely a "forgot to release" leak like the
above — it looks like the freed/reused memory from the leak family above is
what a subsequent heavy call then writes into, producing a genuine
use-after-free read rather than only unbounded growth. Filing separately
because the reproducing trigger (cross-class method call sequencing) and the
crash-on-read symptom are distinct enough to need their own repro, but this
is very likely one bug family with a common root cause in escape analysis/
value ownership at function-return boundaries.

## Notes

contest: `GET /p/{id}` deterministically returned an empty response
(curl exit 52) until the caller-side reconstruction workaround was applied.

Guikingone:
Closes #512.

**The reported symptom does not reproduce on current `main`.** Measured, following #551's revalidation plan rather than assumed:

| | |
|---|---|
| the reporter's chain as a CLI program | byte-identical to the host PHP 8.5.10 |
| the same program under `--web`, worker isolation | **200 consecutive requests, every response identical** — the sequence that used to return an empty body with curl exit 52 |
| `--web-isolation request` | 6 requests, likewise |

`test_object_returned_across_a_heavy_call_keeps_its_fields` pins the shape #512 describes: a producer method **builds** the object inside its own frame and hands it back, a heavy string-churning call to another class runs before the fields are read, and both a string and an int field are read afterwards — `$p->createdAt` came back as `4294967328` in the report, so the int side is not incidental.

It repeats, which is the part that matters. Only then is the allocator handing back blocks it has already freed, which is what a use-after-free read needs to be visible at all; a single pass reads untouched memory and would pass whatever the ownership rules did.

## Why this is a value assertion and not a heap-clean one

That fixture **does** leak — one block per `$p->body` / `$p->lang` handed straight to a string-returning callee.

That is **#1018**, filed out of this revalidation: a property read passed directly to a callee whose return-alias summary is a MAY fact has its release suppressed and never reclaimed, because `Op::ReleaseUnlessAliases` — the runtime per-call decision #619 added for exactly this — is only emitted when both the argument and the result are `Mixed`/`Union`. A plain `string` pair falls back to the unconditional skip.

It is a live bug in the same family, but it is not the shape #512 describes, and asserting a clean heap here would only re-report it under the wrong name.

The two are plausibly connected the other way round: a per-request leak in a persistent worker is a route to the heap exhaustion behind #512's original empty response, which is also why #1018 cross-links #515. #1018 carries the full alloc/free matrix, the reason `trim`/`substr`/`ucfirst` leak while `strtoupper`/`strrev` do not, and the reason a naive widening of that gate would turn the leak into a use-after-free rather than fix it.

#551 can tick #512 and pick #1018 up as its successor.


nahime0 (approved):
test only, merge and forget

Issue #1031, opened 2026-09-15
AOT ReflectionClass::getMethods() followed by eval() reflection on the same class hangs forever


Guikingone:
## Summary

A program that reflects a class **both** through `ReflectionClass::getMethods()` in AOT code and
through `eval()` afterwards never terminates. It spins forever with no output, no diagnostic and
no exit — the buffered output produced before the loop is lost with it.

PHP runs the same program to completion.

## Repro

```php
<?php
interface CasingContract { public function RunIt(int $Times): void; }
trait CasingHelper      { public function HelpMe(): void {} }
class CasingBase        { public function BaseThing(): void {} }

class CasingBox extends CasingBase implements CasingContract {
    use CasingHelper;
    public function Match(): void {}
    public static function StaticOne(): void {}
    public function RunIt(int $Times): void {}
}

echo (new ReflectionMethod(CasingBox::class, "mAtCh"))->getName(), "|";
echo (new ReflectionClass(CasingBox::class))->getMethod("MATCH")->getName(), "|";
echo (new ReflectionMethod(CasingBox::class, "STATICONE"))->getName(), "|";
echo (new ReflectionMethod(CasingBox::class, "basething"))->getName(), "|";
echo (new ReflectionMethod(CasingBox::class, "helpme"))->getName(), "|";
echo (new ReflectionMethod(CasingContract::class, "runit"))->getName(), "|";
echo (new ReflectionMethod(CasingHelper::class, "HELPME"))->getName(), "|";

$names = [];
foreach ((new ReflectionClass(CasingBox::class))->getMethods() as $method) {
    $names[] = $method->getName();          // <-- REMOVE THIS LOOP AND IT TERMINATES
}
sort($names);
echo implode(",", $names), "|";

eval('echo (new ReflectionMethod("CasingBox", "mAtCh"))->getName(), "|";
$c = new ReflectionClass("CasingBox");
echo $c->getMethod("match")->getName(), "|";
$n = []; foreach ($c->getMethods() as $m) { $n[] = $m->getName(); } sort($n);
echo implode(",", $n), "|";
echo implode(",", get_class_methods("CasingBox")), "|";
class EvalCasingBox { public function EvalMatch(): void {} }
echo (new ReflectionMethod("EvalCasingBox", "evalmatch"))->getName();');
echo "\n";
```

- **PHP 8.5.10**: runs to completion.
- **elephc `6e6f02611f`** (macos-aarch64, debug): prints the seven direct rows, then spins.
[...]

PR #1032, opened 2026-09-15
fix(reflection): report a method's declared spelling, not the lookup text

### LINKED ISSUE #571: Bug: ReflectionMethod::getName() loses declared method casing
## Summary

AOT reflection resolves PHP method names case-insensitively, but `ReflectionMethod::getName()` does not consistently return the method's declared spelling.

For a method declared as `Match`:

- direct `new ReflectionMethod(..., "mAtCh")` returns the constructor lookup spelling, `mAtCh`;
- `ReflectionClass::getMethod("MATCH")` returns the lowercase lookup key, `match`.

PHP returns the declared name, `Match`, from both paths.

## Reproduction

```php
<?php
class ReflectionMethodCasingBox {
    public function Match(): void {}
}

$direct = new ReflectionMethod(ReflectionMethodCasingBox::class, "mAtCh");
$viaClass = (new ReflectionClass(ReflectionMethodCasingBox::class))->getMethod("MATCH");

echo $direct->getName(), ":", $viaClass->getName();
```

## Observed behavior

```text
mAtCh:match
```

## Expected behavior

PHP 8.4.19 prints:

```text
Match:Match
```

Method lookup remains case-insensitive, but reflected metadata must retain and expose the declaration spelling.

This affects ordinary identifiers as well as methods whose names are PHP keywords; it is not specific to keyword tokenization.

## Likely area

The direct AOT path in `src/codegen/lower_inst/objects/reflection.rs` normalizes the method name for lookup, but `reflection_method_metadata()` passes the caller-supplied `method_name` into `reflection_method_owner_metadata()` as the reflected name. The `ReflectionClass::getMethod()` path appears to expose the normalized method key instead.

The fix should keep a case-insensitive lookup key while sourcing `reflected_name` from retained declaration metadata.

## Tests to add

- direct `ReflectionMethod` construction with differently-cased lookup text;
- `ReflectionClass::getMethod()` with differently-cased lookup text;
- a normal identifier and a keyword-named method;
- inherited/interface/trait method declarations where applicable;
- consistency between AOT and eval reflection paths.

## Notes

This is distinct from #286, which tracked the previously missing `ReflectionClass::getName()` API.

Guikingone:
Closes #571.

```php
class Box { public function Match(): void {} }

new ReflectionMethod(Box::class, "mAtCh")            ->getName();   // mAtCh
(new ReflectionClass(Box::class))->getMethod("MATCH")->getName();   // match
```

PHP answers `Match` from both. Method lookup **is** case-insensitive, so both find the method; what they reported afterwards was wrong in two different ways — one echoed the caller's own lookup text, the other the lowercase key.

## The fix

The declaration is the only answer that is a property of the program rather than of how it was queried, and it is already retained. `method_decls` holds a class's or interface's source declarations, flattened trait methods included, so `reflection_declared_method_name` reads it back instead of keeping a second copy that could drift from the declaration. A trait's *own* declarations are not reachable from a using class, so `TraitMethodInfo` gains the one field that was missing.

## Every path, not the two in the report

Each row below is a separate retention path, and all eight were wrong:

| | before | after / PHP |
|---|---|---|
| `new ReflectionMethod(Box::class, "mAtCh")` | `mAtCh` | `Match` |
| `ReflectionClass::getMethod("MATCH")` | `match` | `Match` |
| …a **static** method, looked up `"STATICONE"` | `STATICONE` | `StaticOne` |
| …an **inherited** method, looked up `"basething"` | `basething` | `BaseThing` |
| …a **trait** method used by the class, `"helpme"` | `helpme` | `HelpMe` |
| …an **interface's** own declaration, `"runit"` | `runit` | `RunIt` |
| `ReflectionClass::getMethods()` | all lowercase | all declared |
| `getPrototype()`, `getParameters()[n]->getDeclaringFunction()` | lowercase | declared |

## The eval bridge, which the issue asks to agree

The generated `_eval_reflection_methods` table stored the lowercase key, and the interpreter lowercased the caller's text on top of it — so one program answered `Match` in compiled code and `match` one line later inside `eval()`.

The table now records the declaration. That is safe because **every reader compares with `__rt_strcasecmp`**: the stored spelling decides what is *reported*, never which row a lookup finds. `get_class_methods()`, `ReflectionClass::getMethods()`, `getMethod()` and the `ReflectionMethod` constructor inside `eval()` all agree with the AOT answer now. A class declared *inside* eval is untouched — eval's own metadata always kept its spelling, and it is what the AOT rows now match rather than a second thing to fix.

## Three existing tests pinned the defect

`test_eval_reflection_method_accepts_object_targets`, `..._accepts_single_method_string` and `..._create_from_method_name` each asserted `aotrun` for a method declared `aotRun`, sitting next to an eval-declared `MiXeDCase` they asserted correctly — the inconsistency the issue reports, written down in the suite. A fourth, `test_reflection_class_get_methods_preserves_parameter_metadata`, filtered on `getName() === "traitlisted"` for a method declared `traitListed`, so it silently stopped matching anything once the name was right. All four now expect what PHP prints (verified against the host PHP 8.5.10 for the trait case).

## Measured

Both regressions assert the host PHP 8.5.10 output for the same fixture:

```
Match|Match|StaticOne|BaseThing|HelpMe|RunIt|HelpMe|Match|BaseThing,HelpMe,Match,RunIt,StaticOne|RunIt|RunIt
Match|Match|Match,StaticOne|Match,StaticOne|EvalMatch
```

Suites: `codegen::oop` 617, `reflection`-matching 284, `-p elephc-magician` 1194 — clean.

## Not in this PR

- **`ReflectionMethod::__toString()` returns `""`** — the method is unimplemented, not mis-cased. Out of scope here.
- **#1031** — an AOT `ReflectionClass::getMethods()` listing followed by `eval()` reflection on the same class hangs forever. Found while building the fixture; bisected against an unmodified `origin/main` build, so it predates this change and is not caused by it. The eval regression here is shaped to avoid that combination, and the issue carries the reproducing program and the four-way bisect.

## Docs

`docs/php/classes.md` states the rule on the `ReflectionMethod::getName()` row and lists the paths that agree. `docs/internals/eval-runtime.md` records the cross-layer invariant: the generated table carries the declaration, the readers compare case-insensitively, and the interpreter must not re-normalize a name it took from the table.


nahime0:
Follow-up: AOT `ReflectionMethod::__toString()` is still unimplemented (returns `""`). This PR correctly left it out of scope.

Tracked in #1080.

Quick split for anyone reading this later:
- **AOT** — empty string stub (`__string` never filled)
- **eval/Magician** — already formats a real dump (on `main` since #451)
- **#893** — Core builtins, not this API

nahime0 (approved):
lgtm

PR #1034, opened 2026-09-15
fix(optimizer): a generator whose yields are all dead is still a generator

### LINKED ISSUE #673: Generator whose only yield is in a constant-folded branch hangs forever
## Summary

A generator whose only `yield` sits in a branch the constant folder proves dead never
finishes: the compiled binary hangs forever. PHP runs the loop zero times and continues.

## Repro

```php
<?php
function g() { if (false) { yield 1; } return; }
foreach (g() as $v) { echo $v; }
echo "done\n";
```

PHP 8.4:

```
done
```

elephc (`fix/security-round-0804`, macOS ARM64) — no output, never exits:

```
$ perl -e 'alarm 15; exec @ARGV' ./g38
$ echo $?
142        # 128 + SIGALRM, i.e. the alarm killed it
```

## Why it happens

`if (false)` is folded away before generator lowering, so the function body no longer
contains a `yield`. The function is still classified as a generator (correctly — PHP
decides that syntactically, before any folding), but the coroutine it lowers to has no
resume point that can ever reach completion, so the driving `foreach` spins.

This makes the two classifications disagree: `yield`-presence *for generator detection*
must be decided on the pre-folding AST, while the lowered body reflects the post-folding
one. Whichever side is chosen, both must agree.

## Expected

`foreach` over such a generator iterates zero times and the program continues, as in PHP.
A generator that syntactically contains `yield` but whose `yield` is unreachable is still
a valid, immediately-complete generator.

## Note

`function g() { return; yield; }` — the idiomatic "empty generator" — is the same shape
without the folder involved and is worth covering in the same fix.

Guikingone:
Closes #673.

```php
function g() { if (false) { yield 1; } return; }
foreach (g() as $v) { echo $v; }
echo "done\n";
```

PHP prints `done`. The compiled binary printed nothing and never exited.

## Why

PHP decides whether a declaration is a **generator** syntactically, before any folding, and the checker does the same — it types `g()` as `Generator` from the `yield` it can see. The AST passes run *after* checking, and both the propagation and pruning walks stop rewriting a block at its first terminator, so an unreachable `yield` was dropped.

The two classifications then disagreed: the caller still drove a `Generator` while lowering saw an ordinary function and emitted one returning boxed null, so `iter_start`/`iter_next` ran against a null and the `foreach` spun.

## Wider than reported

The issue's note turns out to be the same bug, not a companion: the idiomatic empty generator hangs too, and it needs no folder at all.

```php
function g() { return; yield; }        // also hangs
```

So do the method, static-method and closure forms of both. Eight hanging shapes in total, destroyed by two different passes — constant-branch pruning for the `if (false)` form, the stop-at-terminator rule the propagation pass applies first for `return; yield;`.

## The fix, in two places because the defect has two halves

**`optimize::generator_bodies::rewrite_preserving_yield`** refuses a rewrite that would drop a body's last `yield`. Every pass that rewrites a callable body — propagation and pruning, for functions, methods and closures — goes through it. Keeping the un-rewritten body is conservative: the retained `yield` is unreachable by construction, so the cost is one unexecuted statement in a body that is now lowered as the coroutine it is, and only a body whose every `yield` is dead pays it.

**`attach_generator_source_if_needed`** now reads the *checker's* return type instead of `function.return_php_type`. Those two differ precisely for a generator: `generator_body_return_type` has already rewritten the body's type to `Mixed`, because a generator's compiled body is a coroutine whose `return` feeds `getReturn()`. Testing the rewritten type threw away the very evidence the decision needs, leaving only the `yield` scan.

Both are load-bearing — with either one alone the fixture still fails. That was measured by reverting each in turn, not assumed.

## Measured

Eleven generator shapes plus `iterator_to_array` and manual driving, byte-identical to the host PHP 8.5.10 — including the rows that must **not** change:

| | |
|---|---|
| `folded`, `idiomatic`, declared `: Generator` | `[]` `ret=NULL` |
| `dead_after_return` | `[]` **`ret=7`** — a value returned before the dead `yield` survives |
| `method`, `static`, `closure`, `closure2` | `[]` `ret=NULL` |
| `live` | `[1,2]` — a real generator is untouched |
| `conditional(false)` / `conditional(true)` | `[]` / `[1]` — a reachable `yield` still runs |

The test hanging *is* the regression: before the fix the binary never exits.

Suites: `--bin elephc` 1915, `error_tests` 1534, `codegen::generators` 83, `codegen::optimizer` 366, `codegen::callables` 455, `codegen::control_flow` 230, `codegen::fibers` 95, `codegen::oop` 615 — all clean.

## Docs

`docs/internals/the-optimizer.md` gains "A generator must still look like one", stating the invariant and why the conservative answer is the right one. `docs/php/generators.md` documents the user-visible rule next to the control-flow section.


nahime0:
## Grok review (READ-ONLY)

**Verdict: changes requested — please do not merge as-is.**

CI is green and the #673 hang shapes in the fixture look fixed. The new lowering heuristic for “this callable is a generator” is still the wrong source of truth and can miscompile valid PHP.

### What this PR changes

PHP (and the checker) treat a callable as a generator if the body **syntactically** contains `yield` / `yield from`, including unreachable ones. After typecheck, `propagate` / `prune` stop at the first terminator and can delete that `yield`. Callers still type the result as `Generator`; lowering used to emit a normal function returning boxed null → `foreach` spun forever.

Two halves in the patch:

1. **`optimize::generator_bodies::rewrite_preserving_yield`** — clone body, run prune/propagate, keep the original if the rewrite would leave **zero** yields. Wired into function / method / closure rewrite sites in propagate and prune (including `prune_expr` on `Closure`).
2. **`attach_generator_source_if_needed`** — now keys off the checker `signature_return`, not `function.return_php_type`. For a generator, `generator_body_return_type` has already rewritten the EIR body type to `Mixed` (`getReturn()`), so the old test could only see the yield scan.

Docs + `tests/codegen/generators/control_flow.rs` cover `if (false) yield`, `return; yield;`, `return 7; yield`, `: Generator`, live/conditional yields, method/static/closure, `iterator_to_array`, and manual `valid()`/`current()`.

### Blocker: `return_type == Generator` ≠ “is a generator function”

Checker `Generator` means two different things:

- body had `yield` (real generator; may later be DCE’d)
- declared `: Generator` and **returns** a `Generator` object, no `yield` (factory)

```php
function factory(): Generator { return inner(); }  // NOT a generator function
function inner() { yield 1; }
foreach (factory() as $v) { /* must see 1 */ }
```

After this PR, `attach_generator_source_if_needed` sees `signature_return == Generator` and no yield, sets `is_generator`, and the public symbol becomes a constructor. `return inner()` becomes `getReturn()`; foreach is empty. Pre-PR this failed closed and stayed a normal returning function.

**Fix before merge:** persist a syntactic `is_generator` bit on `FunctionSig` at check time (`body_contains_yield` before opt). Lowering should read that flag — do not treat `return_type == Generator` as generator-ness. Add a factory regression as above.

Tracked as #1086.

### Other notes (non-blocking follow-ups)

- Docs say every body rewrite goes through `rewrite_preserving_yield`. **DCE does not** (`dce_block_with_guards` / `dce_method` are unwrapped). Named functions can still lose their last yield there and today only recover via the unsound `signature_return` proxy. → #1084
- Fixture gaps (same mechanism, not new logic): `while (false)`, dead `switch`, `throw; yield`, dead `yield from`, trait methods. → #1085
- Whole-body revert is conservative (one dead yield blocks *all* prune/propagate in that callable). Acceptable; a surgical keep-one-yield would be cleaner later.
- `rewrite_preserving_yield` comment says “last yield”; code keeps the original iff **any** yield disappears entirely. Behavior is right; wording is slightly off.
- `GeneratorSource` is written and not read elsewhere — leftover metadata, not a hang.

### What looks fine

- Root-cause match with #673.
- Closure vs named-function split is real; both halves are load-bearing for the fixture as written.
- `body_contains_yield` scope rules match PHP (sees yield/yield from, skips nested functions/closures, walks if/loops/try/switch).
- Live generators and `getReturn()` before a dead yield are asserted and must not change.
- Module preamble / focused codegen test match repo policy.

### Follow-ups opened

| Issue | Topic |
|-------|--------|
| #1086 | **Blocker:** syntactic `is_generator` bit; stop inferring from `return_type == Generator` |
| #1084 | Wrap DCE with `rewrite_preserving_yield` (or fix the “every pass” docs claim) |
| #1085 | Extra dead-yield shapes (`while(false)`, switch, throw, `yield from`, traits) |

Guikingone:
Thank you — this was a genuinely good catch, and the blocker was a regression I introduced. Fixed in `89b5957154`.

**Reproduced your factory case exactly** before changing anything:

```
PHP:      factory: 1,2,   inner: 1,2,
this PR:  factory:        inner: 1,2,
```

Your diagnosis was right in every part: `attach_generator_source_if_needed` saw `signature_return == Generator` with no `yield`, set `is_generator`, and the `return inner()` lowered to `Generator::getReturn()`. And as you say, pre-PR this failed closed — so the fallback traded one wrong answer for another, which is the wrong direction.

**Fixed as you prescribed.** `FunctionSig` now carries `is_generator`, taken from `body_contains_yield` on the SOURCE body at check time, before any pass runs — so an unreachable `yield` still counts, which is PHP's rule. Free functions, methods and closures all record it. Lowering reads that bit and **no longer consults the declared return type at all**; the body scan that remains beside it can only ADD generator-ness, never withhold it.

Adding the field meant touching 38 `FunctionSig` literals; everything outside the three declaration sites is `false` (builtins, externs, synthesized signatures), which the compiler enforced site by site.

**Regression added**, covering the five shapes that must disagree with each other — a plain factory, one through a method, a static one, a closure one, and a real generator that also declares `: Generator` and must stay one:

```
factory=1,2, declared=8,9, method=1,2, static=1,2, closure=1,2,
```

byte-identical to host PHP 8.5.10. I also re-introduced the old proxy temporarily to confirm the test is load-bearing — four of the five come back empty:

```
left:  "factory= declared=8,9, method= static= closure="
right: "factory=1,2, declared=8,9, method=1,2, static=1,2, closure=1,2,"
```

**Both non-blocking notes are also addressed:**

- The `rewrite_preserving_yield` comment said "last yield" where the code keeps the original iff EVERY yield disappears. Wording corrected; behaviour was already right, as you said.
- The optimizer page claimed every body rewrite goes through the helper. It does not — DCE is unwrapped. Corrected, and the consequence is now stated rather than implied: mis-classification is no longer possible with the bit recorded, but the body DCE hands to lowering can still have lost its last `yield` (**#1084**). I also recorded your whole-body-revert observation there as the tighter thing to do later.

**Not addressed, deliberately:** the fixture gaps in **#1085** (`while (false)`, dead `switch`, `throw; yield`, dead `yield from`, trait methods). Those are the same mechanism rather than new logic, and with generator-ness now recorded before any pass, none of them can mis-classify — so they are coverage rather than correctness. Happy to fold them in here if you would rather they land together.

Suites: `generators` 87, `optimizer` 366, `callables` 457, `-p elephc --lib` 1684.

PR #1039, opened 2026-09-16
fix(parser): ++/-- on a property or element reads as an expression

### LINKED ISSUE #682: ++/-- on a property or array element is rejected in expression position
## Summary

`++` and `--` on a **property** or an **array element** are only accepted in statement
position. In any expression position — `return`, `echo`, the right-hand side of an
assignment — the parser rejects them. On a plain local variable the same operators work
everywhere, so the gap is specific to non-variable lvalues.

## Repro

```php
<?php
class Counter {
    public int $n = 0;
    public function next(): int { return $this->n++; }
}
$c = new Counter();
echo $c->next(), $c->n;
```

PHP: `01`

elephc:

```
error[3:60]: Expected ';'
```

## Perimeter

Statement position — accepted:

| Snippet | Result |
| --- | --- |
| `$this->n++;` | OK |
| `++$this->n;` | OK |
| `$b[0]++;` | OK |
| `++$b[0];` | OK |

Expression position — rejected:

| Snippet | Message |
| --- | --- |
| `return $this->n++;` | `Expected ';'` |
| `return $this->n--;` | `Expected ';'` |
| `return ++$this->n;` | `Expected variable after '++'` |
| `echo $this->n++;` | `Expected ';'` |
| `$x = $this->n++;` | `Invalid assignment target` |
| `return $o->n++;` (any object, not just `$this`) | `Expected ';'` |
| `return $b[0]++;` | `Expected ';'` |

For contrast, a plain local works in expression position:

```php
<?php
function h() { $y = 1; return $y++; }
echo h();   // 1 — accepted, and correct
```

## Root cause

The AST cannot represent the construct. The expression node carries a variable *name*
rather than an lvalue:

```rust
ExprKind::PostIncrement(name) | ExprKind::PostDecrement(name) => match env.get(name) {
```

— [`src/types/checker/inference/expr/basic.rs`:89](https://github.com/illegalstudio/elephc/blob/main/src/types/checker/inference/expr/basic.rs#L89)

Because the payload is a `String`, `$this->n++` has nowhere to live in an expression, and
the statement layer works around it with dedicated statement parsers
([`try_parse_postfix_incdec`, `try_parse_scoped_postfix_incdec` in
`src/parser/stmt/assign/postfix.rs`](https://github.com/illegalstudio/elephc/blob/main/src/parser/stmt/assign/postfix.rs)),
which is why the statement forms are accepted and the expression forms are not. The prefix
rejection comes from the same restriction, surfacing at
[`src/parser/control.rs`:498](https://github.com/illegalstudio/elephc/blob/main/src/parser/control.rs#L498).

## Fix shape

Widen the `Pre*`/`Post*` variants to hold an lvalue expression instead of a name, then let
the expression parser produce them for the property and index forms. The statement parsers
become a special case of the general node rather than the only path. Every consumer of the
current `(name)` payload has to be updated — `src/types/checker/inference/expr/basic.rs`,
`effects.rs`, `loop_storage.rs`, `src/types/return_alias.rs`, `src/hash_prelude/detect.rs`,
`src/parser/expr/assignment_targets.rs` and the EIR lowering — so this is not a local
change.

Note the evaluation-order requirement PHP imposes: the post-forms yield the value the
property held *before* the update, and the receiver must be evaluated once, not twice
(`$this->items[f()]++` may not call `f()` twice).

## Discovery

Found while verifying the merge of #678 against the pre-merge branch. Present identically
before and after that merge, so it is a pre-existing gap, not a regression.

Guikingone:
Fixes #682.

`echo $obj->n++;`, `return ++$this->count;`, `$x = ++$b[0];` and `$sum += $counts[$k]++;` were all parse errors. Three separate parser gaps, one per spelling, each with its own message.

## What was actually broken

The issue reported the whole surface, but most of it lands today — the postfix form on a property or a simple element has since been desugared. What remained:

**1. Every prefix form on a non-`$this` place.** `parse_prefix_inc_dec` took its fast path on seeing `Token::Variable` alone, consumed the name, and left the place suffix behind. `++$o->n` therefore incremented the *object*:

```
Cannot increment/decrement $o of type Object("C")
```

and `++$b[0]`, `--$b[0]`, `++$m["k"]`, `++$o->items[0]` the same way. `++$this->n` escaped only because `$this` is its own token, never a `Variable`. The fix looks at the next token too: `->`, `?->`, `::` or `[` means the real target is the larger place, so the general l-value path takes it.

**2. Any place with a computed index, in expression position.** `desugar_lvalue_incdec` replayed its target verbatim, so `can_replay_assignment_target` turned away `$b[ix()]` and the operator was left unconsumed: `Expected ';'`. It now **stabilizes** the place first, through the same `AssignmentExpressionLowerer` a compound assignment uses — which is also what makes `ix()` run once, as PHP requires, rather than once per read and once per write.

**3. A postfix increment inside a compound assignment.** `$t += $b[0]++;` has a top-level `++`, so both postfix-incdec *statement* parsers claimed the statement, parsed `$t += $b[0]` as their target, and rejected it with `Invalid assignment target`. They now decline when a top-level assignment operator comes first, because the increment belongs to the right-hand side; the assignment parsers then lower it inside the value expression, where it already worked.

## What is not this issue

Two shapes still fail, and both fail identically with no increment anywhere, so they belong to the assignment path rather than to `++`:

- **A nested array element.** `$r[0][0] = 9;` reports `Nested array assignment requires a Mixed or ArrayAccess target`, and so therefore do `$r[0][0]++` and `++$r[0][0]`. The computed-index form now reaches that real diagnostic instead of `Expected ';'`.
- **`var_export()` on a `?int` property segfaults.** `var_export($o->n, true)` alone is enough — no increment involved.

Both filed separately.

## Verified

Twenty-three shapes against host PHP 8.5.10, all byte-identical: both operators, both spellings, the receiver forms (`$this`, any object, indexed and associative elements, an element of a property, a static property, a plain local), the value each spelling yields, a computed index evaluated once, an increment on each side of a compound assignment, and the statement forms that already worked.

All three fixes were confirmed load-bearing by disabling each in turn:

| disabled | error that comes back |
|---|---|
| prefix suffix check | `Cannot increment/decrement $o of type Object("Counter")` |
| statement guard | `Invalid assignment target` |
| stabilization | `Expected ';'` |

A heap-debug fixture pins that the temporaries the rewrite binds — stabilized receiver, index, captured old value, result — are released. **Two receiver forms are deliberately outside it**: `Counter::$s += 1;` reports `allocs=50 frees=0` over 50 iterations and `$o->items[0] += 1;` reports `allocs=401 frees=351`, both with no increment anywhere and both clean under plain assignment. That leak is in the shared read-modify-write path; the test says so rather than pinning it here.

`cargo test --test codegen_tests -- codegen::operators codegen::control_flow` passes in full (359), as do the 398 parser tests, 1680 lib tests and 1534 error tests.

## Docs and example

`docs/php/operators.md` no longer says reading the result "is not supported yet"; it states both positions, shows the four expression spellings, and pins the evaluate-the-place-once rule with a fixture.

`examples/arithmetic/main.php` gains a `Tally` whose `record()` returns the count before the hit, a `+=` over a post-increment, and an index function that prints once. The whole example is byte-identical to host PHP 8.5.10.


nahime0:
## Follow-ups

| Issue | Topic |
|-------|--------|
| 1137 | Remaining `++`/`--` places: `$o->{$k}++`, nested `$r[0][0]++`, `$cls::$prop++` |
| 1138 | Shared RMW heap leaks on static-prop and property-element `+=` |
| 1139 | Parser-unit tests for the three token guards; ternary statement scan can still claim `$a[0]++` |

#1039: prefix `++$o->n` / `++$b[0]` no longer increments the object/array; expression-position desugar evaluates a computed index once; statement postfix no longer steals `$t += $b[0]++;`. `?->` is only a suffix probe — `++$o?->n` still errors rather than writing through nullsafe.

PR #1054, opened 2026-09-16
fix(strings): implode() reads its array's real element layout, not an assumed one

### LINKED ISSUE #640: implode() over a homogeneous float array prints nothing (raw doubles read through the string-slot layout)
## Summary

`implode()` over a **homogeneous float array** produces no output at all. The elements never reach the `mixed` stringification path: an all-float array keeps a raw layout, and `__rt_implode`'s non-mixed branch reads each 8-byte `double` through the 16-byte `{ptr,len}` string-slot layout, so the "length" it copies is garbage.

## Repro

```php
<?php
$r = [1.5, 2.5];
echo implode(",", $r), "\n";
```

- **PHP 8.4.23**: `1.5,2.5`
- **elephc** (`main` `ba836179c`, macOS ARM64, with or without `--ir-opt`): empty output

Also empty with three or more elements (`[1.5, 2.5, 3.5]`).

## Why it is not #614

Making the array heterogeneous fixes it, because that routes the elements through the boxed-`mixed` loop instead:

```php
$r = ["a", 1.5, 2.5]; echo implode(",", $r);   // a,1.5,2.5 — correct
$r = [1.5, "a", 2.5]; echo implode(",", $r);   // 1.5,a,2.5 — correct
$r = [1.5, 2];        echo implode(",", $r);   // 1.5,2     — correct
```

Homogeneous **int** arrays are unaffected because they have their own dedicated runtime helper (`__rt_implode_int`, `src/codegen_support/runtime/strings/implode_int.rs`). There is no float equivalent, and the generic path's element dispatch only distinguishes "boxed Mixed cells" (`value_type` tag 7) from the string-slot layout.

## Expected

`implode()` over `array<float>` should join the elements using PHP's float-to-string formatting, matching `implode(",", [1.5, 2.5]) === "1.5,2.5"`.

A fix probably means either giving raw float arrays their own runtime path (mirroring `__rt_implode_int`), or teaching the generic loop to dispatch on the element `value_type` tag rather than assuming "not Mixed" implies the string-slot layout.

## Notes

Found while fixing #614 (PR #639); verified identical on the pre-fix build, so it is independent of that change.

### LINKED ISSUE #689: `implode()` on a `?array` segfaults for int/float elements and drops bools (string elements are fine)
## Summary

Passing a value whose declared type is `?array` to `implode()` compiles and then segfaults,
unless the elements happen to be strings. The declared nullability alone is enough — no
property, no object, no null value involved:

```php
<?php
function f(?array $a): void { echo implode(',', $a); }
f([1, 2]);
```

PHP 8.4.23:

```
1,2
```

elephc: exit 139 — SIGSEGV, nothing printed.

Dropping the `?` makes the identical program correct:

```php
function f(array $a): void { echo implode(',', $a); }   // prints 1,2 ✔
```

Verified on `main` `a7a0e090c` (v0.26.3-71-ga7a0e090c), macOS ARM64. Identical with
`--ir-opt=on` and `--ir-opt=off`.

## Element type decides the outcome

```php
<?php
function f(?array $a): void { echo implode(',', $a), "\n"; }
f(["a", "b"]);   // PHP: a,b     elephc: a,b  ✔
f([1, 2]);       // PHP: 1,2     elephc: SIGSEGV
```

The same function, two call sites: the string call prints correctly and the int call crashes
immediately after it.

| Elements | PHP | elephc |
|---|---|---|
| `["a", "b"]` | `a,b` | `a,b` ✔ |
| `[1, 2]` | `1,2` | SIGSEGV |
| `[1.5, 2.5]` | `1.5,2.5` | SIGSEGV |
| `[true, false]` | `1,` | `,` (wrong, no crash) |

Strings work, ints and floats crash, bools produce empty output — the signature of a payload
being read through the string-slot `{pointer, length}` layout: string elements match it, an
int/float element's value is dereferenced as a pointer, and a bool's low value dereferences to
nothing usable.

This is the same class of defect as #640 (`implode()` over a homogeneous float array reading
raw doubles through the string-slot layout) and #675 (the 8-vs-16-byte element slot assumption
across the array builtins), but reached through the *declared nullable type* rather than the
element type, and it crashes rather than printing wrong output.

## Reached through a property too

The parameter form above is the smallest repro, but the shape that found this was a property:

```php
<?php
class C { public ?array $x = null; }
$o = new C();
$o->x = [1, 2];
echo implode(',', $o->x);     // PHP: 1,2   elephc: SIGSEGV
```

Copying into a local first does not help — the local inherits the declared type:

```php
$t = $o->x;
echo implode(',', $t);        // SIGSEGV
```

Narrowing the null away does fix it, which is a usable workaround:

```php
echo implode(',', $o->x ?? []);   // prints 1,2 ✔
```

## What still works

Everything that does not hand the whole value to `implode()` behaves correctly on the same
`?array`, so the container itself is intact:

```php
<?php
class C { public ?array $x = null; }
$o = new C();
$o->x = [1, 2];
var_dump($o->x[0]);              // int(1)     ✔
var_dump(is_array($o->x));       // bool(true) ✔
var_dump($o->x === null);        // bool(false)✔
foreach ($o->x as $v) { … }      // 1;2;       ✔
```

## A genuinely null `?array` also segfaults instead of throwing

Separate symptom on the same argument path, worth guarding in the same fix:

```php
<?php
function f(?array $a): void { echo implode(',', $a); }
f(null);
```

PHP raises `TypeError: implode(): … argument #2 ($array) must be of type array, null given`.
elephc segfaults (exit 139). That is the same missing-null-guard family as #647.

## Relationship to #630

The *other* array builtins reject a `?array` at compile time rather than crashing:

```
count($a)      → error: count() argument must be array or Countable object
array_sum($a)  → error: array_sum() argument must be array
in_array(2,$a) → error: in_array() second argument must be array
```

That refusal is the checker boundary already tracked in #630 (`mixed` values that hold an
array), since `?array` lowers to the same `Mixed` representation, and is **not** what this
issue is about. Filed here is only `implode()`, which accepts the argument and then reads it
through the wrong layout. Note that #630's own repro shape — an array from `json_decode()` —
implodes correctly today, so the crash is specific to the declared `?array` type rather than
to `Mixed`-holding-array values in general.

### LINKED ISSUE #1081: implode() on a foreach value bound from array_chunk() segfaults
## Summary

`implode()` on a `foreach` value bound from an `array_chunk()` result **segfaults** (exit 139). No diagnostic, no output — the process dies mid-loop.

## Reproduction

```php
<?php
$idx = [1, 2, 3, 4, 5];
echo "1:";
foreach (array_chunk($idx, 2) as $p => $items) { echo " " . implode(",", $items); }
echo "\n";
```

PHP: `1: 1,2 3,4 5`

elephc: prints `1:` and then dies with SIGSEGV (exit 139).

## Perimeter

The crash needs the chunk to arrive through the `foreach` binding. Reading the same chunk by index is fine.

| Shape | Result |
| --- | --- |
| `foreach (array_chunk($a, 2) as $p => $items) { implode(",", $items); }` | **SIGSEGV** |
| `$c = array_chunk($a, 2); foreach ($c as $p => $items) { implode(",", $items); }` | **SIGSEGV** |
| `$c = array_chunk($a, 2); implode(",", $c[0]);` | OK |
| `$c = array_chunk($a, 2); foreach ($c as $p => $items) { foreach ($items as $v) {…} }` | OK |
| `$c = array_chunk($a, 2); foreach ($c as $p => $items) { count($items); }` | OK |

So it is specifically `implode()` over the loop-bound inner chunk. `count()` and a nested `foreach` over the same binding are both fine, and `implode()` over the same chunk reached as `$c[0]` is fine.

## Notes

- Reproduces on an **indexed** receiver, so it has nothing to do with the associative-receiver work in #1043 — it was found while writing an example for that and verified on stock `main` with the change stashed.
- Both `preserve_keys` modes crash.
- `array_chunk()`'s result element type is `Array<Array<T>>` (or `Array<AssocArray{…}>` for the key-preserving form), so the `foreach` binding is a container read out of a container. The likely area is how that binding's storage is typed or borrowed before being handed to `implode()`, rather than `array_chunk()` itself — the same chunk read by index and imploded is correct.

## Environment

- elephc 0.27.0, `main` at `2d2af3fdc0`

Guikingone:
Fixes #689. Fixes #640. Fixes #1081.

```php
function f(?array $a): void { echo implode(',', $a); }
f([1, 2]);        // PHP: 1,2      elephc: SIGSEGV
f([1.5, 2.5]);    // PHP: 1.5,2.5  elephc: SIGSEGV
f([true, false]); // PHP: 1,       elephc: ,
f(["a", "b"]);    // PHP: a,b      elephc: a,b   <- the only one that worked
```

Dropping the `?` made the identical program correct, which is the tell: the declared
NULLABILITY is what boxes the operand, and a boxed operand tells the renderer nothing
about its elements.

## What the renderer was assuming

`__rt_implode` dispatched on exactly one element layout: boxed Mixed cells (tag 7) took a
per-element cast, and **everything else** was read as a 16-byte `{pointer, length}` string
slot. A raw scalar element is 8 bytes and carries no length, so an int's VALUE was
dereferenced as a pointer, a double's bits likewise, and a bool pair read as one
`{ptr, len}` produced a zero length — `1,` became `,`.

Strings matched the assumption, which is why the bug looked type-specific rather than
layout-specific.

Statically typed arrays escaped it because the LOWERING picks a renderer from the element
type: `array<int>` gets `__rt_implode_int`, `array<bool>` gets `__rt_implode_bool`. A
`mixed` or union operand has no element type to pick from, so it fell through to the
string renderer.

## The layout is written down; it just was not read

Every indexed array carries its element `value_type` in the packed kind word at
`[ptr - 8]` — the same field the tag-7 check already read. `__rt_implode` now dispatches
on all of it: 0 int, 2 float, 3 bool, 7 boxed Mixed, string slot otherwise.

The scalar arms format through `__rt_itoa`/`__rt_ftoa` and obey the CURSOR invariant this
emitter already documents: the live destination is published as `_concat_off` before each
conversion, because those helpers format into the very buffer the join is filling. Without
it the next element would be written over the separator just copied. `false` renders as the
empty string rather than `"0"`, so it skips the formatter entirely.

That also fixes **#640**, the same mechanism reached from the other side: a homogeneous
`array<float>` had no renderer at all (`unsupported EIR backend feature: implode array
element PHP type Float`) and now goes through the generic one's float arm. A dedicated
`__rt_implode_float` would have had to re-implement the cursor discipline the generic loop
already has.

## The header has to be TRUE, which it was not for appends

`__rt_array_push_int` takes float and bool payloads too — same 8-byte slot — and stamps
`value_type` 0 on the first append, which says "int". An array built by appending therefore
described itself wrongly:

```php
$f = []; for (...) { $f[] = $n / 8; }
implode(',', $f)   // 4593671619917905920,4598175219545276416, ...
```

so the append restamps the header for float and bool elements. When the checker still sees
the array as empty — `$a = []` appended inside a loop keeps its `array<never>` back-edge
type at every push — the value being written is the only description, and that is what gets
stamped.

## The other half of the argument path

A boxed operand can also hold a HASH, or no array at all, and both were read as an indexed
array: a hash printed its header words, and a genuinely null `?array` segfaulted where PHP
raises a TypeError.

`emit_boxed_implode_array_source` decides at run time. A hash is converted to its values
exactly as a statically typed one is; an indexed array is retained. Both branches answer
with an OWNED array, which is what lets the join site release its operand with one
unconditional decref instead of a run-time flag.

The TypeError is raised from a **separate** guard emitted before the join site parks the
glue on the stack or takes any reference — from there the frame is what the surrounding
`try` left, so the throw is catchable and leaks nothing.

PHP's messages are reproduced exactly, including that null is worded against the
string-separator overload, that everything else is worded against the declared `?array`,
and that a boolean is named by its VALUE (`true given`, never `bool given`). An object is
named by its class, read from the same metadata table `get_class()` uses.

## The throw has to be visible to the optimizer

`implode()` is otherwise pure, so the first working guard raised its diagnostic and nothing
caught it: a `try` installs no handler around a call that cannot throw, and an unused call
is eliminable along with its diagnostic. The effect summary is now decided per call site —
`MAY_THROW` only when the operand is boxed, which is the only way the guard can fail. A
statically typed array operand keeps the pure summary and stays eliminable.

## Verified

Against host PHP 8.5.10, byte-identical: the issue's four element types plus empty,
heterogeneous, `PHP_INT_MAX`, `-0.0` and exponential floats; through `?array`, `mixed`,
`array|string`, a property, a local copy and the `?? []` workaround; the keyed and
sparse-key forms; the one-argument `join()` form; a three-byte separator; arrays built by
appending in a loop; and all seven TypeError spellings, caught and printed.

`--heap-debug` reports `leak summary: clean` over 64 iterations for both the retained
indexed payload and the converted hash values.

The eleven new tests were confirmed load-bearing by reverting the source changes: all
eleven fail, seven of them by aborting the test process.

`cargo test --test codegen_tests -- codegen::strings codegen::arrays codegen::runtime_gc
codegen::array_basics` passes in full (1345), as do 1915 unit tests and 1534 error tests.
The two emitter unit tests that pin the cursor discipline now count one publish per
formatting arm plus the finalizer, named rather than a literal.

## Known, unchanged, and not this

A caught `TypeError` leaks one block per throw. It is not this guard's: `count()`'s own
runtime TypeError leaks identically on `main`, so it belongs to the exception-object
ownership family.

## Docs and example

`docs/php/strings.md` states that `implode()` takes an indexed or associative array of any
element type, including one reached through a slot whose shape is only known at run time,
and that a non-array throws `\TypeError`.

`examples/string-ops/main.php` joins ints, floats, bools and a hash through a `?array`
parameter beside the existing `explode`/`implode` pair, and shows the TypeError being
caught. Its output is verbatim host PHP 8.5.10.


Guikingone:
I tried to rebase this onto current `main` and stopped, because it is not a mechanical rebase — and the attempt surfaced something worth recording before anyone else tries.

**The defect is still live.** On `main` at `2d2af3fdc0`:

```php
function f(?array $a): void { echo implode(',', $a), "\n"; }
f(["a", "b"]);   // a,b
f([1, 2]);       // SIGSEGV (exit 139)
```

So the PR is still needed; it just cannot be replayed as-is.

**Why the rebase is not mechanical.** `main` has since reworked the same emitter (the
grow-into-an-owned-block work for #515): the destination can now be either the shared scratch
or a GROWN heap block, and every cursor publish branches on which. This branch adds four
formatting element arms to the same loop. The emitter halves merge cleanly — git only flags the
assertions — but the merged result is not correct:

```
sub x14, x9, x14      (arm64)   5 occurrences   ← one per formatting arm + finalizer
ldr x14, [sp, #88]    (arm64)   2 occurrences   ← only main's two sites
```

The publish happens at all five sites, but the **grown-destination fallback only exists at
two of them**. The arms this branch adds were written against the pre-rework emitter, where
there was only ever a scratch destination, so they restore the wrong offset once a join has
grown. Making the assertions agree would hide that, not fix it.

**What it needs**: the four added arms re-derived on top of the current two-destination
discipline, then the cursor tests re-pinned against the real counts. That is implementation
work on the emitter, on both architectures, not a conflict resolution — so I left the branch at
its pre-rebase state rather than push a half-merged one.

Every other PR in this batch is now rebased onto `main` and mergeable.

Guikingone:
Re-derived on top of `main`'s two-destination cursor discipline and pushed; this branch is mergeable again.

**What the conflict actually was.** This branch added three formatting element arms (raw int, raw float, raw `true`) and gave them their own cast prologue, which open-coded a scratch-only `_concat_off` publish — correct against the emitter as it stood when the branch was written, where the destination was always the shared scratch. `main` has since made the destination either the scratch or a GROWN owned block, and routed both of its own publish sites through a shared `emit_implode_publish_offset_{aarch64,x86_64}` helper that branches on which.

The merged result kept both halves and was internally inconsistent — visible as the three assertions disagreeing:

```
sub x14, x9, x14      5   ← every arm publishes
str x14, [x13]        5   ← every arm stores
ldr x14, [sp, #88]    2   ← only main's two sites have the grown arm
```

**The fix**: the two added prologues now delegate to the shared helper instead of open-coding, so all five sites (four formatting arms + the finalizer) obey the same rule. The cursor tests assert `FORMATTING_ARMS + 1` on all three strings, which is what makes a future open-coded site fail loudly rather than skew one count.

**One correction to my earlier comment on this PR**, where I said the added arms "restore the wrong offset once a join has grown". I went looking for a fixture that observes that and could not build one, so here is what I actually measured. The open-coded form publishes `cursor - _concat_buf`; `__rt_itoa` then writes at `_concat_buf + offset`, so the arithmetic cancels back to the real cursor, and the `MIXED_CAST_HEADROOM` reservation that runs first keeps the write in bounds. Joins of 108 KB of ints, 148 KB of floats and 80 KB of bools — all well past the 64 KiB scratch, all through the formatting arms — are byte-identical to PHP with the open-coded form *and* with the helper.

So this is not a live miscompile I fixed; it is a shared invariant these three arms were silently opting out of, working only by an accident of the offset arithmetic. Going through the helper is still the right resolution — but the PR should not claim a bug it does not have.

**Verification**
- Issue #689's repro, which is still live on `main`: `f([1, 2])` through `?array` segfaults there, prints `1,2` here. All four element kinds match host PHP 8.5.10.
- The grown-destination joins above, byte-identical to PHP.
- `strings/implode_boxed.rs` 9/9, `runtime_gc/implode_boxed_operand.rs` 2/2, implode unit tests 7/7, `codegen::strings` 361, `-p elephc --lib` 1684 — all green.

Guikingone:
Rebased onto `main` @ `bf5437b687`, plus one test for a shape this PR fixes that was not pinned: an `(array)` cast of a `mixed` **return value**.

```php
function f(): mixed { return [1, 2]; }
echo implode(",", (array) f());
```

On a build of `main` that program exits **139 (SIGSEGV)** — the int elements' values are dereferenced as pointers, so there is no wrong output to notice before the process dies. With this branch it prints `1,2`, and all seven layouts (`int`, `float`, `bool`, `string`, empty, single, signed) match host PHP 8.5.10 byte for byte.

Worth recording where I ran into it: PR #968's branch carries a `tests/implode_element_layout_tests.rs` asserting exactly these seven cases, and its CI fails there with `binary failed (exit None)` — it has the test but not this PR's runtime change. That file belongs with this work, so I have pinned the same coverage here in `tests/codegen/strings/implode_boxed.rs` and will drop the stray copy from #968.

Verified locally after the rebase: every test matching `implode` 42/42, `codegen::strings` 362/362, `codegen::arrays` 578/578, `runtime_gc` 303/303.

PR #1064, opened 2026-09-17
fix(runtime): implode() renders a boxed object through its own __toString

Guikingone:
An object element of a boxed array joined as the **empty string** — silently, with no warning
and no crash:

```php
class S { public int $n = 0; public function __toString(): string { return "S" . $this->n; } }

$k = count($argv) - 1;      // opaque, so the element really is a boxed Mixed
$a = [pick($k), pick($k)];  // pick() returns an S

echo implode(",", $a);
// php 8.5.10: S7,S7
// elephc:     ,
```

## Why it happened

`implode()` converts each element with `__rt_mixed_cast_string`, whose tag dispatch covered
int/string/float/bool/array/resource and sent everything else to the "unsupported → empty
string" arm. Objects landed there.

What made it easy to miss is that **every other string context was already correct**. For the
very same value, in the same program:

| context | result |
| --- | --- |
| `echo $m` | `S7` |
| `(string)$m` | `S7` |
| `"x" . $m` | `xS7` |
| `"v=$m"` | `v=S7` |
| `sprintf("%s", $m)` | `S7` |
| `implode(",", [$m])` | *(empty)* |

`sprintf` is the instructive one: it converts through a sibling helper that already resolves
`__toString` from the dense `_class_tostring_ptrs` table. Only `implode`'s helper lacked the arm.

## The fix

The new tag-6 arm resolves `__toString` through that same dense table, so a boxed object now
renders identically in `implode()` and `sprintf("%s")`. Inherited methods resolve too — `Child
extends S` prints `S9`.

**Ownership follows tag 1 (string), not tag 9 (resource).** `__toString` returns storage
belonging to the *method*, so the arm persists it and releases the original when persist hands
back a different block — the same dance `__rt_sprintf_mixed_string` performs. Over 200
iterations allocations and frees balance exactly and the heap summary is clean.

## Known remaining divergence, deliberately not pinned

A class publishing **no** `__toString` keeps the empty-string result, where PHP raises
`Error: Object of class X could not be converted to string`. This runtime cannot throw that
yet, and the behaviour is unchanged from `main`, so the fixture leaves that case out rather
than recording elephc's answer as though it were PHP's.

## Also in here

The resource-arm unit test sliced from its own label to a **named later one**, so any arm
inserted between the two was measured as part of the resource arm. It now stops at whatever
label comes next, which is what it meant all along.

## Verification

- Byte-identical to PHP 8.5.10 on every fixture above.
- Both regression tests fail without the fix: `||42` where `S7|S9|42` is expected, and `1000`
  where `1600` is expected.
- The heap fixture uses the **loop counter** as its opaque selector, not `count($argv)`:
  reading `$argv` materializes it and leaves two live blocks at exit, which would drown the
  signal the fixture exists to read.
- `cargo test --lib -p elephc`: 1680 pass. All 29 implode codegen tests pass.
- The x86_64 arm assembles clean for `x86_64-unknown-linux-gnu`.


Guikingone:
Addressed in `34c93852a5`. `examples/magic-methods` already declares a `__toString()`, so it gains the shape this PR fixes rather than getting a new example of its own: an object sitting beside scalars in one heterogeneous array, rendered by `implode()` through the same conversion the interpolation, concatenation and `(string)` casts beside it use.

```
implode: plain, 42, @ada, 2.5, 1
interp:  @ada
concat:  @ada
cast:    @ada
```

Before the fix that first line did not render the element through `__toString()` — which is exactly what makes it worth showing next to the three spellings that always worked. Output byte-identical to host PHP 8.5.10, and `lowers_examples_corpus` is green.

Guikingone:
The two red shards (`Codegen Tests linux-aarch64 8/16` and `linux-x86_64 8/16`) were mine, not a flake — same test on both arches:

```
codegen::objects::magic_methods::test_example_magic_methods_compiles_and_runs
assertion `left == right` failed
  left: "…static create(nahime)\nimplode: plain, 42, @ada, 2.5, 1\ninterp:  @ada\nconcat:  @ada\ncast:    @ada\n"
 right: "…static create(nahime)\n"
```

The review-response commit extended `examples/magic-methods/main.php` with the four `__toString()` contexts, and that test asserts the example's **entire** stdout. `lowers_examples_corpus` — the check I did run — only verifies the example lowers, not what it prints, which is why this only surfaced in CI.

The compiler output was right: the actual string matches host PHP 8.5.10 byte for byte. Only the expectation was stale, and it is now one `concat!` entry per line so the next edit to the example points at the line it has to add.

Also rebased onto `main` @ `bf5437b687`.

Verified locally after the rebase: `magic_methods` 25/25, `codegen::objects` 254/254, every test matching `implode` 32/32, `builtins_misc` 36/36, `lowers_examples_corpus` green.

nahime0:
Follow-ups from leftover gaps after this change:

- #1265 — Throw `Error` when a boxed object with no `__toString` is cast to string via `__rt_mixed_cast_string` (today empty; `sprintf("%s")` already throws).
- #1266 — Resolve Magician/eval objects in `__rt_mixed_cast_string` the way `sprintf` does (negative / out-of-range / null-slot class ids).
- #1267 — Free `__rt_mixed_cast_string` persist allocations on non-implode callers (`array_diff_assoc` / `array_intersect_assoc`, eval string-bytes bridge, callback argument coercion). Includes the ownership gap noted in https://github.com/illegalstudio/elephc/pull/1064#discussion_r4049408438.
- #1268 — Size implode mixed-cast headroom for an arbitrary `__toString` writing through `_concat_buf` (64-byte reserve matches itoa/ftoa; related #1074 is concat_grow leaks, not this).

None of these block this PR.

Issue #953, opened 2026-09-10
__rt_cstr copies unbounded PHP string lengths into a 4096-byte scratch buffer: silent out-of-bounds write reachable from (int)/(float)/is_numeric() and the file, process and regex helpers


nahime0:
## Summary

`__rt_cstr` copies a PHP string's **full byte length** into `_cstr_buf`, a fixed 4096-byte `.comm` reservation, with no length check and then appends a NUL terminator. Any PHP string longer than 4095 bytes that reaches the helper therefore writes past the buffer into whatever the linker placed after it. The same applies to `__rt_cstr2` / `_cstr_buf2`.

This is not a theoretical path: it is reachable from `(int)$s`, `(float)$s`, `is_numeric($s)`, loose comparison, mixed-array arithmetic, and roughly 35 call sites across the file, process, and regex helpers. There is no diagnostic and, in small programs, no crash — the write lands inside other reserved runtime state.

`__rt_str_to_cstr` already exists as the unbounded, heap-backed alternative with the *same* input convention, so the fix is mostly a matter of routing.

## Root cause

- `src/codegen_support/runtime/data/fixed.rs:741-742` reserves `_cstr_buf` and `_cstr_buf2` at 4096 bytes each.
- `src/codegen_support/runtime/io/cstr.rs:29-48` (AArch64) and `:84-105` (x86_64) implement the copy: the length arrives in `x2` / `rdx`, is moved straight into the loop counter (`mov x11, x2` at `:36`, `mov rcx, rdx` at `:89`), and the loop runs to zero. Nothing compares the length against 4096, and the terminator is written after the last copied byte.

The helper's own doc comment documents the buffer as "4096 bytes" but the contract never states a maximum input length, and no caller enforces one.

## Minimal reproduction

```php
<?php
$s = str_repeat("1", 9000);   // any PHP string > 4095 bytes
$n = (int)$s;                 // copies 9001 bytes into a 4096-byte buffer
echo "survived\n";
```

## Steps to reproduce

1. Save the program above as `main.php`.
2. Compile and run it: `cargo run -- main.php && ./main`.
3. It prints `survived`. The out-of-bounds write happened silently; increase the length to widen the blast radius.

## Reachable from (measured)

Verified with a 200 000-byte numeric string, all reaching `__rt_cstr` and all completing without a diagnostic:

| program | path |
|---|---|
| `(int)$s` | `__rt_str_to_int` → `__rt_cstr` |
[...]

PR #1066, opened 2026-09-17
fix(strings): (int) over a numeric string caps like PHP instead of wrapping

Guikingone:
Found while investigating #953, which lists `(int)$s` as one of the paths into the unbounded
`__rt_cstr` buffer. The buffer is a separate problem; on the way there, the *values* turned out
to be wrong.

## PHP applies two different rules

A float **value** is reduced modulo 2^64. A numeric **string** is **capped**. elephc was routing
the string through the float rule.

```php
(int)1e19      // php: -8446744073709551616   — the float rule, and elephc agrees
(int)"1e19"    // php:  9223372036854775807   — the STRING rule
```

`__rt_php_float_to_int` implements the float rule correctly and is untouched; `__rt_str_to_int`
just should not have been calling it.

## What was wrong

| expression | PHP | before |
| --- | --- | --- |
| `(int)"1e19"` | `9223372036854775807` | `-8446744073709551616` |
| `(int)"-1e19"` | `-9223372036854775808` | `8446744073709551616` |
| `(int)"1.8e19"` | `9223372036854775807` | `-446744073709551616` |
| `(int)"1e100"` | `9223372036854775807` | `0` |
| `(int)"1e308"` | `9223372036854775807` | `0` |
| `(int)str_repeat("1", 310)` | `0` | `9223372036854775807` |
| `(int)("-" . str_repeat("1", 310))` | `0` | `-9223372036854775808` |

The wrapping cases are the ones that matter: `(int)$_GET['page']` on an oversized value produced
a large negative number instead of `PHP_INT_MAX`, with no warning.

The last two are the reverse failure, and they sit on the *integer-form* path: `strtoll`
saturates a 310-digit integer to `PHP_INT_MAX`, but PHP classifies a string whose value
overflows the double as `IS_DOUBLE`, and INF casts to 0. So the non-finite check has to run
**before** the integer/float form choice — that string is integer-form.

## The fix

- Non-finite parsed double → `0`, checked first, for either form.
- Float-form → cap. AArch64 needs no bounds of its own, because `fcvtzs` saturates by
  definition, which is exactly PHP's cap for every finite double. x86_64's `cvttsd2si` reports
  the "integer indefinite" pattern instead — already correct for a negative overflow, so only a
  positive one is turned into `PHP_INT_MAX`.
- Integer-form → unchanged, `strtoll`'s exact value.

That last point is load-bearing and has its own fixture: `2^53 + 1` is exactly representable as
an integer but not as a double, so capping the parsed double instead of taking `strtoll`'s
answer would round `(int)"9007199254740993"` down by one.

## Verification

- 23 shapes byte-identical to PHP 8.5.10, spanning the int64 boundary (`…807`, `…808`, `…809`),
  the finite/INF boundary (309 vs 310 digits), both signs, leading whitespace, and float-form
  spellings.
- The same values through `intval()`, `settype()`, mixed-array arithmetic and loose comparison.
- Both fixtures fail without the fix.
- x86_64 assembles clean for `x86_64-unknown-linux-gnu`; `--lib` 1680, `strings::` 392 and
  `operators::` 127 pass.


Guikingone:
The one red check here is not this change. `Curl Codegen Tests (macos-aarch64 4/4)` fails on `curl_monitor_excludes_write_callback_cpu_from_network_wait`, a timing assertion, and it has now failed twice in a row on this runner with the same shape:

```
run 1:  control 372625 ns → in-callback 7815417 ns
run 2:  control 432417 ns → in-callback 7418999 ns
```

against `WAIT_SLACK_NS = 5_000_000`. The defect that assertion guards produced **133.9 ms** against a 1.2 ms control — 18x further out than either sample here. This is loopback scheduling jitter on a shared runner, not billed callback CPU: the burn is ~130 ms and would show as that.

This PR touches `__rt_str_to_int` and adds `__rt_php_float_to_int_cap`; nothing on the curl monitoring path. `Build & Test` is the aggregation gate and fails only because that shard did.

#1024 is exactly this: it raises the bound to 20 ms — better than 2x over these samples, still ~6.6x below the gap the defect produces — and, because widening a tolerance can only move a test toward vacuous, it adds an assertion that the in-callback burn measured from the capture exceeds the slack by 2x, so a fully misbilled callback can never satisfy the comparison. Its comment already cites the 7.80 ms sample from run 1; run 2's 7.42 ms is a third data point.

Re-running does not help — it has now been re-run once and reproduced. #1024 should land first, then this one goes green on a rebase.

Guikingone:
Addressed in `37225c6c3a`. The runtime page listed `__rt_php_float_to_int` among the array/scalar entry points and said nothing about the capping sibling, so the reason a numeric string does not use the wrapping helper lived only in the code.

It now documents the rule table and why the split exists:

| source | rule | helper | `(int)` of `1e19` |
|---|---|---|---|
| a float VALUE | reduce modulo 2^64 | `__rt_php_float_to_int` | `-8446744073709551616` |
| a numeric STRING | CAP at the integer range | `__rt_php_float_to_int_cap` | `9223372036854775807` |

plus the reason both helpers check NaN and the infinities *before* converting: `fcvtzs` saturates by definition while `cvttsd2si` answers with the indefinite pattern, which is what made a bare conversion at a call site produce a per-target `(int)NAN`. All four values verified against host PHP 8.5.10.

nahime0:
## Follow-ups

| Issue | Topic |
|-------|--------|
| 1175 | Docs: wrapping float→int helper returns `x9`/`r11`, not `x0`/`rax` |
| 1176 | `str_to_int` / `numeric` preambles still say truncate/wrap-only |
| 1177 | Optional fold/`intval` coverage for `"1e19\"` and 310-digit overflow |

#1066: `__rt_str_to_int` no longer routes float-form numeric strings through the float-value wrap (`__rt_php_float_to_int`). Non-finite → `0`; integer-form → exact `strtoll`; float-form → saturate/cap to `PHP_INT_MAX`/`MIN` (PHP string rule). `__rt_php_float_to_int` (value rule) is unchanged.

PR #1067, opened 2026-09-17
fix(codegen): box a nullable-int union before it reaches a mixed parameter

### LINKED ISSUE #1046: Spreading an empty array drops integer parameter defaults
## Summary

Spreading an **empty array** into a call whose parameters have INTEGER defaults loses the defaults: the parameters arrive empty instead of taking their declared values.

## Reproduction

```php
<?php
function f($a = 0, $b = 0) { echo "$a/$b"; }
$e = [];
f(...$e);
```

PHP: `0/0`

elephc: `/`

## Perimeter

| Call | PHP | elephc |
| --- | --- | --- |
| `function f($a = 0, $b = 0){} $e = []; f(...$e);` | `0/0` | **`/`** |
| `class P { __construct($a = 0, $b = 0) } $e = []; new P(...$e);` | `0/0` | **`/`** |
| `$c = "P"; $e = []; new $c(...$e);` | `0/0` | **`/`** |
| `$g = "f"; $e = []; $g(...$e);` | `0/0` | **`/`** |
| `$c = "P"; new $c();` | `0/0` | `0/0` — correct |
| `$c = "P"; $e = [1]; new $c(...$e);` | `1/0` | `1/0` — correct |
| `class P2 { __construct($a = 'x', $b = 'y') } $e = []; new $c(...$e);` | `x/y` | `x/y` — correct |
| `$o->$m(...$e)` (dynamic method) | `0/0` | `0/0` — correct |

Two things narrow it sharply. **String** defaults survive the same call, and a NON-empty spread survives it too — so it is the empty spread combined with an integer default. An empty argument list written as `()` rather than `(...$e)` is also correct, so the defaults themselves are fine.

The output `/` rather than `0/0` says the parameters are neither the default nor a coerced zero: they interpolate as the empty string, which an uninitialized integer slot read through a string conversion would not produce either.

## Why it is filed separately

Found while fixing #685 (a string-keyed spread through a dynamic call target). It is unrelated: the failing calls include a plain `f(...$e)` with a statically known function name, which #685's change does not touch, and `--emit-ir` confirms the rewrite that fix adds never fires on these fixtures.

## Environment

Guikingone:
## The report

`$e = []; f(...$e)` lost every **integer** parameter default:

```php
function f($a = 0, $b = 0) { echo "$a/$b"; }
$e = [];
f(...$e);          // PHP: 0/0     elephc: /
```

Two things in the issue narrowed it sharply: a **string** default survived the
same call (`x/y`), and a **non-empty** spread survived it too. So the defaults
themselves were fine, and so was the spread rewrite.

## What it actually is

Neither the spread nor the defaults. An empty spread lowers each parameter to
`count($e) > n ? $e[n] : <default>`. Over an `array<never>` the read arm types
as `null` and the default arm as `int`, so the merged local is the union
`null|int` — and that union is the one case where a type's *stored shape* does
not follow from its variant:

```rust
// PhpType::codegen_repr()
PhpType::Union(members) if null_repr_is_tagged() && nullable_int_union_members(members)
    => PhpType::TaggedScalar,   // a payload register + a tag register
PhpType::Union(_) => PhpType::Mixed,   // a boxed cell
```

`emit_box_current_value_as_mixed` matched the **variant**:

```rust
PhpType::Mixed | PhpType::Union(_) => {}   // "already a cell" — emit nothing
```

so a nullable-int union took the no-op arm and the callee received the raw
payload word where a Mixed cell **address** was expected. `null|string` was
never affected because it genuinely is a boxed cell — which is exactly why the
issue's table shows `x/y` surviving the call that loses `0/0`.

This is not really a spread bug. The minimal reproduction has no spread in it:

```php
function p($v) { var_dump($v); }
$e = [];
p(count($e) > 0 ? $e[0] : 7);   // PHP: int(7)   elephc: (nothing)
```

## The fix

Ask `codegen_repr()` instead of matching the variant, and route a union that is
stored as a tagged scalar to the arm that already boxes a declared
`TaggedScalar` — same storage, same boxing. One central emitter serves ~150 call
sites; its previous behaviour for this shape was to emit nothing at all, which
is never correct.

## Verification

The issue's full perimeter table, diffed against host PHP 8.5.10 — identical,
including the rows that were already correct:

| Call | PHP | before | after |
| --- | --- | --- | --- |
| `f(...$e)` | `0/0` | `/` | `0/0` |
| `new P(...$e)` | `0/0` | `/` | `0/0` |
| `new $c(...$e)` | `0/0` | `/` | `0/0` |
| `$g(...$e)` | `0/0` | `/` | `0/0` |
| `$o->m(...$e)` | `0/0` | `0/0` | `0/0` |
| `new $c()` | `0/0` | `0/0` | `0/0` |
| `new $c(...[1])` | `1/0` | `1/0` | `1/0` |
| `new P2(...$e)` (string defaults) | `x/y` | `x/y` | `x/y` |

Tests, each verified to fail without the change:

- `null_sentinel/tagged.rs` — the mechanism (a tagged scalar reaching a `mixed`
  parameter), the reported shape, and all four call forms. Reverting the fix
  reproduces the issue's table exactly: `left: "f:/|P:/|P:/|f:/|M:/|P:1/0|"`.
- `runtime_gc/tagged_scalar_boxing.rs` — the fix makes these calls **allocate**
  where they previously did not, so three 300-iteration fixtures (int arm, null
  arm, two-parameter empty spread) pin that the existing argument-temporary
  cleanup releases the new cell. All end `leak summary: clean`.

Also measured: the perimeter fixture's 6 residual blocks are pre-existing — a
spread-free control program leaks the same 6.

`cargo test -p elephc --lib` 1680/1680. Full codegen suite running.

`docs/internals/memory-model.md` gains the invariant this bug fell through:
deciding "is this already a Mixed cell?" must go through `codegen_repr()`, never
a `PhpType::Union(_)` match.

Fixes #1046.


Guikingone:
I went looking for this and **could not reproduce it**. Reporting what I measured rather than changing code on a leak I cannot observe.

**Your premise about the code is right.** `reference_arguments.rs` does call `materialize_direct_call_arg_for_param` and does return `cleanup_slots: Vec::new()`, so if that call allocates a `Mixed` cell, nothing in that materializer releases it.

**The conclusion does not follow in practice.** Five shapes, all under `--heap-debug`, all passing a nullable-int union to an untyped non-reference parameter:

| shape | iterations | result |
| --- | --- | --- |
| `$o->take($v)` instance, no by-ref param | 300 | `allocs=3303 frees=3303`, clean |
| `Sink::stake($v)` static | 300 | (same run) clean |
| `$o->take(&$out, $v)` — signature HAS a by-ref param, union beside it | 300 | `allocs=3303 frees=3303`, clean |
| `Sink::stake(&$out, $v)` static, same | 300 | (same run) clean |
| `ignore(&$out, $v)` / `declared(&$out, mixed $v)` / `three(&$out, $a, $b, $c)` | 400 | `allocs=4003 frees=4003`, clean |

`peak_live_bytes` stays flat (216, then 256) across every run, which is what rules out a per-call leak rather than a small constant one. Values match host PHP 8.5.10 in all cases, including the `null` arm, so the union really is being passed — the boxing is firing, not being skipped.

The cell does get allocated; something releases it. The most likely explanation is that an untyped parameter receiving a freshly boxed `Mixed` takes ownership of it and the callee's own scope cleanup does the release — which is why the direct-call path needs planned cleanup (it can pass *borrowed* values) and this path does not.

That is an inference, not a measurement, so I would rather not claim it as settled. If you have a concrete shape that leaks — a different receiver kind, a variadic tail, a nullsafe call, something I did not think to try — post it and I will fix it properly. I would rather add cleanup for a real case than add it speculatively on a path where the accounting currently balances.

PR #1068, opened 2026-09-17
fix(ir_lower): type a nested array literal with context, not the syntactic guess

### LINKED ISSUE #984: Reading elements of a typed array held inside a `mixed` value returns a raw pointer (not the boxed element)
## Summary

When an array with a non-`array` element type (`Array(Str)`, `Array(Int)`, …) ends up boxed inside a `mixed` value — for example because it was stored under a string key, a property, or a superglobal entry like `$_SERVER['argv']` — reading one element by index (`$box[0]`) does not unbox the cell. It returns the raw 8-byte payload interpreted as an integer, and `foreach` over the array iterates zero times. `count()` and other ops on the outer array are unaffected.

Repro (any standalone compiled CLI program, current `main` — also reproduces on v0.26.6, so this is pre-existing, not a regression):

```php
<?php
$n = ["a" => 1];         // outer associative array (Mixed values)
$n["nested"] = $argv;     // box a typed `Array(Str)` into a Mixed cell

echo count($n["nested"]), "\n";   // 2  ✓
var_dump($n["nested"][0]);        // int(140726352842350)  ✗  PHP: string($argv[0])
foreach ($n["nested"] as $v) { }  // zero iterations         ✗
echo "done\n";
```

`$argv` directly, in any combination, is unaffected. The bug is in the indirection through the `mixed` value, not in the inner array.

## Same family as the limitations already filed from #913

- #924 — closure/dynamic-callable invocations leak one boxed Mixed cell per string/mixed argument
- #925 — invoker array-arg marshalling garbage (bare `array $attributes` receives `[7]`, loop-grown assoc boxed with the indexed tag)
- #923 — array read-copy leaks (the `isset($this->prop[$k]) ? … : []` shape)

This issue is the read-only counterpart of #923 / #925: instead of writing through a Mixed cell corrupting the slot, *reading* an element through it skips the unbox step and reads the raw payload as `int`.

## Why it matters

The CLI environment work in #802 made `$_SERVER` carry `argv` as a typed `Array(Str)`. The doc-update for 0.27.0 documents the limitation under `docs/php/system-and-io.md` and shows `foreach ($argv as $i => $arg)` as the workaround, but `$_SERVER['argv']` is itself an `Array(Str)` indexed by string key inside a `mixed` value, so:

- `count($_SERVER['argv'])` is correct,
- `$_SERVER['argv'][0]` is wrong,
- `foreach ($_SERVER['argv'] as $a)` iterates zero times.

The `xml_parse_into_struct()` workload from #913 works around the same shape by reading through `??` and copying instead of reading through the typed-array-in-mixed path; the cli/getenv path can't do the same without copying on every read.

## Expected

`$box[0]` on a `mixed` cell whose payload is `Array(Str)` should return `string`, `foreach` should iterate the typed array in order, and a `string`-keyed read on a Mixed-typed `Array` should return its element (typed scalar or `null`).

## Where it likely lives

- Lowering of `ArrayIndex` / `Index` expressions whose base is a `mixed` value with an array payload. The default path probably reads the cell tag, then takes the union arm, but `Mixed::Array(T)` is stored as a boxed cell whose tag is the union tag and whose payload is the array pointer; the index op probably follows the union arm and uses the pointer as the array, then reads the element as a raw 8-byte slot.
- The same path probably explains why `foreach` over a Mixed-typed array iterates nothing: the iterator is built over a payload the element reader cannot address.
- Suggest checking `src/codegen/lower_inst/builtins/arrays/` and the mixed-array get/set helpers under `src/codegen_support/runtime/arrays/` (compare how an `Array(Mixed)` value held under `Array(Mixed)` is unboxed vs. how it is not when held under a string key in a different outer shape).

## Repro

`examples/environment/main.php` (new in 0.27.0) exercises the failure mode for `$_SERVER['argv']` specifically; a smaller standalone case is the snippet at the top of this issue.

## Validation

- A focused codegen test under `tests/codegen/` that asserts `$box = ["x" => [10, 20]]; $box["x"][1];` answers `int(20)` and `foreach ($box["x"] as $v) { … }` runs `count($box["x"])` times.
- The same expectation for `Array(Str)`, `Array(Int)`, and `Array(Mixed)` payloads inside a string-keyed outer array.
- A regression that `$_SERVER['argv'][0]` and `foreach ($_SERVER['argv'])` answer like `$argv`.
- Mutating the typed array through the box (`$box["x"][] = …`) is a separate question and is *not* in scope here.

Guikingone:
## First: the reported repro no longer fails

The issue's own snippet, and its suggested validation case, both pass on current
`main`:

```php
$n = ["a" => 1]; $n["nested"] = $argv;
count($n["nested"]);        // 3   ✓
$n["nested"][0];            // the string  ✓   (issue reported int(1407…))
foreach ($n["nested"] …)    // 3 iterations  ✓
$_SERVER['argv'][0];        // the string  ✓
$box = ["x" => [10, 20]]; $box["x"][1];   // int(20)  ✓
```

Something between the filing and now fixed those. But the *class* of defect is
still live, one nesting level deeper — same symptom, an inner array's pointer
read back as an integer:

```php
$v = ["p", "q"];
$a = ["outer" => ["inner" => $v]];
var_dump($a["outer"]["inner"][1]);   // PHP: string(1) "q"   elephc: int(0)
```

No `$argv`, no superglobal, no `mixed` — four lines.

## What it is

An array literal's stamped element type is load-bearing, not decoration:
`HashGet` reads an element back in the shape the container's `php_type` claims.
The IR shows the stamp and the reality disagreeing:

```
v4: php=array<string, array<string, int>>   = hash_new   ← the OUTER literal's stamp
v6: php=array<string, array<string>>        = hash_new   ← what the inner hash really is
…
v12: php=array<string, int>   = hash_get v10 "outer"
v14: php=int|null             = hash_get v12 "inner"     ← reads an array pointer as an int
```

`assoc_array_literal_value_type_for_ir` has one arm per element shape, and had
none for a nested array literal — so it fell through to
`infer_expr_type_syntactic`, which is context-free. It cannot see a local's type
and answers `Int` for any variable, hence the fabricated `int`.

The inner literal's own call got it right (`v6`) one line earlier. Only the
outer literal's *view* of it was fabricated.

## Why it survived

Two accidents:

- The all-literal spelling `["j" => ["k" => ["p", "q"]]]` is the one case the
  syntactic guess gets right unaided, so the obvious test shape passes.
- An **indexed** outer literal was always correct:
  `array_literal_element_type_for_ir` already carries the two nested-literal
  arms its associative sibling was missing. `[["k" => $v]]` works;
  `["j" => ["k" => $v]]` does not.

## The fix

Add the two arms, recursing into the same context-aware functions that lower the
nested literal — restoring symmetry with the indexed sibling.

## Verification

Seven shapes measured against host PHP 8.5.10, all identical after the fix
(`a`, `d`, `e` were wrong before):

| Shape | before | after |
| --- | --- | --- |
| `["outer" => ["inner" => $argv]]`, chained read | `0` | `one` |
| all literals, 2 and 3 levels deep | `q` / `y` | unchanged ✓ |
| built in two statements | `one` | unchanged ✓ |
| read via an intermediate local | `0` | `one` |
| inner from a loop-grown `Array(Str)` | `0` | `one` |
| indexed outer `[["inner" => $argv]]` | `one` | unchanged ✓ |

Four tests in `tests/codegen/arrays/nested.rs`; three fail without the change
(`left: "int(0)\n"`), and the fourth pins the indexed-outer path that already
worked so a later unification cannot break it.

Suites: `arrays` 572, `types` 313, `objects` 254, `oop` 619, `-p elephc --lib`
1680 — all green, including `lowers_examples_corpus`. Heap debug on the repro is
`leak summary: clean`.

`docs/internals/the-ir.md` gains the invariant: the two stamp functions must stay
context-aware and in step, and `infer_expr_type_syntactic` is a last-resort
fallback whose `Int`-for-any-variable answer fabricates the stamp if an element
shape reaches it.

Fixes #984.


nahime0:
## Follow-ups

| Issue | Topic |
|-------|--------|
| 1178 | Indexed stamp still missing `ScopedConstantAccess` (assoc already has it) |
| 1179 | `?:` / `??` / `clone` / `match` wrappers still hit syntactic Int-for-variable |
| 1180 | `assoc_array_literal_type_from_spreads` still types spread inners syntactically |
| 1181 | Intermediate-local test comment says int-valued, fixture is strings |
| 1182 | If closing #984, pin original `$n[\"nested\"] = $argv` assignment shape separately |

#1068: `assoc_array_literal_value_type_for_ir` types nested array literals through the same context-aware helpers used for lowering, instead of `infer_expr_type_syntactic` (which answered `Int` for any `Variable`). That stopped the outer hash stamping `array<string, int>` over an inner hash that held `array<string>`, which made `HashGet` read the inner pointer as an integer.

Issue #576, opened 2026-07-19
Untyped functions reachable only via dynamic callables get Int-defaulted params: returned strings silently become int(0)


nahime0:
## Summary

An **untyped** function that is only reachable through dynamic callables (a runtime string name passed to `call_user_func`, `array_map`, an `ob_start` handler, …) gets its untyped parameters **defaulted to `Int` by the checker**, because no direct call site exists to infer them from. Any value the function returns that flows from such a parameter is then typed `Int`, and the runtime-callable invoker **silently coerces the real value through an int cast**: a returned string becomes `int(0)` instead of the string.

No diagnostic is emitted — the value is silently corrupted at runtime.

## Reproduction

```php
<?php
function h($b, $p) { return $b; }

$fn = 'h';
var_dump(call_user_func($fn, "probe", 9));
```

**elephc** (`main` @ 9377e567a, macos-aarch64, debug build):

```
int(0)
```

**PHP 8**:

```
string(5) "probe"
```

## The masking variant (why this is easy to miss)

Adding a single direct call anywhere in the program teaches the checker the real parameter types, and the dynamic path is then correct too:

```php
<?php
function h($b, $p) { return $b; }

var_dump(h("direct", 9));                 // string(6) "direct"
$fn = 'h';
var_dump(call_user_func($fn, "probe", 9)); // string(5) "probe"  ← now correct!
```

This makes the bug look nondeterministic across "similar" programs: any A/B comparison must use byte-identical sources.

## Scope observations

| Variant | Result |
|---|---|
| `return $b;` (string arg) | `int(0)` — corrupted |
| `return $p;` (int arg, `("x", 12345)`) | `int(12345)` — coincidentally correct (int survives the int cast) |
| Typed params `function ht(string $b, int $p): string` | correct |
| Closure with untyped params | correct |
| `call_user_func('h', ...)` (literal name, statically resolved) | correct |
| `$fn("probe", 9)` (dynamic variable call) | correct (interestingly) |
| Statically-typed consumers | `strlen($r)` fails to compile with `EIR backend error: unsupported EIR backend feature: strlen for PHP type Int`, confirming the checker types the call result as `Int` |

The same mechanism affects `ob_start('h')` output handlers on the `feat/output-buffering-builtins` branch (handler receives/returns through the same runtime-callable invoker), which is how it was found — but the bug is independent of that branch and reproduces on `main`.

## Analysis

[...]

PR #1083, opened 2026-09-18
fix(ir_lower): a closure returning a user call keeps the callee's return type

### LINKED ISSUE #1028: A closure that returns a user function call is typed int: returned strings silently become 0
## Summary

A closure whose body is `return <user function call>;` is typed as returning `int`, whatever the
callee actually returns — **including when the callee declares `: string`**. The string is read
back through an int slot and the program silently prints `0`.

No diagnostic. No null, no dynamic callable, no `call_user_func` — a direct call to a directly
declared closure.

## Repro

```php
<?php
function tag(int $n): string { return "t" . $n; }

$a = function ($v) { return tag($v); };
echo $a(1), "|";
```

| | |
|---|---|
| PHP 8.5.10 | `t1\|` |
| elephc `6e6f02611f` | `0\|` |

The checker's view is confirmable: `strlen($a(1))` refuses to lower with
*"strlen cannot lower checked operand type Int"*, so the call result really is typed `Int`.

## What does and does not reproduce

Same `tag()`, six closures:

| closure body | PHP | elephc |
|---|---|---|
| `function ($v) { return tag($v); }` | `t1` | **`0`** |
| `fn ($v) => tag($v)` | `t2` | **`0`** |
| `function ($v) { return untyped($v); }` (callee has no return hint) | `u3` | **`0`** |
| `function () { return tag(3); }` (no parameters at all) | `t3` | **`0`** |
| `function (int $v) { return tag($v); }` (typed parameter) | `t1` | **`0`** |
| `function ($v): string { return tag($v); }` (declared return) | `t2` | `t2` |
| `function ($v) { $s = tag($v); return $s; }` (through a local) | `t4` | `t4` |
| `function ($v) { return "d" . $v; }` (no call) | `d4` | `d4` |
| `function ($v) { return strlen(tag($v)); }` (int result) | `2` | `2` |

So it is not about parameters, arity, or the callee's hint. It is the **closure's own return
type**, inferred from a body whose `return` is a call.

## Analysis

`infer_closure_call_type` reaches `ExprKind::Closure { body, .. }` and takes the closure's
return type from `infer_return_type_syntactic(body)`
(`src/types/checker/inference/ops.rs:835`).

That function is **syntactic**: it has no symbol table, so it cannot look `tag` up. It routes
each `return` expression through `infer_expr_type_syntactic`, whose `FunctionCall` arm is a
hardcoded list of *builtins* ending in

```rust
_ => PhpType::Int,     // src/types/checker/inference/syntactic.rs:359
```

A user-defined callee matches nothing in that list and lands on the `Int` fallback — the same
Int-default mechanism as #576, reached through a completely different path (a direct closure
call rather than a dynamic callable).

The three shapes that work say the same thing from the other side: a declared `: string`
skips the inference, a local gives the checker a semantically typed binding to read, and a
non-call body never reaches the fallback.

## Why the fallback is load-bearing

It cannot simply become `Mixed`: the comments around it record that arms were added *because*
`Int` was wrong there (`dechex($n)` inside an array literal produced `[0]`), which means other
callers depend on the fallback's current shape. The fix likely belongs at the closure site —
resolving the callee semantically when the schema is available, rather than widening a
syntactic helper that runs before schemas exist.

## Acceptance criteria

- [ ] `$f = function ($v) { return tag($v); }; echo $f(1);` prints the callee's real return value
- [ ] Holds for `fn () =>` arrow bodies, zero-parameter closures and typed parameters
- [ ] Holds whether or not the callee declares a return type
- [ ] A callee returning `array`, `object` and `float` is covered as well as `string`
- [ ] `strlen($f(1))` compiles and agrees with PHP, so the checker's type is right and not only the runtime bytes
- [ ] The shapes that already work stay working: declared `: string`, via a local, non-call bodies, int results
- [ ] Codegen regression covering the silent-`0` shape
- [ ] Both architectures

## Related

- #576 — the same `Int` default reached through dynamic callables with no direct call site.

## Found via

Writing the regression for #567: the fixture's arrow-function row returned `0` for reasons
that had nothing to do with #567's null argument.

Guikingone:
```php
function tag(int $n): string { return "t" . $n; }
$f = function ($v) { return tag($v); };
echo $f(1);          // PHP: t1    elephc: 0
```

Silent — no diagnostic, a declared `: string` read back through an int slot.

## It is not where the issue said it was

The issue placed this in the **checker**, at `infer_return_type_syntactic`. That
turned out to be wrong, and worth stating plainly because it changes what the
fix has to be.

Probing `resolve_closure_return_type` on the issue's own fixture file:

```
PROBE: closure at line 7  inferred Str      ← $a = function ($v) { return tag($v); }
PROBE: closure at line 10 inferred Str      ← fn ($v) => tag($v)
PROBE: closure at line 13 inferred Str      ← untyped callee
PROBE: closure at line 16 inferred Str      ← zero-parameter
PROBE: closure at line 19 inferred Str      ← typed parameter
```

The checker gets every one of them right: `collect_return_infos` routes each
`return` through the full semantic `infer_type`, which resolves `tag` from the
schema. `infer_return_type_syntactic` is reachable for a directly-invoked
closure literal, but it is not what types the reported shape.

## Where it actually is

The EIR re-derives the closure's return type for itself, in
`direct_closure_return_expr_type`:

```rust
if matches!(expr.kind, ExprKind::FunctionCall { .. }) {
    if let Some(ty) = builtin_call_types.get(&expr.span) {   // BUILTIN calls only
        return ty.clone();
    }
}
…
crate::types::checker::infer_expr_type_syntactic(expr)       // → Int for a user callee
```

`builtin_call_types` records builtin call results. A user callee misses, falls
past every remaining arm, and lands on the syntactic fallback — whose
`FunctionCall` arm is a hardcoded builtin list ending in `_ => PhpType::Int`.

The three shapes that already worked confirm it from the other side: a declared
`: string` never reaches the re-derivation, a local gives the lowering a typed
binding instead of a call, and a non-call body never reaches the fallback.

## The fix

Resolve a user callee from the module's function signatures, which the lowering
already holds. `parent.functions` is threaded down the six-function chain that
types a directly-returned expression, so **array-literal elements** built out of
user calls are covered by the same lookup — that chain is shared.

Widening the syntactic fallback was never an option, and the issue said so: the
comments around it record arms added *because* `Int` was wrong there
(`dechex($n)` in an array literal produced `[0]`), so other callers depend on its
current shape.

## Verification

Every row of the issue's table, plus the return types it asked for, diffed
against host PHP 8.5.10:

| closure body | before | after |
| --- | --- | --- |
| `function ($v) { return tag($v); }` | `0` | `t1` |
| `fn ($v) => tag($v)` | `0` | `t2` |
| `function ($v) { return untyped($v); }` (no hint on callee) | `0` | `u3` |
| `function () { return tag(3); }` | `0` | `t3` |
| `function (int $v) { return tag($v); }` | `0` | `t1` |
| `function ($v) { return flt($v); }` (`: float`) | `0` | `2.5` |
| `function ($v) { return arr($v); }` (`: array`) | `0` | `26` |
| `function ($v) { return make($v); }` (`: Box`) | refused | `b7` |
| `function ($v): string { … }` | `t2` | unchanged ✓ |
| `function ($v) { $s = tag($v); return $s; }` | `t4` | unchanged ✓ |
| `function ($v) { return "d" . $v; }` | `d4` | unchanged ✓ |
| `function ($v) { return strlen(tag($v)); }` | `2` | unchanged ✓ |

**The checker-type criterion**: `strlen($f(1))` compiles and answers `2`. Before
the fix it did not merely print the wrong value — it refused to lower at all
with *"strlen cannot lower checked operand type Int"*, which is the clearest
possible statement that the result was typed `int`.

8 tests in `callables/closure_call_returns.rs`; 7 fail without the change
(`left: "0|0|0"`), and the eighth is the pin that the previously-working shapes
stay working — it passes either way by design.

Suites: `callables` 465, `types` 320, `optimizer` 366, `-p elephc --lib` 1684 —
all green.

Both architectures: the change is in target-independent EIR typing, so there is
no per-arch code path.

Fixes #1028.


Guikingone:
Correct, and it was the more serious half of this PR. Fixed in `8e3944f154`.

Reproduced your exact shape:

```php
function values($v) { return [$v]; }
$f = function ($v) { return values($v); };
var_dump($f("hello")[0]);
```

```
before: int(4363925416)      ← the element's own pointer, read as an integer
after:  string(5) "hello"
PHP:    string(5) "hello"
```

You named the cause precisely. Resolving the callee from `functions` fixed the scalar case but copied the signature's return type RAW, where ordinary call lowering goes through `eir_user_function_return_type` — which widens a container return to Mixed elements when the callee has an untyped by-value parameter, because such a parameter arrives as a boxed Mixed and anything built from it has Mixed elements whatever the signature's inferred element type says. The closure's contract said `array<string>` while the call produced `array<mixed>`.

**Worth stating plainly: this was strictly worse than what it replaced.** On `main` the same program does not compile at all ("count for PHP type Int"), so the first version of this change traded a hard refusal for a silent wrong answer. That is the wrong direction to move, and it is exactly what your review caught.

Now routed through the same normalization, so lowering and ordinary call lowering agree. Two regression tests added, `test_closure_call_returning_a_container_normalizes_its_element_type` (indexed and associative, read by index and by `foreach`) and `test_closure_call_container_keeps_a_float_elements_runtime_tag` — `var_dump` is the discriminating consumer there, since `echo` of a bad float element can still look plausible.

Verified across indexed, associative, hinted and unhinted containers, byte-identical to host PHP 8.5.10. `callables` 467, `types` 320, `arrays` 572, `-p elephc --lib` 1684.

**One thing this does not fix**, recorded so it is not mistaken for done: `strlen()` over such an element is still refused at compile time, because the CHECKER's closure return inference has the same raw-signature shape this commit fixes in the lowering. That is a false rejection rather than a miscompile, and the values themselves now read back correctly — but the two sides still disagree, and closing that is a separate change.

Guikingone:
I hit #1028 independently and opened #1110 before noticing this PR existed; I have closed mine as the duplicate. This one came first and already routes through `eir_user_function_return_type` rather than the raw signature type, which is the load-bearing part — three reviewers proved on my branch that taking `sig.return_type` raw miscompiles a container returned by an untyped-parameter callee:

```php
function f($x) { return [$x]; }
$g = function (int $v) { return f($v); };
echo $g(5)[0];        // PHP: 5, raw signature type: a cell pointer
```

So nothing needs to move over. Three measurements from my branch that may be worth adding as coverage here, all reproduced against host PHP 8.5.10 and none of them in #1028's table:

```php
function half(int $n): float { return $n / 2; }
$a = function ($v) { return half($v); };
var_dump($a(3));                       // was int(1),  PHP float(1.5)

function mk(int $n): array { return [$n, $n + 1]; }
$a = function ($v) { return mk($v); };
count($a(1));                          // was REFUSED: count for PHP type Int

$a = function ($v) { return tag($v); };
strlen($a(1));                         // was a compiler PANIC, not a refusal
```

Also worth recording on the issue: its analysis points at `infer_closure_call_type` taking `infer_return_type_syntactic`, and that path is not the one this program takes — instrumenting `resolve_closure_return_type` prints `Str`. The checker is right; the wrong `int` is lowering's own independent inference of the same type.

nahime0:
Follow-ups from this PR (left open on purpose; tip already covers free user-function calls via `eir_user_function_return_type`):

- #1269 — closure return of method / static method still typed as `Int`
- #1270 — checker still rejects `strlen` on an element of a closure-returned container (lowering layout fixed; checker half still disagrees)
- #1271 — document direct closure return typing under `docs/internals/` (Greptile P2 / AGENTS.md)
- #1272 — closure return of extern call still typed as `Int`

Skipped here: dynamic callables (covered elsewhere, e.g. #576); #1028 is this PR.

Issue #1094, opened 2026-09-18
A property write through a `mixed` (or object-union) receiver is silently dropped for every class but stdClass


Guikingone:
## Summary

A property write whose **receiver is statically `mixed`** (or an object union such as `T|int`) is silently discarded. No warning, no error, no crash — the store simply never happens, and a read of the same property immediately afterwards returns the old value.

The runtime helper that serves this path, `__rt_mixed_property_set`, only handles a `stdClass` payload. Every other class silently drops the write, by design — the emitter says so in its own doc comment:

```rust
/// Unboxes the Mixed, validates the tag is 6 (object) and the class_id matches
/// `_stdclass_class_id`, then delegates to `__rt_stdclass_set`. Non-stdClass payloads
/// silently drop the write (no return value).
```
— `src/codegen_support/runtime/objects/stdclass.rs:373-378`

The rationale in the companion comment is that this "mirrors PHP's *attempt to assign property on non-object* warning behaviour". That holds for a non-object payload. It does not hold for a declared-class object, which is what the payload actually is in every repro below: PHP writes the property.

## Minimal repro

```php
<?php
class T { public int $n = 1; }
function w(mixed $o): void { $o->n = 9; var_dump($o->n); }
$t = new T();
w($t);
var_dump($t->n);
```

```
$ php a.php        $ elephc -q a.php && ./a
int(9)             int(1)
int(9)             int(1)
```

The **callee's own** read-back is already wrong, so the write is not landing in a copy — it is not landing anywhere.

## What is and is not affected

| receiver's static type | result |
|---|---|
| `mixed` | **write dropped** |
| `T\|int` (object union) | **write dropped** |
| `T` | correct |
| `?T` | correct |
| `mixed`, but the write goes through a method (`$o->set(9)` → `$this->n = $v`) | correct |

The property's own declaration does not matter: a typed `int` property, a typed `string` property, an untyped `public $n`, and an undeclared (dynamic) property all drop the write identically.

```php
<?php
class T { public int $n = 1; }
function w(T|int $o): void { $o->n = 9; }   // union receiver: same loss
$t = new T();
w($t);
var_dump($t->n);                             // elephc: int(1)  —  PHP: int(9)
```

Dynamic properties lose the value too, where PHP deprecates but still writes:

```php
<?php
class T { public int $n = 1; }
$o = null;
for ($i = 0; $i < 1; $i++) { $o = new T(); $o->zz = 9; }
[...]

PR #1095, opened 2026-09-18
fix(ir_lower): release the box a narrowing or retaining store leaves behind

### LINKED ISSUE #1041: Compound assignment to a static property or a property's array element leaks one allocation per execution
## Summary

A **read-modify-write** on a static property or on an array element of an object property leaks one allocation per execution. Plain assignment to either is clean, and the read alone is clean, so the leak is specific to the compound path.

## Reproduction

```php
<?php
class T { public static int $s = 0; }
for ($i = 0; $i < 50; $i++) { T::$s += 1; }
echo T::$s;
```

```
$ elephc -q --gc-stats a.php && ./a
50
GC: allocs=50 frees=0
```

Fifty iterations, fifty allocations, **zero** frees.

```php
<?php
class T { public array $items = [1, 2]; }
for ($i = 0; $i < 50; $i++) { $o = new T(); $o->items[0] += 1; }
echo $o->items[0];
```

```
GC: allocs=401 frees=351
```

Again exactly one per iteration.

| fixture (50 iterations) | allocs / frees |
|---|---|
| `T::$s += 1;` | `50 / 0` — **leaks** |
| `T::$s++;` | `50 / 0` — **leaks** |
| `++T::$s;` | `50 / 0` — **leaks** |
| `T::$s = $i;` | `0 / 0` — clean |
| `$x = T::$s;` | `0 / 0` — clean |
| `$o->items[0] += 1;` | `401 / 351` — **leaks** |
| `$o->items[0]++;` | `401 / 351` — **leaks** |
| `$o->items[0] = 9;` | `301 / 301` — clean |
| `$x = $o->items[0];` | `250 / 250` — clean |
| `$a = ($o->items[0] += 1);` | `400 / 400` — clean |

Two things stand out. The increment spelling is irrelevant — `+= 1` leaks exactly as `++` does — so this is the compound read-modify-write path, not the increment operators. And the EXPRESSION form of the same compound assignment (`$a = ($o->items[0] += 1)`) is clean, which points at the statement lowering rather than at the shared value computation.

The counts scale exactly with the loop, so it is one per execution rather than a fixed process-lifetime cost.

## Why it is filed separately

Found while fixing #682 (`++`/`--` in expression position). Every leaking fixture above is an ordinary compound assignment with no increment in it, and every one of them behaves identically before that change, so this is an independent pre-existing bug on a path #682 does not touch. The regression test there documents the exclusion rather than pinning this behaviour.

## Environment

Guikingone:
Fixes #1041.

## The defect

Both halves of the issue are the same mistake in two places: **lowering picked the "moving store" ownership discipline for a store the backend actually performs as an independent one.** A moving store consumes its value operand, so lowering emits nothing after it. An independent-value store leaves the producer's reference outstanding, so lowering must release it — and neither site did.

### Half one: `T::$s += 1` on a typed static property

`ichecked_add` returns a **boxed** Mixed, because the addition may overflow to float. The store narrows it back out with `__rt_mixed_cast_int`:

```
v2: Heap(Mixed) php=mixed own=owned = ichecked_add v0 v1 ; effects: reads_heap, alloc_heap
store_static_property v2 data[4]          ← nothing releases v2
```

```asm
bl __rt_int_add_checked
stur x0, [x29, #-24]
; op=store_static_property
ldur x0, [x29, #-24]
bl __rt_mixed_cast_int      ; reads a payload word out of the box
str x0, [_static_prop_T_s]  ; the slot now holds an int and owns nothing of the box
```

`static_property_store_retains_independent_value` only recognised two independent-value shapes — boxing a concrete value into a Mixed slot, and unboxing a Mixed into an object slot. A Mixed narrowed into an `Int`, `Float`, `Bool`, `Str`, or tagged-scalar slot fell through to the moving-store branch, whose assumption (the store consumes the value) is false for every one of them. One 40-byte cell per assignment.

The same predicate also `Acquire`d a *borrowed* Mixed before such a store, which leaks in the other direction — that retain has no matching release once the slot only keeps a payload word:

```php
function bump(mixed $m): void { T::$s = $m; }   // int slot: 50 allocs, 0 frees
```

The fix states the rule as what it actually is — a store is an independent-value store when it **crosses the boxed/unboxed boundary in either direction** — and enumerates the narrowing slots codegen really casts out of. Deliberately not "any slot that is not Mixed": an array or iterable slot falls through the store lowering's catch-all arm untouched, so the Mixed pointer itself is what lands in the slot and the store does still consume it.

### Half two: `$o->items[0] += 1` on a property array

`release_property_array_insert_value_after_retain` bailed out whenever the element type was `Mixed`. That is right for **assoc** storage — `__rt_hash_set`'s value materializer only increfs a value that cannot hand over its own reference (`value_can_own_mixed_box_source`), so an owned temporary is moved in. It is wrong for **indexed** storage, where `lower_mixed_array_set_*` increfs unconditionally when the value is already a Mixed cell:

```rust
} else {
    ctx.load_value_to_result(value)?;
    abi::emit_incref_if_refcounted(ctx.emitter, &value_ty);   // ← the slot takes its own reference
}
```

The local-array path has always emitted the matching `release` (`$a[0] += 1` on an `array<mixed>` local is clean); the property path skipped it. The bail-out now applies only where the store really moves: assoc storage, and a concrete value that codegen boxes first (that boxing path releases the box it made, so a second release there would double-free).

`__rt_array_push_refcounted` retains the same way, so `$o->items[] = f()` leaked identically and is fixed by the same change.

## Measurements

Every row of the issue's table, at its stated iteration count, before → after:

| fixture | before | after |
|---|---|---|
| `T::$s += 1;` | `50 / 0` | `50 / 50` |
| `T::$s++;` | `50 / 0` | `50 / 50` |
| `++T::$s;` | `50 / 0` | `50 / 50` |
| `T::$s = $i;` | `0 / 0` | `0 / 0` |
| `$x = T::$s;` | `0 / 0` | `0 / 0` |
| `$o->items[0] += 1;` | `401 / 351` | `401 / 401` |
| `$o->items[0]++;` | `401 / 351` | `401 / 401` |
| `$o->items[0] = 9;` | `301 / 301` | `301 / 301` |
| `$x = $o->items[0];` | `250 / 250` | `250 / 250` |
| `$a = ($o->items[0] += 1);` | `400 / 400` | `400 / 400` |

Under `--heap-debug`, live blocks at exit, across the wider battery this was measured on (20 iterations each unless noted):

| shape | before | after |
|---|---|---|
| `int $s += $i` | 20 blocks / 800 B | clean |
| `$s++` and `++$s` | 40 / 1600 | clean |
| `?int $s += $i` | 20 / 800 | clean |
| `bool $b = <mixed>` | 20 / 800 | clean |
| `int $s = <borrowed mixed>` | 20 / 800 | clean |
| `$last = (T::$s += 2)` | 20 / 800 | clean |
| `T::$s += 5; $kept = T::$s; T::$s += 5;` | 2 / 80 | clean |
| `$o->items[0] += 1` ×2 elements | 40 / 2784 | clean |
| `$o->items[0]++` | 20 / 1344 | clean |
| `$o->items[0] = $o->items[0] + 1` | 20 / 1344 | clean |
| `$o->items[] = $o->items[0] + 1` | 20 / 944 | clean |
| `$o->items[0] = f()` (`f(): mixed`) | 20 / 1344 | clean |
| `$snapshot = $o->items;` then 20 element writes | 20 / 1376 | clean |
| `float $f += …`, `string $s .= …` (controls) | clean | clean |
| `$o->m['k'] += 1` (assoc control) | clean | clean |
| `$o->items[0] = [$i, $i+1]` / `= new C()` (boxing controls) | clean | clean |

No fixture changed its stdout, and every one still matches host PHP 8.5.10.

## Also in this PR

`emit_mixed_result_as_int` and `value_is_owned_mixed_store_temporary` in `src/codegen/lower_inst/static_properties.rs` are removed. They held a `__rt_decref_mixed` for exactly this case, gated on the producer being a checked-arithmetic temporary — but the block was **unreachable**: the branch above it returns for every `value_ty.codegen_repr()` in `{Mixed, Union}`, and this block's own guard requires `Mixed`, a strict subset. That dead release is what made the leak look handled. It is covered by `test_issue_1041_static_int_property_compound_add_heap_clean`, which is exactly the shape it claimed to serve.

## Tests

`tests/codegen/runtime_gc/compound_assign_stores.rs`, 20 tests. 15 fail on `main`, and the 5 that pass there are the controls (float slot, string slot, assoc element, boxed array element, boxed object element) — which is the point of having them: they pin the disciplines that must **not** gain a release.

Every fixture runs under `--heap-debug` and asserts `leak summary: clean`, so a release emitted twice fails here as a double free rather than passing quietly. The aliasing/COW and read-back fixtures cover the ownership side: a snapshot taken before the loop must not observe the element writes, and a value read out between two compound writes must survive the release of the box it came from.

Suites run locally, rebased on `main` @ `bf5437b687`: `runtime_gc` 321/321, `codegen::arrays` 578/578, `codegen::objects` 254/254, every test matching `static` 393/393.

## Adjacent findings, measured but not fixed here

Each was checked against `main` and this branch, and scaled across iteration counts so a fixed process-lifetime residue is not mistaken for a leak.

1. **A `string` static property assigned from a Mixed source still leaks one persisted string per assignment.** This PR takes it from 3 blocks per iteration to 1: over 10/20/40 iterations, `main` leaves 31/61/121 blocks and this branch leaves 11/21/41. The residue is `__rt_str_persist`'s copy, which the slot never releases when overwritten. The same loop with a source declared `string` instead of `mixed` leaves a flat 1 block at any count, so it is specific to the Mixed→`Str` narrowing arm.

   ```php
   class T { public static string $s = ''; }
   function make(int $i): mixed { return 'v' . $i; }
   for ($i = 0; $i < 40; $i++) { T::$s = make($i); }
   ```

2. **Integer overflow on a typed static property wraps instead of raising.** `T::$s += PHP_INT_MAX; T::$s += 1;` gives `int(-9223372036854775808)`; PHP raises `TypeError: Cannot assign float to property T::$s of type int`. Unchanged by this PR.

3. **A property write through a `mixed` receiver is silently dropped** — filed as #1094. `function w(mixed $o): void { $o->n = 9; }` loses the write for every class but `stdClass`, which is also why `$o = null;` before a loop makes `$o->items[0] = 9` inside it a no-op.

Two things I first read as leaks and then disproved, recorded so nobody re-chases them: a **static property's array element** (`T::$items[0] = 7;`) and a **static `mixed` property holding a container** both leave a *constant* live-block count — 4 blocks at 20 and at 40 writes, 2 blocks at 10 and at 40 — which is the static slot's own live value at process exit, not a per-write leak.


nahime0:
What changed in this PR:
- Aligns EIR ownership with stores that leave the producer's reference outstanding (narrowing / retaining), so the box is released after the store.
- Static half: typed compound-assign (`T::$s += 1`) when `ichecked_add` returns boxed Mixed and the store narrows it.
- Array half: indexed Mixed element writes (`$o->items[0] += 1` / push) where incref + retaining store left the producer live.
- Heap-debug tests fail closed on a second release; controls pin assoc / boxing / float / string-concat paths.

Follow-ups:
- #1200 — Mixed→string static persist leak (`__rt_str_persist` copy never released on overwrite)
- #1201 — typed static int overflow wraps instead of PHP `TypeError`
- #1202 — instance `property_store_keeps_independent_ref` still omits Mixed→`TaggedScalar` (`?int`)

Already open: [#1094](https://github.com/illegalstudio/elephc/issues/1094) (property write through mixed / object-union receiver dropped).

PR #1097, opened 2026-09-18
fix(ir_lower): read a builtin's return type when it is an array-literal element

### LINKED ISSUE #1096: An array literal whose element is an array-returning builtin call stores (int)$array — every such element becomes int(1)
## Summary

An array literal whose element is a **call to a builtin returning an array** stamps the literal with the wrong element type, and lowering then inserts an `(int)` cast to match the stamp. `(int)` of a non-empty array is `1`, so every such element silently becomes `int(1)`.

## Reproduction

```php
<?php
$a = [1, 2, 3, 4, 5];
$c = [array_slice($a, 0, 2), array_slice($a, 2)];
var_dump($c);
```

```
PHP                              elephc
array(2) {                       array(2) {
  [0]=> array(2) {                 [0]=> int(1)
    [0]=> int(1)                   [1]=> int(1)
    [1]=> int(2)                 }
  }
  [1]=> array(3) { … }
}
```

No warning, no error.

## The `1` is `(int)$array`, not a first element

`array_reverse([1,2,3,4,5])` starts with `5`, and it still renders `int(1)`. `explode(",", "a,b")` starts with `"a"`, and it renders `int(1)` too. The value is PHP's `(int)` of a non-empty array, which the IR shows directly:

```
v5:  Heap(Array) php=array<int> own=owned = array_new capacity=1     ← literal stamped array<int>
v9:  Heap(Array) php=array<int> own=owned = runtime_call … runtime.array_slice
v10: I64 php=int = cast v9 I64                                        ← the returned array, cast to int
release v9
array_push v5 v10
```

## Affected builtins

Every array-returning builtin I tried, in an indexed literal:

| element expression | elephc | PHP |
| --- | --- | --- |
| `[array_slice($a, 0, 2), array_slice($a, 2)]` | `[int(1), int(1)]` | two arrays |
| `[array_values($a)]` | `int(1)` | the 5-element array |
| `[array_keys($a)]` | `int(1)` | `[0,1,2,3,4]` |
| `[array_merge($a, [6])]` | `int(1)` | the 6-element array |
| `[array_reverse($a)]` | `int(1)` | `[5,4,3,2,1]` |
| `[explode(',', 'a,b')]` | `int(1)` | `["a","b"]` |
| `['k' => array_slice($a, 0, 2)]` | *refused:* `hash_set value PHP type Array(Int)` | the array |

A `foreach` over such a literal binds `int(1)` as well, so the loop body sees scalars where PHP sees arrays.

## What is not affected

| shape | result |
| --- | --- |
| a **user function** returning `array`: `[mk(), mk()]` | correct |
| the call bound to a local first: `$s = array_slice(…); [$s, $s]` | correct |
| the call mixed with a plain literal: `[array_slice(…), [9, 9]]` | correct |
| scalar-returning builtins: `[count($a), 7]`, `[strtoupper('a'), 'b']`, `[sqrt(4.0), 1.5]`, `[in_array(1, $a), false]` | correct |

The user-function case is the tell: `array_literal_element_type_for_ir` looks the callee up in `ctx.functions` and `ctx.extern_functions`, and a **builtin is in neither**, so it falls through to `infer_expr_type_syntactic`, which answers `Int` for any call it cannot name (`src/ir_lower/expr/indexed_array_literals.rs:427-436`, and the assoc twin at `assoc_array_literals.rs:112-121`). Adding one non-call element widens the merge and hides it, which is why this is easy to miss.

The checker has already resolved these calls — the return type is in `ctx.builtin_call_types`, keyed by span, which several other lowering sites already read.

## Downstream symptoms this explains

- `implode(",", $c[0])` on such a literal is refused with `implode array PHP type TaggedScalar` or `… PHP type Int` — the operand is a scalar as far as the backend knows.
- `count($c[0])` is refused with `count for PHP type TaggedScalar`.

## Environment

- elephc `0.27.0`, `main` @ `bf5437b687`
- macOS aarch64, host PHP 8.5.10

Guikingone:
Fixes #1096.

```php
$a = [1, 2, 3, 4, 5];
$c = [array_slice($a, 0, 2), array_slice($a, 2)];
var_dump($c);     // PHP: two arrays     elephc: [int(1), int(1)]
```

No warning, no error.

## What it is

The element type stamped on `ArrayNew`/`HashNew` is load-bearing: an element whose own type does not match the stamp is *converted* on the way in, and reads come back in the shape the stamp claims. `array_literal_element_type_for_ir` resolved a call element by looking the callee up in `ctx.functions` and `ctx.extern_functions` — and a **builtin is in neither**, so it fell through to `infer_expr_type_syntactic`, which answers `Int` for any call it cannot name.

The literal was therefore stamped `array<int>`, and lowering inserted the conversion that stamp demands:

```
v5:  Heap(Array) php=array<int> own=owned = array_new capacity=1     ← the wrong stamp
v9:  Heap(Array) php=array<int> own=owned = runtime_call … runtime.array_slice
v10: I64 php=int = cast v9 I64                                        ← the returned array, cast to int
release v9
array_push v5 v10
```

`(int)` of a non-empty array is `1`, which is where every `int(1)` came from. It is not a mis-read first element: `array_reverse([1,2,3,4,5])` starts with `5` and `explode(",", "a,b")` starts with `"a"`, and both rendered `int(1)`.

The checker has already resolved these calls and records the result in `ctx.builtin_call_types`, keyed by span — a map several other lowering sites already read. Both literal kinds now consult it before the syntactic fallback.

## Before → after

| element expression | before | after |
| --- | --- | --- |
| `[array_slice($a, 0, 2), array_slice($a, 2)]` | `[int(1), int(1)]` | two arrays |
| `[array_values($a)]` | `int(1)` | `1,2,3` |
| `[array_keys($a)]` | `int(1)` | `0,1,2` |
| `[array_merge($a, [3])]` | `int(1)` | `1,2,3` |
| `[array_reverse($a)]` | `int(1)` | `3,2,1` |
| `[explode(',', 'a,b')]` | `int(1)` | `a\|b` |
| `['k' => array_slice($a, 0, 2)]` | *refused:* `hash_set value PHP type Array(Int)` | the array |
| `foreach ([array_slice(…), array_slice(…)] as $row)` | binds `int(1)` | binds the arrays |
| `count($c[0])` on such a literal | *refused:* `count for PHP type TaggedScalar` | the count |

The associative twin (`assoc_array_literals.rs`) had the identical hole; there it surfaced as a compile refusal rather than a wrong value, which is why it was never reported.

Every fixture matches host PHP 8.5.10.

## Controls

Two shapes hid this, and both are pinned as tests that must keep passing:

- a **user function** returning `array` resolves through `ctx.functions` and was always correct;
- a literal that **mixes** the call with a plain element (`[array_slice(…), [9, 9]]`) widens the merge to the right type.

And because the fix reads `ctx.builtin_call_types`, which answers for *every* builtin rather than only the array-returning ones, `[count($a), 7]`, `[strtoupper('a'), 'b']`, `[sqrt(4.0), 1.5]` and `[in_array(1, $a), false]` are pinned too — they are what hold the fix to reading the recorded type instead of assuming one.

## Ordering note: this wants #1054 to land first

One shape changes from a compile refusal to a crash if this merges alone:

```php
$c = [array_slice($a, 0, 2), array_slice($a, 2)];
foreach ($c as $s) { echo implode(",", $s), ";"; }
```

- on `main`: refused, `implode array PHP type Int` (the binding is a scalar, so it never reaches the renderer);
- with **this PR alone**: the binding is now correctly an array, reaches `__rt_implode`, and **SIGSEGVs** — that is #1081, whose fix is #1054;
- with **this PR + #1054**: `1,2;3,4,5;`, matching PHP.

I verified that combination locally by applying this change on top of #1054's branch. Nothing here regresses anything that works today — the shape does not compile on `main` — but it is strictly better to merge #1054 first, and I am happy to rebase this onto it if you prefer a stacked review.

## Tests

`tests/codegen/arrays/nested.rs`, 6 tests. The 4 issue tests fail on `main`; the 2 controls pass there, which is the point of having them.

Suites run locally on `main` @ `bf5437b687`: `codegen::arrays` 584/584, `codegen::strings` 352/352, `codegen::objects` 254/254, `runtime_gc` 301/301, `codegen::types` 320/320, `ir_lower` unit 44/44.


Guikingone:
Measured against this branch (built at `0753811228`): **this stamp is still a miscompile for one class of builtin**, and I have a fix for it.

```php
$c = [array_map(fn(int $x): int => $x * 2, [1, 2, 3])];
echo $c[0][1];
// PHP:        4
// this branch: 4341024904      ← a Mixed cell pointer, read as an array element
```

## Why

The element type is taken from `builtin_call_types` verbatim. That map holds the **checker's** type. The value the call produces is typed by `resolve_registry_builtin_result_type`, which overrides the checker for a `Declared` or `Shared` result contract, and re-derives its own for a `Checked` one whose runtime target rejects the checked type against the operands it was really given.

`array_map()` is the clearest case: its check hook answers `array<int>` for a callback declared `: int`, while its `eir_result_type` returns a boxed `Mixed` cell — deliberately, and its own doc says so. So the literal is stamped `array<array<int>>` over a container of boxed cells, and nothing at the push reconciles them: `coerce_array_literal_element_to_storage_type` converts only `Int`, `Bool`, `Float`, `Str`, and a heap stamp is stored as-is.

`array_pop` on a nested source and `preg_split` have the same disagreement.

## The fix

A **scalar** result is taken precisely — a scalar stamp is converted at push, so a disagreement there is repaired rather than misread. Anything else is stamped `Mixed`, which is the one element type that is right whatever the call's own contract answers: a concrete value is boxed into it on the way in, a `Mixed` value is stored as-is. It is also what a literal mixing a builtin call with a plain element already merged to — which is why `[array_slice(...), [9, 9]]` was correct all along while `[array_slice(...)]` alone was not.

## Where it is

I hit the same defect independently and pushed **#1108**, which is this fix plus the `Mixed` rule. Three reviewers (Kimi K3, GLM 5.3, DeepSeek) found the verbatim-stamp defect in my first cut before I did; the shapes above are theirs, reproduced against host PHP.

#1108 also covers the **bool** half, which neither this PR nor #1096 mentions: every unlisted bool builtin alone in a literal became `int(1)`, in both literal kinds —

```php
var_dump([in_array(1, [1, 2])][0]);                  // was int(1),  PHP bool(true)
var_dump([is_array([1])][0]);                        // was int(1),  PHP bool(true)
var_dump(["k" => str_contains("abc", "b")]["k"]);    // was int(1),  PHP bool(true)
```

These two PRs are duplicates and only one should land. Mine is the later one, so I have no objection to closing it and bringing the `Mixed` rule and the bool tests here instead — whichever you prefer. What should not land is the verbatim stamp on its own.

nahime0 (changes requested):
## Changes requested

`array_literal_element_type_for_ir` / `assoc_array_literal_value_type_for_ir` now consult `ctx.builtin_call_types` before `infer_expr_type_syntactic`. That correctly stops the #1096 path where a builtin fell through to syntactic `Int`, stamped `array<int>`, and `coerce_array_literal_element_to_storage_type` cast the returned array to `1`.

The stamp is still taken from the **checker** map verbatim. The value actually stored is typed by `resolve_registry_builtin_result_type`, which overrides `Checked`/`Shared` contracts. Example: `array_map`'s check hook records `array<int>` while `eir_result_type` is `Mixed`. Coercion only converts `Int`/`Bool`/`Float`/`Str`, so a heap stamp is stored as-is and a later indexed read can treat a Mixed cell as an int (garbage / wrong shape). The same hole exists for `array_slice` over an untyped/`Mixed` ABI source, plus `array_pop` / `preg_split`.

## What to change before merge
- Do not stamp the checker type verbatim for Shared/Checked builtins whose EIR result is `Mixed`. Prefer the scalar-precise / else-`Mixed` rule in #1108, or call `resolve_registry_builtin_result_type` at the stamp site, so the literal element type matches what lowering actually produces.

## Related (non-blocking here)
- #1107 — span-only `builtin_call_types` keys
- Solo bool-builtin literals (`[in_array(...)]`) should be fixed by a correct stamp path but are only covered in mixed form today
- Rebase / land around #1054 before `foreach`+`implode` on these literals (#1081)

Guikingone:
Closing in favour of #1108, which is the change this review asks for.

@nahime0 your requested change — do not stamp the checker type verbatim for a builtin whose EIR result is `Mixed`; take a scalar result precisely and stamp everything else `Mixed` — is exactly the rule #1108 implements, and both PRs close #1096. Keeping this one open would mean carrying the verbatim stamp you identified as still miscompiling `[array_map(fn(int $x): int => $x * 2, [1, 2, 3])]`, alongside the PR that removes it.

The related items you listed (#1107, the solo bool-builtin literal, and landing around #1054 / #1081) carry over to #1108.

Issue #1098, opened 2026-09-19
Promoting [] to a hash inside a function releases the array before array_to_hash reads it (hang / bad refcount)


Guikingone:
## Summary

Promoting an empty array local to a hash **inside a function body** emits a `release` of the array *before* `array_to_hash` reads it. The freed table is still used, and a later reindexing sort walks it: the program hangs, or fails under `--heap-debug` with `heap debug detected bad refcount`.

## Reproduction

```php
<?php
function f(): int { $h = []; $h["b"] = 2; $h["a"] = 1; rsort($h); return count($h); }
echo f();
```

```
$ php a.php                       2
$ elephc -q a.php && ./a           (hangs forever)
$ elephc -q --heap-debug a.php && ./a
Fatal error: heap debug detected bad refcount
```

The same body at **top level** is correct:

```php
<?php
$h = []; $h["b"] = 2; $h["a"] = 1; rsort($h); echo count($h);   // 2, correct
```

## The IR shows the release before the read

`--emit-ir` for the function above:

```
v0: Heap(Array) php=array<never> own=owned  = array_new capacity=0
v1: Heap(Array) php=array<never> own=owned  = acquire v0
store_local v1 slot[0]
release v0
...
v2: Heap(Array) php=array<never> own=maybe_owned = load_local slot[0]
v3: Str  php=string own=persistent = const_str data[0]
v4: I64  php=int    = const_i64 2
v5: Heap(Mixed) php=mixed own=owned = load_local slot[0]     ← reloaded as Mixed
release v5                                                    ← RELEASED HERE
v6: Heap(Hash) php=array<mixed, int> own=owned = array_to_hash v2   ← then read
hash_set v6 v3 v4
store_local v6 slot[0]
```

`release v5` drops the local's array, and `array_to_hash v2` then converts that same (now freed) table. Note also that `v5` is typed `Heap(Mixed) php=mixed` for a slot that holds `Heap(Array) php=array<never>` — the reload's type does not match the storage.

The top-level version never emits that release: there the promotion happens at the literal, before any store.

## Perimeter

Everything below is one function body unless marked otherwise.

| shape | result |
| --- | --- |
| `$h = []; $h["b"] = 2; $h["a"] = 1; rsort($h); return count($h);` | **hang** |
| same with `sort()` | **hang** |
| same, `return $h[0];` after the sort | **hang** |
| same, `foreach` after the sort | correct (2, 3) |
| same with `asort()` (key-preserving) | correct |
[...]

PR #1099, opened 2026-09-19
fix(arrays): array_keys() reads a hash key's runtime form, not its declared type

### LINKED ISSUE #1072: array_keys() after sort()/rsort() of a promoted associative array crashes: keys are ints, checker still says strings
## Summary

After `sort()` / `rsort()` on an associative array that was promoted from `[]` by string-key writes, the keys are integers at runtime (PHP reindexes) but the checker keeps the array typed with string keys. `array_keys()` then produces a "string" list whose entries are the runtime's int-key sentinel, and consuming it crashes.

## Reproduction

```php
<?php
$h = [];
$h["b"] = 2; $h["a"] = 1;
sort($h);
var_dump(array_keys($h));
echo implode(",", array_keys($h)), "\n";
```

- PHP 8.5: `array(2) { [0]=> int(0) [1]=> int(1) }` then `0,1`
- elephc (`feat/core-align`, `df7d817d3`): compiles, then SIGSEGV in `__rt_str_persist_copy` with a length of `-1` (the `key_hi == -1` inline-int-key sentinel read as a string length). On linux-aarch64 the same shape surfaced as `heap debug detected bad refcount`.
- elephc `main` (`129bf261e`) refuses the program at compile time instead: `EIR backend error: unsupported EIR backend feature: array_keys associative key PHP type Mixed into result PHP type Str`.

So the branch turned a compile-time rejection into a miscompile. Found while writing a regression for the reindexing hash sort on #893 (the fixture now avoids reading the keys back).

## Where to look

The static key type after a reindexing sort: `lower_hash_reindexing_sort` rebuilds the table with keys `0..n-1` (`__rt_array_slice_to_hash`), but the receiver keeps its `AssocArray { key: Str, .. }` type, and `array_keys` lowering trusts it (`src/codegen/lower_inst/builtins/arrays/misc_dispatch.rs`, `values.rs`). Either the sort should retype its receiver to int keys (as PHP's semantics say) or `array_keys` on a hash must dispatch on the runtime key form.

Guikingone:
Fixes #1072.

```php
$h = [];
$h["b"] = 2; $h["a"] = 1;
sort($h);
var_dump(array_keys($h));   // PHP: [0, 1]   elephc: SIGSEGV
```

## What it is

`sort()`/`rsort()` renumber a hash to `0..n-1`. The receiver keeps its `AssocArray { key: Str }` type across the by-reference call — `record_reference_alias_root` pins by-ref builtin arguments as never retype-eligible — so at `array_keys($h)` the declared key type says string while the runtime keys are integers.

`__rt_hash_iter_next` returns `(cursor, key_ptr, key_len)` with **`key_len == -1` marking an integer key**. The key materializer dispatched on the *declared* key type, so the `Str` arm called `__rt_str_persist` unconditionally and persisted that `-1` as a string length: a read at `ptr[-1]` with a huge unsigned count.

`foreach` over the same array was always right, because it reads the key form at run time. That is what the fix makes `array_keys()` do too.

## The change

- **`emit_assoc_mixed_key_append_{aarch64,x86_64}`**: `Str` is folded into the `Mixed` arm, so it takes the same runtime sentinel dispatch (`cmn x2, #1` / `cmp rdx, -1`) instead of assuming a string.
- **`array_keys()`'s check hook**: a string-keyed hash answers `Array<Mixed>` rather than `Array<Str>`, because `Array<Str>` has nowhere to put an integer key. An int-keyed hash is unchanged — reindexing an int-keyed hash still yields int keys.

`asort()`/`ksort()` only relink the chain, so their keys still come back as strings; a hash holding both forms now renders each as itself. That is what pins the fix to *reading* the key form rather than assuming every hash key became an integer.

## Before → after

| fixture | before | after |
| --- | --- | --- |
| `sort($h); array_keys($h)` on a string-keyed hash | **SIGSEGV** | `[0, 1]` |
| `rsort($h); array_keys($h)` | **SIGSEGV** | `[0, 1]` |
| same on a hash *literal* rather than promoted from `[]` | **SIGSEGV** | `[0, 1]` |
| same with string values | **SIGSEGV** | `[0, 1]` |
| `asort($h); array_keys($h)` | `["a","b"]` | `["a","b"]` |
| `ksort($h); array_keys($h)` | `["a","b"]` | `["a","b"]` |
| `["b" => 2, 5 => 1]` → `array_keys()` | `["b", 5]` | `["b", 5]` |
| **`array_keys([1, "b", 2.5])`** | *refused:* `array_keys associative key PHP type Mixed into result PHP type Int` | `[0, 1, 2]` |

Every fixture matches host PHP 8.5.10.

## From the multi-model review

Kimi K3, GLM 5.3 and DeepSeek each reviewed this branch with full read access to the tree. None found an ownership bug, a sizing mismatch or a register error; all three independently verified the boxed-key ownership chain (`__rt_str_persist` → `emit_box_current_owned_string_as_mixed` → one slot → `__rt_array_free_deep` tag-7 walk → `__rt_decref_any`), that `indexed_key_element_size(Mixed)` = 8 agrees with the tag-7 stamp and the `lsl #3` append offsets on both arches, and the x86_64 register usage against `__rt_hash_iter_next`'s ABI.

Two findings came back and **both are addressed in this PR**:

1. **GLM and DeepSeek converged on `keys.rs:458`/`:488`**: the old `Str` *result* arm still called `__rt_str_persist` unconditionally, so it still contained the exact #1072 bug. It was unreachable only because the check hook had stopped producing `Array<Str>` — "the fix rests entirely on the check hook". Fair. Those arms are now deleted, `require_supported_assoc_result_type` no longer *accepts* the `(Str key, Str result)` pair at all, and the two string-key append helpers that only it used are gone. A hash can no longer be materialized into an `Array<Str>`: the backend refuses at compile time rather than emitting code that reads `ptr[-1]`.
2. **GLM flagged a sibling gap** (with an explicit caveat that they had not confirmed it reachable): `array_keys()` on a statically `array<mixed>` receiver is a hard codegen error. It reproduces — `array_keys([1, "b", 2.5])` does not compile on `main`. Same root cause as this issue: an `array<mixed>` value can be hash-backed at run time (`lower_dynamic_mixed_array_keys` exists to branch on exactly that), so its keys can be strings and an `Array<Int>` result cannot hold them. The check hook now answers `Array<Mixed>` for that receiver too, which is the last row of the table above.

## Tests

`tests/codegen/arrays/key_sort.rs`, 6 tests: the repro, the `rsort`/literal/string-value shapes, the key-preserving controls (`asort`/`ksort`/untouched), a both-key-forms hash, the heterogeneous-literal receiver, and a 50-iteration `--heap-debug` fixture over both a key-preserving and a reindexed hash — so a release emitted twice fails as a double free rather than passing quietly.

Suites run locally on `main` @ `bf5437b687`: `codegen::arrays` 583/583, every test matching `array_keys` 13/13, `runtime_gc` 301/301, `codegen::strings` 352/352, `codegen::eval`, `codegen::types`.

## Found while working on this, filed separately

**#1098** — promoting `[]` to a hash **inside a function body** emits a `release` of the array before `array_to_hash` reads it. `--heap-debug` reports `heap debug detected bad refcount`; without it the program hangs. It needs no `array_keys()` at all, and it is why the tests here use top-level code or hash literals rather than in-function promotion.


nahime0:
What changed in this PR:
- After `sort`/`rsort` on a string-keyed hash, `array_keys()` no longer treats `key_hi == -1` as a string length; materialization uses the key's runtime form (same sentinel as `__rt_hash_iter_next`).
- Checker hook: string-keyed hash / `array<mixed>` → `Array<Mixed>` result; int-keyed stays `Array<Int>`.
- Removed `Array<Str>` result arms and string-slot append helpers; `(Str key, Str result)` is rejected at compile time.

Follow-ups:
- #1186 — after by-ref `sort`/`rsort` the checker still leaves the receiver as `AssocArray{key:Str}` while runtime keys are int; today only `array_keys` papers over it
- #1187 — pin `usort` / `natsort` / `array_multisort` on the same runtime-key path
- #1188 — `indexed_key_element_size` still special-cases `Str=16` (dead layout)

Already open and independent: #1098 (`[]`→hash promotion inside a function releases before `array_to_hash`).

PR #1104, opened 2026-09-19
test(arrays): pin array_push growth through a by-reference parameter

### LINKED ISSUE #1088: codegen: pin by-ref array_push through __rt_array_grow
## Context

Follow-up from Grok review of #1038 / #677.

#1038 republishes the receiver between values because each append can `__rt_array_grow`. The growth fixture (many values into a small array) covers a **plain local**. The by-ref fixture (`viaRef`) only pushes two values onto `[1]` and likely never reallocates.

Write-back through `RefCell` / `store_value_to_local` looks correct in review, but it is not pinned under growth.

## Acceptance

Codegen (or heap-debug) test: `function f(array &$a) { array_push($a, …many…); }` starting from a 1-element array, forcing at least one `__rt_array_grow`, with correct final contents/count and a clean heap summary.

Related: #1038, #677.

Guikingone:
Closes #1088.

One test, no source changes. `array_push()` republishes its receiver between values because each
append can reach `__rt_array_grow`, which frees the previous buffer right after handing back the
new pointer.

For a plain local that republish is `store_value_to_raw_local`. For a **by-reference parameter**
it is a different write: the receiver is a reference cell, loaded with `LoadRefCell` and
published with `store_value_to_ref_cell_local` *through the cell pointer* into the caller's
frame. Nothing pinned that second store — the by-ref fixture in
`test_array_push_accepts_phps_full_variadic_signature` pushes two values onto `[1]`.

## The test has teeth, measured

Disabling **only** the reference-cell arm of `store_value_to_local`, leaving the plain-local arm
intact, makes this the single failing test out of the **745** in `arrays::`. Disabling both arms
fails six — so the suite covered the local path thoroughly and covered this one not at all.

Under lldb it reaches `__rt_array_grow` **39** times, all through `__rt_array_push_int` / `_str`
— the typed lowering, not the boxed `__rt_mixed_array_append` path, which republishes inside the
Mixed cell and would pin nothing here.

## The variants

- **all at once** — the plain growth republish;
- **one per call** — the hedge against a future bulk-append lowering that would drain the first
  variant of reallocations while keeping it green;
- **shared with a copy** — growth follows a copy-on-write split, and `count($copy)` pins it;
- **string values** — a stale pointer there is a use-after-free of the elements, not only of the
  buffer;
- **a forwarding wrapper** — the reference crosses two frames.

The **stdout** assertion is the load-bearing one: a dropped republish is a use-after-free rather
than a leak, so the caller reads a stale count instead of 33. The heap assertion catches the
copy-on-write variant's orphan and any double release, which reports as a fatal before the
summary prints.

Behaviour is correct today on all of them, so this is coverage rather than a fix. Expected values
are verbatim host PHP 8.5.10 output.

```
cargo test --release --test codegen_tests arrays::indexed  156 passed, 0 failed, 1 ignored
cargo test --release --test codegen_tests array_push        33 passed, 0 failed
cargo test --release --test codegen_tests arrays::         745 passed, 0 failed, 1 ignored
```

## Review

Reviewed by Kimi K3, GLM 5.3 and DeepSeek. All three traced the by-ref receiver to the
reference-cell store and agreed the test does not re-pin the local path.

Two things they corrected, both of which changed what this PR says:

- **DeepSeek caught an arithmetic contradiction in my own justification.** I had claimed the
  existing `viaRef` fixture reaches `__rt_array_grow` zero times; DeepSeek derived from the
  capacity model (`[1]` → capacity 1, `max(old * 2, 8)`) that it must grow once, and that my
  count of 19 for the new fixture was inconsistent with the same model. Both of my numbers were
  measurement errors — a `grep` that did not match lldb's auto-continue output. Re-measured: the
  old fixture grows **1** time and the new one **39**. The justification now leads with the
  mutation evidence, which does not depend on either number.
- **DeepSeek also pointed out that the heap assertion is not what catches this failure** — the
  stdout one is, because a missed republish is a use-after-free rather than a leak. The doc
  comment said the opposite; it now says which assertion catches what.

GLM flagged that the whole coverage claim hinges on `array` resolving to `PhpType::Array(_)`
rather than `Mixed`, and gave the disambiguating experiment: if the boxed path were taken, the
republish would happen inside the Mixed cell and the test would pin nothing. Measured — 265
typed int pushes and 39 typed string pushes against 1 boxed append — so the typed path is taken.

Kimi noted that the one-per-call variant collapses to the same lowering as all-at-once today, and
argued for keeping it as the bulk-append hedge; that reasoning is now in the test's doc comment.
Kimi also observed the fixture is not self-checking about growth having occurred — if literal
allocation ever pre-sizes, it could keep passing while reaching `__rt_array_grow` zero times,
which is how the old fixture came to pin so little. The one-per-call variant is a partial hedge;
the measured counts are recorded in the comment so the premise can be re-checked rather than
assumed.


nahime0:
What changed in this PR:
- Test-only pin for [#1088](https://github.com/illegalstudio/elephc/issues/1088): `array_push` growth through an `array &$a` parameter is heap-clean.
- Covers bulk push, one-per-call, COW split, string elements, and two-frame forward; stdout matches PHP; heap summary catches COW orphans / double-free.
- No production code changes. Closes the gap left by the 2-value `viaRef` fixture in `test_array_push_accepts_php_full_variadic_signature`.

Follow-ups:
- #1189 — fixture does not prove `__rt_array_grow` actually ran (pre-size / bulk-append could stay green)
- #1190 — property / static / nested-element by-ref receivers still only have 2-value coverage
- #1191 — Mixed boxed append growth republish not pinned (typed `Array(_)` only)

PR #1106, opened 2026-09-19
docs(codegen): the slice prologue is no longer indexed-only, and the flag is saved first

### LINKED ISSUE #1093: nit: dedupe lower_array_slice rustdoc; save AArch64 preserve_keys before emit_slice_bounds
## Context

Cosmetic follow-up from Grok review of #1042 — **not a merge blocker**.

1. Duplicated rustdoc on `lower_array_slice` in `reduce_sets.rs`.
2. AArch64 saves the `preserve_keys` flag (`x4`) *after* `emit_slice_bounds`. Safe today (prologue only clobbers `x9`/`x10`); x86_64 is more conservative and spills first. Optionally save before the prologue for consistency.
3. `slice_bounds.rs` docs still describe themselves as indexed-only, but the hash path reuses them.

## Acceptance

Docs/comments match reality; optional AArch64 save-order aligned with x86_64. No behavior change required.

Related: #1042.

Guikingone:
Four in-source corrections from the Grok review of #1042. None of them changes what a program
computes, and the one that touches emitted code is verified at the machine-code level.

## The duplicated rustdoc, twice over

`lower_array_slice`'s doc comment was pasted twice, verbatim — seven lines of it.

`lower_slice_like_args` in the sibling `slice_splice.rs` had the same defect **three** times, and
the first of the three is **stale**: it predates the `length_present` flag, still calls the tuple
a triple, and describes two staged integers where there are now three. That is the copy a reader
meets first, which is what makes this worth a commit rather than a shrug. Found by the DeepSeek
reviewer, not by me.

## The prologue is not indexed-only

`slice_bounds.rs` described itself as serving "every slice-like **indexed-array** runtime helper"
and listed four callers. There are seven, one of them hash-backed:

| caller | |
|---|---|
| `array_slice` | |
| `array_slice_refcounted` | |
| `array_slice_to_hash` | |
| `array_splice` | |
| `array_splice_refcounted` | |
| `array_splice_str` | |
| `hash_slice` | added in #1042 |

Its ARM64 ABI line called `x0` a "source **indexed-array** pointer" for the same reason, while the
hash caller passes a hash. The prologue itself never cared: it reads the source length from `[x0]`,
which is the element count for one and the entry count for the other.

## The one real change: the flag is saved first

`__rt_hash_slice`'s AArch64 side took `mov x21, x4` — the `preserve_keys` flag out of its argument
register — **after** `emit_slice_bounds`. The x86_64 side of the same helper spills its flag to
`[rbp - 88]` **before** the prologue. Both are safe; the point is that they should not disagree
about whether the ordering matters, because one of them would have to re-read the prologue's
clobber list the next time it changes. AArch64 now saves first as well.

Verified at the machine-code level rather than by reading the emitter — `otool -tV` on a linked
binary:

```
__rt_hash_slice:
    sub  sp, sp, #0x80
    stp  x29, x30, [sp, #0x70]
    add  x29, sp, #0x70
    stp  x19, x20, [sp, #0x50]
    stp  x21, x22, [sp, #0x60]
    mov  x21, x4          ← immediately after the callee-saved pair
    ldr  x9, [x0]         ← the prologue's first instruction
```

## Verification

Eight associative `array_slice()` shapes diff byte-identical against host PHP 8.5.10: both
`preserve_keys` modes, a negative offset, an omitted length, a negative length, an offset past the
end, and an int-keyed hash in both modes.

- `cargo test --release --test codegen_tests arrays::` — 744 passed, 0 failed, 1 ignored.
- `cargo test --release --test codegen_tests slice` — 40 passed, 0 failed.

No test is added: nothing observable changed, and the existing hash-slice coverage in `arrays::`
exercises the reordered sequence on every call.

## Review

Kimi K3, GLM 5.3 and DeepSeek reviewed this against the base. All three read the whole
`emit_slice_bounds` sequence, including its five branch targets, and independently confirmed:

- it writes only `x1`/`x2`/`x9`/`x10` and never touches `x4` or `x21`, matching its documented
  clobber list exactly;
- nothing between the function entry and the new position writes `x4`, and the caller sets it
  immediately before the `bl`;
- `x21` is saved by `stp x21, x22, [sp, #96]` *before* the new write and restored at the single
  exit;
- `__rt_hash_slice` has exactly one entry point — every internal label is branched to only from
  inside the helper — so no path bypasses the prologue;
- the seven-caller list is right in both directions.

None could name a program whose behaviour differs. Their one finding was the `slice_splice.rs`
stack, which is fixed here.

Closes #1093.


nahime0:
What changed in this PR:
- Dropped the duplicated `lower_array_slice` rustdoc in `reduce_sets.rs`.
- Collapsed the triple-stacked `lower_slice_like_args` rustdoc in `slice_splice.rs` to the current `(array, offset, length, length_present)` copy (the discarded first copy was stale).
- Updated `slice_bounds.rs` module docs: prologue is not indexed-only (seven callers); `x0` is a generic source-array pointer.
- Moved AArch64 `__rt_hash_slice` `mov x21, x4` before `emit_slice_bounds`, matching the x86_64 spill of `preserve_keys`.

Follow-ups:
- #1192 — `emit_slice_bounds` instruction comments still say \"indexed-array logical length\" on ARM64 and x86_64
- #1193 — kept `lower_array_slice` rustdoc still inverts `$preserve_keys` (same on `array_chunk` / `array_reverse`)

PR #1110, opened 2026-09-20
fix(closures): a closure returning a user call keeps the callee's type

### LINKED ISSUE #1028: A closure that returns a user function call is typed int: returned strings silently become 0

Guikingone:
```php
function tag(int $n): string { return "t" . $n; }
$a = function ($v) { return tag($v); };
echo $a(1);
// PHP: t1
// was: 0
```

A closure with no declared return type gets one from `closure_signature_from_ast`, and
`lower_return_expr` **coerces** the body's value into it. A wrong answer there is not a diagnostic,
it is a conversion.

`direct_closure_return_expr_type` looked a returned call up in `builtin_call_types` and, missing,
fell through to `infer_expr_type_syntactic`, whose `FunctionCall` arm is an allowlist of builtin
names ending in `_ => PhpType::Int`. A user callee matches nothing in that list, so the signature
said `int` and the coercion turned the callee's `string` into `0`.

## The issue blames the checker, and the checker is right

The analysis in #1028 points at `infer_closure_call_type` taking `infer_return_type_syntactic(body)`
(`inference/ops.rs:835`). That is not the path this program takes. Instrumenting
`resolve_closure_return_type` — which is what runs for `$a = function ($v) { return tag($v); }` —
prints `Str`: it collects returns through `collect_return_infos`, which calls
`self.infer_type(expr, env)`, the real inference with a symbol table.

So this is **two independent inferences of one type, disagreeing**, and the backend's is the one
the call site lowers against. The fix resolves the callee in `functions` / `extern_functions` and
runs the answer through the same two helpers a direct call's result goes through,
`eir_user_function_return_type` and `normalize_value_php_type`.

## The review round: "the same as a direct call" had to become true

All three reviewers (Kimi K3, GLM 5.3, DeepSeek) independently found the same hole in the first
cut, each with its own repro, and all three prescribed the same fix. The first cut took
`sig.return_type` **raw**:

```php
function f($x) { return [$x]; }          // untyped param, no declared return
$g = function (int $v) { return f($v); };
echo $g(5)[0];                            // PHP: 5   was: 4368664304
```

A callee with an untyped parameter receives it under the boxed-Mixed ABI whatever the checker
specialized, so a container it returns carries boxed elements — while the checker's signature still
says `array<int>`. `dynamic_param_container_return_type` is what widens that for every other
consumer, including the callee's own body and every direct call. Taking the raw type stamped the
closure `array<int>` over boxed storage, and nothing at the return boundary can notice:
`IrType::from_php` maps every `Array(_)` to one heap kind, and `coerce_container_to_return_type`
only ever widens *toward* `Mixed`.

Reproduced against host PHP before changing anything, and again after. Kimi's `AssocArray` variant
(`return ["k" => $i];`) and GLM's and DeepSeek's `[$x]` / `[1, 2]` variants are all in one new test.
Kimi's second finding — the missing `Never`→`Void` normalization the call path applies — is closed
by the same change, since `normalize_value_php_type` is now applied to both arms.

## More shapes than the issue lists, and two worse than `0`

```php
function half(int $n): float { return $n / 2; }
$a = function ($v) { return half($v); };
var_dump($a(3));                       // was int(1),  PHP float(1.5)

function mk(int $n): array { return [$n, $n + 1]; }
$a = function ($v) { return mk($v); };
count($a(1));                          // was REFUSED: count for PHP type Int

$a = function ($v) { return tag($v); };
strlen($a(1));                         // was a compiler PANIC:
                                       // checked builtin strlen failed backend-neutral EIR lowering
                                       // at 5:6: strlen cannot lower checked operand type Int
```

A callee declared **after** the closure that returns it behaves the same way, and so does an
immediately invoked closure literal.

## The threading got shorter

The closure-signature subtree carried `builtin_call_types` alone. It now carries
`ClosureReturnCallees { functions, extern_functions, builtin_call_types }` — one parameter where
there would otherwise be three, and the group is forwarded whole and read only at the leaf.

## Verification

15 shapes measured against host PHP 8.5.10, aarch64.

**Now correct:** the issue's five broken spellings (closure, arrow fn, hintless callee, no
parameters, typed parameter), an immediately invoked closure literal, a callee declared after the
closure, a `float` callee, an `array` callee, and `strlen()` of the result.

**Unchanged:** a declared closure return type (which skips the inference), a return through a local
(which gives lowering a typed binding), a body with no call, an `int` callee — the one the old
fallback answered right by accident — and `strtoupper` / `count` / `strlen` builtin callees. Three
of the twelve tests are controls pinning exactly those.

All twelve suites green on this commit, 6443 tests:

| suite | result |
|---|---|
| `closure_call_returns` | 12 passed, 0 failed |
| `callables::` | 586 passed, 0 failed |
| `closures::` | 98 passed, 0 failed |
| `arrays::` | 744 passed, 0 failed, 1 ignored |
| `oop::` | 621 passed, 0 failed |
| `runtime_gc::` | 301 passed, 0 failed |
| `strings::` | 394 passed, 0 failed |
| `regressions::` | 520 passed, 0 failed |
| `generators::` | 85 passed, 0 failed |
| `spl::` | 258 passed, 0 failed |
| `eval` | 1280 passed, 0 failed |
| `error_tests` | 1544 passed, 0 failed |

Docs: `docs/internals/the-ir.md` gains "Closure Return Types Inferred by Lowering" — the
per-expression-kind table of where the type comes from, and why the two inferences must agree.

## Known not fixed, reported on the issue

A closure returning `array_keys($v)` still refuses with `array_keys associative key PHP type Mixed
into result PHP type Int`. That is the **builtin** half of the same defect — the builtin arm reads
`builtin_call_types` by span and finds nothing for that call — and this change adds the user and
extern arms only.

Closes #1028.


Guikingone:
Closing as a duplicate of #1083, which came first and already has the load-bearing part.

I picked #1028 off the open-issues list without checking for an existing PR — my mistake. #1083 routes the callee's return type through `eir_user_function_return_type` rather than taking `sig.return_type` raw, which is exactly what three reviewers made me add to this branch after they proved the raw type miscompiles an untyped-parameter callee's container return:

```php
function f($x) { return [$x]; }
$g = function (int $v) { return f($v); };
echo $g(5)[0];        // PHP: 5, raw signature type: a cell pointer
```

So #1083 already lands the right design and nothing here needs to move over. The measurements from this branch are on #1028 for the record: the issue blames the checker and the checker is right (`resolve_closure_return_type` infers `Str`; the wrong `int` comes from lowering's own independent inference), and three shapes it does not list are worse than `0` — a `float` callee printed `int(1)`, an `array` callee refused to compile, and `strlen()` of the result panicked the backend.

One gap neither PR closes: a closure returning `array_keys($v)` still refuses with `array_keys associative key PHP type Mixed into result PHP type Int` — the builtin half of the same defect.

Issue #1115, opened 2026-09-20
Assigning a mixed value to an object-refined untyped property leaks three heap blocks per assignment


Guikingone:
`coerce_typed_assign_value`'s property-write caller leaks. The same coercion on a typed LOCAL is balanced; the property path is not.

## Repro

```php
<?php
class Item {
    public function __construct(public string $v) {}
}

class Holder {
    public $slot;
}

function pick(int $i): mixed {
    return $i >= 0 ? new Item("x") : null;
}

$total = 0;
for ($i = 0; $i < 200; $i++) {
    $h = new Holder();
    $h->slot = new Item("seed");
    $h->slot = pick($i);          // <- mixed RHS into an Object-refined property
    $total = $total + strlen($h->slot->v);
}
echo $total, "\n";
```

```
200
GC: allocs=1601 frees=1001
HEAP DEBUG: allocs=1601 frees=1001 live_blocks=600 live_bytes=25600
```

**600 live blocks for 200 assignments = 3 per assignment.** Output matches PHP (`200`); only the heap is wrong.

## It is a leak, and it is the mixed assignment

| program | live_blocks |
| --- | --- |
| the repro, 200 iterations | 600 |
| the repro, 400 iterations | 1200 |
| the same loop **without** `$h->slot = pick($i);` | **0 — `leak summary: clean`** |

So it scales linearly and the baseline is clean: it is the `mixed` → object-property store, not the objects themselves.

## Where

`lower_property_assign` (`src/ir_lower/stmt/instance_property_writes.rs:54`) routes the value through `coerce_typed_assign_value` (`src/ir_lower/stmt/typed_foreach.rs:60`), which emits `Op::MixedUnbox` when the target is `Object(_)` and the source is `Mixed`. `--emit-ir` on the repro shows it:

```
v13: Heap(Object) php=Item own=owned = mixed_unbox v12 ; span: 21:5
```

The identical coercion on a typed LOCAL is balanced — `Op::MixedUnbox` is an owning temporary (see the section added to `docs/internals/memory-model.md`), so `store_local` acquires and then releases the unbox's own reference. The property path's retain/release pair (`PropSet` + `release_property_assignment_source_after_retaining_store`, `src/ir_lower/stmt/property_array_writes.rs:214`) does not come out even.

## Not caused by the ownership fix in #983

[...]

Issue #1114, opened 2026-09-20
new ReflectionClass() leaks two heap blocks (2160 bytes) per construction


Guikingone:
Every `new ReflectionClass(...)` leaves two heap blocks behind, and `unset()` does not reclaim them. The leak scales linearly with the number of constructions, so a program that reflects in a loop — a container compiler, an ORM hydrator, any autoconfigure pass — grows without bound.

## Repro

```php
<?php
class Plain {}

$total = 0;
for ($i = 0; $i < 200; $i++) {
    $r = new ReflectionClass(Plain::class);
    $total = $total + strlen($r->getName());
}
echo $total, "\n";
```

```
$ elephc -q --heap-debug --gc-stats leak.php && ./leak
1000
GC: allocs=5201 frees=4801
HEAP DEBUG: allocs=5201 frees=4801 live_blocks=400 live_bytes=432000 peak_live_bytes=443552
HEAP DEBUG: leak summary: live_blocks=400 live_bytes=432000
```

**400 live blocks for 200 constructions = 2 per construction, 2160 bytes each time.**

## It is a leak, not a live value at exit

Doubling the loop doubles the residue:

| iterations | live_blocks | live_bytes |
| --- | --- | --- |
| 200 | 400 | 432000 |
| 400 | 800 | 864000 |

## It does not depend on the class being reflected

- a class with no attributes at all (the repro above): 400
- a class carrying two attributes: 400
- with `unset($r);` at the end of the body: 400 — the explicit release does not reclaim them either
- constructing without calling any method (`$r = new ReflectionClass(Plain::class);` alone): 400

So it is the construction itself, not a metadata slot that some accessor populates lazily, and not
the `$r` rebinding.

## Scale

`live_bytes` grows by 2160 bytes per construction. Reflecting 10 000 classes — the order of
magnitude a Symfony container compile reaches — is 21 MB that is never returned, on top of
whatever the program itself holds.

## Notes for whoever picks this up

- `--heap-debug` reports the leak but not the allocation site; the two blocks average 1080 bytes,
  which is larger than a bare object header, so the pair is worth identifying before fixing
  (`emit_reflection_attrs_property` in
  `src/codegen/lower_inst/objects/reflection/string_attrs_emit.rs` and the owner materializers
  around it are where the private slots are written).
- Because of this, an ownership test that touches `ReflectionClass` cannot assert
  `leak summary: clean` today; `tests/codegen/runtime_gc/reflection_attribute_filter.rs` works
[...]

Issue #1113, opened 2026-09-20
is_subclass_of()/is_a() answer false for a string class name, and the name-keyed hierarchy builtins refuse a runtime name


Guikingone:
`is_subclass_of()` and `is_a()` take a class NAME as well as an object in PHP. elephc answers `false` for every string form, silently, and the rest of the name-keyed hierarchy family refuses a non-literal name outright. Together that leaves compiled code with no way to ask "is class A a subclass of class B" when either name is only known at runtime.

## Repro

```php
<?php
class Base {}
class Derived extends Base {}
interface I {}
class Impl implements I {}

$o = new Derived();
echo "obj:",       is_subclass_of($o, "Base") ? "y" : "n", "\n";
echo "str:",       is_subclass_of("Derived", "Base") ? "y" : "n", "\n";
echo "iface:",     is_subclass_of("Impl", "I") ? "y" : "n", "\n";
echo "self:",      is_subclass_of("Base", "Base") ? "y" : "n", "\n";
echo "is_a_obj:",  is_a($o, "Base") ? "y" : "n", "\n";
echo "is_a_str:",  is_a("Derived", "Base", true) ? "y" : "n", "\n";
```

| | elephc | php 8.5.10 |
| --- | --- | --- |
| `obj` | `y` | `y` |
| `str` | **`n`** | `y` |
| `iface` | **`n`** | `y` |
| `self` | `n` | `n` |
| `is_a_obj` | `y` | `y` |
| `is_a_str` | **`n`** | `y` |

Three wrong answers, no diagnostic. Measured at `89a923f4b4`.

## Cause

`lower_is_a_relation` (`src/codegen/lower_inst/builtins/types.rs:619`) forwards to
`static_relation_holds`, which opens with:

```rust
let PhpType::Object(object_class) = ctx.value_php_type(object)? else {
    return Ok(false);
};
```

A string first operand is `PhpType::Str`, so the function returns `false` before it looks at
anything. The parent-chain and interface walks below it are correct — they are simply never
reached for the name form.

## The rest of the family refuses rather than answering wrongly

The name-keyed hierarchy builtins all work with a string LITERAL and reject a variable:

```php
<?php
class Base {}
class Derived extends Base {}

function probe(string $n): int { return count(class_parents($n)); }
echo probe("Derived");
```

```
error[5:41]: class_parents() first argument must be an object or string literal in AOT mode
```

The same for `class_implements()` and `class_exists()`. With a literal they are correct:
[...]

PR #1116, opened 2026-09-20
fix(reflection): honour the $name filter on getAttributes()

### LINKED ISSUE #983: ReflectionClass::getAttributes($name) ignores its filter and returns every attribute
`ReflectionClass::getAttributes($name)` ignores its filter argument and returns every attribute on the target.

### Measured

```php
<?php
#[Attribute] class A1x { public function __construct(public string $v = "x") {} }
#[Attribute] class B1x {}

#[A1x("one")]
#[B1x]
class Target {}

$r = new ReflectionClass(Target::class);
echo "all:", count($r->getAttributes()), "\n";
echo "filtered:", count($r->getAttributes(A1x::class)), "\n";
```

| | `php -n` 8.5.6 | eval bridge |
|---|---|---|
| `getAttributes()` | `2` | `2` |
| `getAttributes(A1x::class)` | **`1`** | **`2`** |

The unfiltered call and `getName()` on each result are both correct, so the attributes themselves are read properly — only the filter is dropped. `ReflectionAttribute::newInstance()` and `getArguments()` were checked in the same pass and match php on the ordinary path.

Not covered by this measurement: the `IS_INSTANCEOF` flag, and php's validation refusals (wrong target, non-repeatable attribute used twice). Those need their own cases.

### Why it matters

Symfony's container asks for one attribute at a time and assumes the answer is filtered: `Compiler/RegisterAutoconfigureAttributesPass.php` reads `getAttributes(Autoconfigure::class)` and throws `AutoconfigureFailedException` as soon as it sees more than it expects, so the first class carrying two different attributes turns a silent wrong answer into an exception several frames away. `AutowirePass` and the kernel's attribute scanning use the same shape.

### Severity

Silent wrong value: nothing signals it at the call, and the consequence appears elsewhere as a wrong branch or an unrelated-looking exception.

Guikingone:
Fixes #983.

`ReflectionClass::getAttributes($name)` ignored its argument. The synthesized method declared
`params: Vec::new()` and returned `$this->__attrs` whole, so the compiled path refused the call
("Method ReflectionClass::getAttributes expects 0 arguments, got 1") and the eval bridge — which
reaches the same method through the runtime hook, where arity is not checked — answered with every
attribute on the target. **The issue's perimeter is half right: the silent wrong answer is the eval
path only; AOT was a compile error.**

## The filter

`getAttributes(?string $name = null, int $flags = 0)`, on all five owners that carry attributes and
on the eval path. The comparison is `strcasecmp`, because PHP folds ASCII case for a class name and
does not resolve a leading separator. Measured on 8.5.10: `getAttributes("markerone")` finds
`#[MarkerOne]`, `getAttributes("\MarkerOne")` does not, and `#[\App\Marker]` is found by
`Marker::class` and `"App\Marker"` but not `"\App\Marker"`.

## Two things that had to be settled

**The element representation differs by who populated `__attrs`.** The AOT emitter stamps the array
`Object(ReflectionAttribute)` and stores bare pointers; the eval bridge builds a plain Mixed array.
The caller reads the result as `array<ReflectionAttribute>` because `patch_reflection_attribute_result`
patches the signature that way. Pushing the loop variable straight through stored a Mixed box the
caller misread: `get_class($f[0])` answered `JsonException`, `newInstance()` segfaulted. Retyping
`__attrs` instead fixed AOT and **segfaulted the eval path**. Landing each match in a typed local
normalizes both representations.

**That typed local is the compiler's only `Op::MixedUnbox` site**, and the op's lowering ends in
`emit_unbox_mixed_to_owned_refcounted_result` — the value already owns a reference. The EIR read it
as a borrow and acquired a second one, which the single slot release in the epilogue never balanced.
The emitted ARM64 shows `bl __rt_mixed_unbox` followed by two `bl __rt_incref`. Five heap blocks
leaked per matched attribute, scaling linearly. `value_is_owning_temporary` now says the unbox is
owning. The coercion's other caller — a typed property write — has a leak of its own that predates
this and is unaffected by it either way (#1115, measured with the predicate on and off).

## `IS_INSTANCEOF` is refused, not approximated

`ReflectionAttribute::IS_INSTANCEOF = 2` is declared, since PHP has exactly that one constant, but
the flag is refused. It needs a subclass test on a class name known only at runtime, and no AOT
builtin answers one: `is_subclass_of()` reports `false` for a string first operand, and
`class_parents()` / `class_implements()` / `class_exists()` reject a non-literal name (#1113).
Answering by exact name would return a subset of PHP's answer with no diagnostic, and the call was
already a compile error.

Refusing everything but `0` costs nothing: PHP accepts exactly `0` and `2` and raises
`ValueError: Argument #2 ($flags) must be a valid attribute filter flag` for the rest (measured with
1, 3, 4).

**The refusal has two halves, because one cannot cover every spelling.** A checker guard is keyed on
call syntax, and seven spellings reach the same body. The checker refuses the four where the
argument list is visible — positional, named (`flags:` can sit at index 0), through a spread, and on
a `mixed` receiver, which dispatches on the runtime class id without reaching the class path. The
body throws a `ReflectionException` for the other three, which hand it arguments with no visible
list: a first-class callable, `call_user_func_array`, and a dynamic method name (desugared to
`call_user_func` by the parser). All three were measured answering `1` where PHP answers `2`.

Three spellings stay accepted, because PHP answers them with no subclass test: a null name, no name
at all (`getAttributes(flags: 2)` — an absent `$name` is null), and a `$flags` folding to `0`,
including a literal `false`. The `mixed`-receiver refusal only fires when every class declaring the
method is a Reflection owner.

## Tests

- `tests/codegen/oop/reflection_attribute_filter.rs` — 26 behaviour tests: the repro, every owner,
  case folding, a namespaced attribute, a repeatable attribute, a class with none,
  `getName()`/`newInstance()` through a filtered element, the eval path, and every flag spelling,
  refused and accepted alike.
- `tests/codegen/runtime_gc/reflection_attribute_filter.rs` — 4 ownership tests. They cannot assert
  `leak summary: clean` because `new ReflectionClass()` leaks two blocks per construction on its own
  (#1114), so each compares two programs that construct the same number of owners and differ only in
  the call. Verified by mutation: forcing the ownership predicate to `false` fails three of the four
  while every behaviour test still passes.

Docs: `docs/php/classes.md` (the filter, the refused flag and every spelling),
`docs/internals/memory-model.md` (the unbox ownership rule), `examples/attributes/main.php`.

Reviewed in two rounds by Kimi K3, GLM 5.3 and DeepSeek; five findings were reproduced against host
PHP and fixed here (the spread bypass, the `mixed`-receiver bypass, the three indirect spellings,
the absent-name and `false` false-refusals, and the over-broad `mixed` refusal), and three more were
measured and discharged — a subclass of a Reflection owner does not compile at all, a namespaced
attribute matches PHP, and the property-write path's leak is pre-existing.


Guikingone:
Correct, and the remedy it names is the right one.

`get_attributes_filter_arguments` answered `Hidden` for every `Spread`, and `Hidden` is refused —
so `$r->getAttributes(...['name' => Marker::class, 'flags' => 0])` became a compile error for a
list the rest of the compiler normalizes before anything else looks at it.
`expand_static_assoc_spread_args` is the expander the call planner itself runs, and running it
first makes the filter read the spelling PHP accepts.

The `Hidden` arm stays, and has to: a spread whose contents are a runtime array survives the
expansion as a `Spread`, and nothing there can tell `$flags` from absent. That one is still
refused, which is the honest answer.

Measured against host PHP 8.5.10 on two attributes, one of them filtered:

| call | elephc before | elephc after | PHP |
| --- | --- | --- | --- |
| `getAttributes()` | 2 | 2 | 2 |
| `getAttributes(Marker::class)` | 1 | 1 | 1 |
| `getAttributes(name: Marker::class)` | 1 | 1 | 1 |
| `getAttributes(...['name' => Marker::class])` | **compile error** | 1 | 1 |
| `getAttributes(...['name' => Marker::class, 'flags' => 0])` | **compile error** | 1 | 1 |

And the runtime form is still refused, as it must be:

```php
$args = [];
$args['name'] = Marker::class;
$r->getAttributes(...$args);   // still: cannot be read through a spread
```

`docs/php/classes.md` said a spread means "the flag cannot be read at all, so the call is
refused". That is now true only of the runtime form, so the sentence distinguishes the two rather
than being left to contradict the behaviour — the same obsolete-documentation shape flagged on
#1217.

Two fixtures added: the five spellings, and the runtime spread still refused. The first fails with
the fix reversed, and nothing else in the file does.

Issue #1119, opened 2026-09-20
ReflectionParameter::getDeclaringFunction()->getParameters() is always empty


Guikingone:
`getDeclaringFunction()` hands back a reflector whose parameter list is empty, whatever the function takes.

## Repro

```php
<?php
class Demo { public function plain(int $a, string $b): bool { return true; } }
function freeFn(int $x, string $y): void {}

$p = (new ReflectionMethod('Demo', 'plain'))->getParameters()[0];
echo "method params:", count($p->getDeclaringFunction()->getParameters()), "\n";

$q = (new ReflectionFunction('freeFn'))->getParameters()[0];
echo "fn params:", count($q->getDeclaringFunction()->getParameters()), "\n";
```

| | elephc | php 8.5.10 |
| --- | --- | --- |
| `method params` | **0** | 2 |
| `fn params` | **0** | 2 |

`getNumberOfRequiredParameters()` is correct on the same object, so the count and the list disagree with each other as well.

## Cause, and why it is not an oversight

`emit_reflection_parameter_declaring_function_property` (`src/codegen/lower_inst/objects/reflection/parameter_property_emit.rs`) builds the reflector from `empty_reflection_metadata()` and fills the name, class, flags, counts and return type — but not `parameter_members`. `ReflectionDeclaringFunctionMember` (`src/codegen/lower_inst/objects/reflection.rs`) has no parameters field at all.

That is load-bearing. Emitting the parameters would emit a `ReflectionParameter` object for each, and each of those emits its own `__declaring_function` — a `ReflectionMethod` that would again emit its parameters. The emission does not terminate. The empty list is what breaks the cycle.

## What #1080 did and did not do

#1080 gave these reflectors a `__toString()` dump, which would have claimed `Parameters [0]` for a method that has two. A dump is a STRING and carries no cycle, so it is rendered from the real member (`reflection_declaring_function_dump`) and carried across in `ReflectionOwnerMetadata::rendered_to_string`. The dump is therefore correct while `getParameters()` is still empty — the two now disagree, which is a reason to fix this rather than a new defect.

## What a fix needs

The cycle has to be broken somewhere other than by dropping the data. Options worth weighing:

- emit the parameter objects WITHOUT their own `__declaring_function` when they belong to a declaring-function reflector (one level of nesting, then stop);
- make `__declaring_function` lazy — materialize it when the property is read rather than when the parameter is emitted;
[...]

Issue #1117, opened 2026-09-20
eval ReflectionMethod/Function __toString() differs from PHP in three places the compiled path gets right


Guikingone:
The eval bridge has rendered `ReflectionMethod::__toString()` and `ReflectionFunction::__toString()` since #451, and #1080 gives the compiled path the same surface. Comparing the two on the same program turns up three places where eval differs from PHP and the compiled path does not.

## Repro

```php
<?php
interface Shape { public function area(): float; }

abstract class Base implements Shape {
    abstract public function area(): float;
    public function none() {}
}

function freeFn(array $a = []): void {}

$rows = [];
$rows[] = (new ReflectionMethod('Base', 'area'))->__toString();
$rows[] = (new ReflectionMethod('Base', 'none'))->__toString();
$rows[] = (new ReflectionFunction('freeFn'))->__toString();

eval('
$ev = [];
$ev[] = (new ReflectionMethod("Base", "area"))->__toString();
$ev[] = (new ReflectionMethod("Base", "none"))->__toString();
$ev[] = (new ReflectionFunction("freeFn"))->__toString();
');

for ($i = 0; $i < 3; $i++) {
    echo $rows[$i] === $ev[$i] ? "SAME" : "DIFFERS", "\n";
}
```

All three print `DIFFERS`.

### 1. The prototype marker is missing

```
compiled: Method [ <user, prototype Shape> abstract public method area ] {
eval:     Method [ <user> abstract public method area ] {
```

PHP prints `<user, prototype Shape>` for a method that overrides an interface or parent declaration. `eval_reflection_function_method_header` (`crates/elephc-magician/src/interpreter/reflection/formatting.rs:273`) builds the modifier list and the name but has nothing for the prototype; `EvalReflectionFunctionMethodTarget::Method` carries no prototype field.

### 2. A callable with no parameters and no return type gets a parameters block

```
compiled: Method [ <user> public method none ] {
          }
eval:     Method [ <user> public method none ] {
            - Parameters [0] {
            }
          }
```

PHP prints an EMPTY body there, and prints the block as soon as either half has something to say — measured on 8.5.10 with `function f() {}` against `function f(): void {}` and `function f($a) {}`. `eval_reflection_function_method_to_string` (`formatting.rs:249`) always emits the block.

### 3. An array default is dropped

```
compiled: Parameter #0 [ <optional> array $a = [] ]
eval:     Parameter #0 [ <optional> array $a ]
```

[...]

PR #1120, opened 2026-09-20
feat(reflection): render the PHP dump from AOT __toString()

### LINKED ISSUE #1080: AOT ReflectionMethod::__toString() still returns empty string
## Summary

AOT `ReflectionMethod::__toString()` is still unimplemented: it returns `""`. Eval/Magician already formats a PHP-like dump. Follow-up from [#1032](https://github.com/illegalstudio/elephc/pull/1032), which correctly left this out of scope (casing / `getName()` only).

## Current behaviour

| Path | `ReflectionMethod::__toString()` |
|---|---|
| Compiled AOT | returns `""` |
| Inside `eval()` (Magician) | real dump (already on `main` since #451) |

AOT wires `__toString` as a getter of a private `__string` slot that defaults to `""` and is never filled when a `ReflectionMethod` is materialized (`owner_class.rs` / `class_scalar_methods.rs` / `member_object_emit.rs`). There is no `reflection_method_to_string` counterpart to the existing AOT `ReflectionProperty::__toString` formatter.

Eval already handles it via `eval_reflection_function_method_to_string_result()` in Magician (`callable_api.rs` + `formatting.rs`), with tests pinned.

The same empty-string stub pattern also applies to AOT `ReflectionFunction::__toString`.

## Expected

AOT `ReflectionMethod::__toString()` should match PHP (and Magician), similar to how AOT already formats `ReflectionProperty::__toString()`.

## Related

- PR [#1032](https://github.com/illegalstudio/elephc/pull/1032) — declared this out of scope; confirmed still unimplemented on that branch and on `main`
- PR [#451](https://github.com/illegalstudio/elephc/pull/451) — eval formatter already landed
- Open [#893](https://github.com/illegalstudio/elephc/pull/893) (`feat(core): …`) is Core builtins, not this API
- Issue [#571](https://github.com/illegalstudio/elephc/issues/571) — `getName()` casing (addressed by #1032), not `__toString`

## Notes

No dedicated issue existed for this gap before this one.

Guikingone:
Fixes #1080.

`ReflectionMethod::__toString()` and `ReflectionFunction::__toString()` answered the **empty string**
in compiled code. Both declare a private `__string` slot read back through a getter, and no emitter
ever filled it. The eval bridge has formatted these since #451, so the same program answered
differently depending on which side of an `eval()` the reflection ran.

Measuring the family turned up a third hole the issue does not name: **`ReflectionParameter::__toString()`
was wired to the `__name` slot**, so `(string) $parameter` answered `a` where PHP answers
`Parameter #0 [ <required> int $a ]`. A parameter line is what a method dump is made of, so it is
fixed here too.

## What the output is

PHP's, minus one line. PHP prints `  @@ <file> <line> - <line>` under the header, and a compiled
binary cannot honestly answer it: the source it was built from need not exist where the binary runs,
and baking the build machine's path in would be worse than leaving it out.

Everything else matches 8.5.10, measured shape by shape — the modifier order, the
`<user, prototype Interface>` marker on an override, `<required>`/`<optional>`, by-reference and
variadic markers, the return line, and the **empty body** PHP prints for a callable with no
parameters and no return type (it prints the parameters block as soon as either half has something
to say).

## Four corrections to how defaults render

- A **global constant** default carried neither its name nor its value, so `int $n = LIMIT` printed
  nothing where PHP prints `= LIMIT`, and `isDefaultValueAvailable()` answered false while
  `getDefaultValue()` threw. Both now fold, so the reflector answers `true`, `"LIMIT"` and `7` —
  and `N\LIMIT` inside a namespace, which is also what PHP answers.
- An **object** default dropped its arguments: `new Foo(1, 'a')` printed `new Foo`. PHP exports the
  written AST, so the arguments stay and the class name is fully qualified — and `new Foo()` keeps
  its empty parentheses, which means only the arguments actually WRITTEN may be printed, not the
  constructor defaults the metadata backfills for `getDefaultValue()`.
- A **float** outside a narrow decimal range expanded every digit: `1e100` printed as 101 digits
  where PHP prints `1.0E+100`.
- A **union** printed in the order it was DECLARED. PHP renders a union from its type mask, which is
  fixed: class names first in the order written, then `static`, `callable`, `object`, `array`,
  `iterable`, `string`, `int`, `float`, `bool`, `false`, `true`, `null`. So `int|string` comes back
  as `string|int` while `int|float` stays put. The table is derived from measurement, row by row.
  It is shared with `ReflectionProperty::__toString()`, so
  `test_reflection_property_to_string_formats_aot_metadata` — which pinned the declared order —
  now pins PHP's.

## Two things the dump exposed rather than caused

`getDeclaringFunction()` builds its reflector from metadata that deliberately carries **no
parameters**: emitting them would emit a `ReflectionParameter` for each, and each of those emits its
own declaring function, which does not terminate. A dump rendered from that metadata claimed
`Parameters [0]` for a method that has two, so it is rendered from the real member and carried
across as a string, which has no such cycle. `getParameters()` on that reflector is still empty —
#1119, with three ways out.

An internal callable printed `<user>`, contradicting the same object's `isInternal()`. It now prints
`<internal>`; PHP names the extension (`<internal:Core>` for `strlen`, `<internal:standard>` for
`count`), which the metadata reaching the renderer does not carry.

## Tests

`tests/codegen/oop/reflection_to_string.rs` — 15 tests: the repro, modifiers and the prototype
marker, the empty-body rule, every parameter shape, every default shape, the function form, a
parameter on its own, the two emitters agreeing, the eval parity shape, the global and namespaced
constant defaults, object defaults, exponent floats, the declaring-function dump, and the internal
origin tag.

## Reviewed

Kimi K3, GLM 5.3 and DeepSeek. Four major findings reproduced against host PHP and fixed (the
constant-default API contradiction, the `getDeclaringFunction()` dump, the `<user>` tag on internal
callables, and the dropped object-default arguments), plus the exponent floats. Two findings
measured and discharged: mixed keyed/unkeyed array defaults already render correctly, and a
namespaced constant should answer with the resolved name — PHP answers `N\LIMIT` too. One reviewer
correctly caught that the docs claimed the compiled and eval paths agree; they now differ in four
places where the compiled path matches PHP and eval does not, and both the docs and #1117 say so.


Guikingone:
Seven findings, all measured. Four are fixed here; three are filed with the host PHP output they
should match, and measuring two of them turned up things the review did not name.

## Fixed

**Array constants preventing compilation** — confirmed, and worse than described. It is not that
reflection rejects the metadata; the whole program fails to build:

```
$ elephc run repro.php
EIR backend error: unsupported EIR backend feature:
  ReflectionClass constant metadata expression ArrayLiteral([...])
```

`ReflectionConstantValue` has no array variant, and routing `ConstRef` through that fallible
evaluator made the unsupported arm reachable from a parameter default for the first time. The
global-constant arm now answers `Ok(None)` like every other unsupported default here, so the
program builds. The metadata is still short of PHP, and that is filed as #1230 — but the state it
lands in is PHP's own: `isDefaultValueAvailable()` is false and `isDefaultValueConstant()` raises
`ReflectionException: Internal error: Failed to retrieve the default value`, which is byte for byte
what PHP 8.5.10 does for a parameter with no default. Measured on `function noDefault(int $x)`,
both print the same two lines.

**Builtin declaring dumps** — confirmed and fixed. `reflection_builtin_function_signature`, already
consulted two lines below to decide the `<internal>` label, is the signature source:

```
before: Function [ <internal> function strlen ] { - Parameters [0] { } - Return [ int ] }
after:  Function [ <internal> function strlen ] { - Parameters [1] {
          Parameter #0 [ <required> string $string ] } - Return [ int ] }
```

**Non-finite floats** — confirmed and fixed. `INF`, `-INF`, `NAN` now, matching PHP exactly.

**The displaced docblock** — confirmed and fixed; each function has its own again.

## Filed, with the reason

**The `&` marker → #1231, and it is bigger than the marker.** Wiring it showed `returnsReference()`
is a BAKED CONSTANT `false` in the checker, so the API lies about the same declarations the dump
does. Measured:

| | elephc | PHP |
| --- | --- | --- |
| `function &getv` | `false` | `true` |
| `function plain` | `false` | `false` |
| `public function &get` | `false` | `true` |
| `public function plain` | `false` | `false` |

Two of four wrong. The marker and the API need the same absent field, so they are one issue:
shipping the marker while `returnsReference()` still answers `false` would leave two paths
disagreeing about one declaration, which is the disease this PR is treating. The fix mirrors
`is_generator` end to end — a `__returns_reference` slot, the flag on three metadata types, two
emission sites and the checker method — and `is_generator` currently appears in 12 files.

**Origin tags.** Measured grammar is `<origin[, inherits X | overwrites X][, prototype X][, ctor]>`.
`prototype` is already rendered; the rest need the DECLARING class beside the reflected one, and
`reflection_listed_method_to_string` receives only the member — at its `getMethods()` call site the
class in scope is `"ReflectionMethod"`, the reflector, not the reflected class. One correction to
the review: PHP 8.5 emits **no `dtor` tag** — `__destruct` prints a plain `<user>`, measured.

**Constant expressions in object defaults.** Confirmed: PHP prints `new \Foo(Foo::VALUE)` and
elephc prints `new \Foo(7)`, because the arguments are folded to values before the dump and the
written spelling is not carried.

## Found while testing, not a review finding

**#1229: casting `getDeclaringFunction()`'s result to string SEGFAULTS.** This whole file is the
program:

```php
<?php
echo (string) (new ReflectionParameter("strlen", "string"))->getDeclaringFunction(), "\n";
```

Exit 139, no output. It reproduces identically at the merge-base `89a923f4b4`, so it predates this
branch — but it is why the new builtin fixture reads two names before the dump instead of being a
one-liner, and the fixture says so.

## Verification

reflection 307, oop 857, attributes 162, lib 1685 + crates: all 0 failed. Three fixtures added,
one per fix; each fails with its `src` half reverted and nothing else in the file does.

One caveat rather than a clean claim: the workspace `--lib` sweep is flaky here. The first run
failed `elephc-iconv::honors_translit_and_ignore_suffixes`, the second passed that and failed
`elephc-pcntl`, the third was clean. Neither crate is reachable from reflection codegen, and the
iconv test passes on its own (23/23). A different crate failing each run is the parallel-race
shape, not this diff.

Issue #1121, opened 2026-09-20
A tagged nullable int is still misread by json_encode, six backend refusals, and as an array element


Guikingone:
`int|null` is the one union `PhpType::codegen_repr()` does not map to `Mixed`: under the default tagged null representation it becomes `TaggedScalar`, an unboxed two-word `{payload, tag}` pair. #1040 fixed the call ABI, which covered every consumer reached through a `mixed` parameter. Eight consumers are still wrong, in three distinct mechanisms.

## Repro

```php
<?php
function nint(int $i): ?int { return $i >= 0 ? 5 : null; }
$n = nint(1);
echo json_encode($n), "\n";
```

## The table, measured on 8.5.10 against the #1040 branch

Twenty consumers probed with a `?int` holding `5`, one per compilation so each refusal names itself:

| consumer | elephc | php |
| --- | --- | --- |
| `gettype`, `is_int`, `intdiv`, `max`, `sprintf`, `str_repeat`, `count`, `.`, `strval`, `abs`, `array_fill`, `str_pad`, `match` | correct (13) | |
| `var_export` | correct since #1040 | `5` |
| **`json_encode($n)`** | **`null`** — silently wrong | `5` |
| `is_numeric($n)` | refused: *unsupported EIR backend feature: is_numeric for PHP type TaggedScalar* | `true` |
| `in_array($n, [5])` | refused: *loose in_array needle PHP type TaggedScalar for indexed-array element PHP type Int* | `true` |
| `array_sum([$n])` | refused: *array_sum for PHP type Array(TaggedScalar)* | `5` |
| `implode(",", [$n])` | refused: *implode array element PHP type TaggedScalar* | `5` |
| `[$n => 1]` | refused: *hash key PHP type TaggedScalar* | `[5 => 1]` |
| `serialize($n)` | refused: *serialize() of TaggedScalar* | `i:5;` |
| **`var_export([$n], true)`** | **`array (\n  0 => ,\n)`** — empty element | `array (\n  0 => 5,\n)` |

Only `int|null` is affected. `?string`, `?float`, `?bool` and `?array` are already boxed Mixed and all twenty consumers answer correctly for them — measured.

## Three mechanisms, not one

[...]

PR #1122, opened 2026-09-20
fix(codegen): box a tagged nullable int for a mixed parameter

### LINKED ISSUE #1040: var_export() on a nullable declared property segfaults
## Summary

`var_export()` on a **nullable** declared property segfaults. The property does not have to hold `null` — a `?int` holding `5` crashes too. `var_dump()` on the same property is fine, and so is `var_export()` on a non-nullable `int` property.

## Reproduction

```php
<?php
class N { public ?int $n = null; }
$o = new N();
echo var_export($o->n, true), "\n";
```

```
$ elephc -q ve1.php && ./ve1
Segmentation fault: 11
```

PHP prints `NULL`.

It is not about the stored value:

```php
<?php
class N { public ?int $n = 5; }
$o = new N();
echo var_export($o->n, true), "\n";   // PHP: 5 — elephc: Segmentation fault: 11
```

| fixture | result |
|---|---|
| `class N { public ?int $n = null; }` + `var_export($o->n, true)` | **segfault** |
| `class N { public ?int $n = 5; }` + `var_export($o->n, true)` | **segfault** |
| `class N { public int $n = 5; }` + `var_export($o->n, true)` | `5` — correct |
| `class N { public ?int $n = null; }` + `var_dump($o->n)` | correct |
| `class N { public ?int $n = null; }` + `$o->n++; var_dump($o->n)` | `int(1)` — correct |

So the nullable property's storage reads back correctly through every other consumer; only `var_export()` mishandles it. The shape of the difference — a boxed/nullable cell where the non-nullable path has a raw slot — suggests `var_export`'s argument lowering dereferences the cell without unboxing, but I have not confirmed which side.

## Why it is filed separately

Found while fixing #682 (`++`/`--` in expression position). The first fixture has no increment in it at all, so it is not related to that change — it just happened to be the shape a probe fixture used.

A segfault is worse than a wrong answer: there is no diagnostic and no exit status to check.

## Environment

Guikingone:
Fixes #1040.

`var_export()` on a `?int` segfaulted. `emit_box_current_value_as_mixed` matched on the **declared**
type, where `PhpType::Union(_)` means "already a boxed Mixed" — true of every nullable union except
this one. With the default tagged null representation `int|null` is an unboxed two-word
`{payload, tag}` pair, so it reached the `PhpType::Mixed | PhpType::Union(_)` arm before
`PhpType::TaggedScalar` was considered and **nothing was emitted**. The callee read the raw payload
word as a Mixed pointer, and the caller's own `__rt_decref_mixed` ran on the integer.

The match is now on the representation. One line — the `TaggedScalar` arm that boxes the payload
with its dynamic tag was already there, sitting unreachable below the arm that swallowed its only
input.

Instrumentation found it, not reading: a temporary `eprintln!` in
`materialize_direct_call_arg_for_param` printed
`source=Union([Int, Void]) repr=TaggedScalar param=Mixed repr=Mixed`, proving the boxing arm was
reached and the helper emitted nothing.

## The issue's perimeter, corrected on three points

Measured at `89a923f4b4`:

- **not properties** — a nullable-int local and a method's `?int` return crash identically;
  `var_export(nint(1), true)` needs no class at all;
- **not nullable in general** — `?string`, `?float`, `?bool` and `?array` are already boxed Mixed
  and were always correct;
- **not only `var_export`** — of twenty consumers probed against a `?int` holding `5`, thirteen were
  correct, `var_export` segfaulted, `json_encode` answered `null`, and `is_numeric`, `in_array`,
  `array_sum`, `implode`, `serialize` and a hash key were refused by the backend.

This fixes every consumer reached through a call's `mixed` parameter — free function, method,
closure, first-class callable, `call_user_func`, variadic, and a `mixed` property write, all
measured. A by-reference `mixed` parameter still refuses a tagged scalar at the backend, which is
the honest outcome: the writeback would have to unbox back into two words. The remaining eight
consumers are three separate mechanisms, filed as #1121 with the full table.

## Two earlier cuts, both wrong, both caught by measurement

**Boxing at the EIR argument boundary** fixed `var_export` and broke `abs($n)` → `4330504864`. A
builtin's result type comes from its own check hook, which reads the ARGUMENT's type: the checker
typed the call `int` because it saw `int|null`, and a boxed argument made the runtime return a boxed
Mixed the caller read as a raw integer. Narrowing to a DECLARED `mixed` parameter did not help —
`abs`'s registry parameter is declared `mixed` too.

**The same box on the user-call path only** produced correct output and leaked one heap block per
call: `release_owned_call_arg_temporaries_with_signature` decides from the SOURCE type, and
`call_arg_gets_independent_mixed_box` returns false once the source is already Mixed, so nothing
released the fresh box. Fixing the ABI leaves that bookkeeping intact — `allocs=1001 frees=1001`,
`leak summary: clean`, against `live_blocks=200` over 200 iterations for the EIR cut.

## Tests

`tests/codegen/null_sentinel/tagged.rs`, seven added. Three fail with the line reverted (verified by
mutation). The rest are anti-regression pins that pass either way by design: the other nullable
scalars, the builtins that must keep receiving the value **unboxed**, every other `mixed` call shape,
and a generator parameter.

Full codegen suite: 9017 passed, 0 failed.

## Reviewed

Kimi K3, GLM 5.3 and DeepSeek. Three major findings did **not** reproduce and are recorded here
because the measurements are worth having: two reviewers independently reported a leak in
`zval_pack`, reading `ctx.value_php_type(value)` as the declared `Union([Int, Void])` — it is the
EIR value type, already `TaggedScalar`, so `__rt_mixed_free_deep` fires and the heap is
`leak summary: clean` with and without the guard they proposed changing (DeepSeek reached the same
conclusion by reading). GLM also reported generators with a `?int` parameter broken end-to-end;
measured, `gen(?int $n)` yields `5|7|` and `gen2(?int $n, string $tail)` yields `5|'x'|`, both
correct. DeepSeek's two minor findings were right and are fixed: the `Union(_)` half of the first
arm is now unreachable and the comment said otherwise, and the call shapes it listed as untested are
now pinned.


Guikingone:
Half right, and the half that is right is broader than the finding. Measured, then filed as #1232.

The leak is real: a caught exception does cost one heap block here. It is **not** about the box,
and not about tagged nullable ints.

## The control the claim needs

Measured with `--heap-debug` as a slope across 100 and 300 iterations, on this branch:

| call that throws | blocks per caught exception |
| --- | --- |
| `takesNothing()` — no argument at all | **0.00** |
| `takesInt($i + 1)` — `int` to an `int` parameter, no boxing anywhere | **1.00** |
| `takesMixed(nint($i + 1))` — `?int` to a `mixed` parameter, the shape in the finding | **1.00** |

The bare throw is clean, so it is not the exception object. And an `int` argument does no boxing
at all yet leaks exactly the same one block, so it is not the Mixed cell this branch creates.
Whatever a caller owes for an argument is paid on the normal-return edge and not on the unwind
edge, whatever the representation.

## Whether this branch caused it

No. Measured at the merge-base `89a923f4b4` and here:

| | merge-base | this branch |
| --- | --- | --- |
| no argument | 0.00 | 0.00 |
| `int` argument | 1.00 | 1.00 |
| `?int` to a `mixed` parameter | 0.00 | 1.00 |

That last row is the one worth reading carefully. 0.00 at the merge-base is not cleanliness — it
is the defect: NOTHING was emitted for the tagged nullable int, which is exactly why
`var_export()` on a `?int` segfaulted and why this branch exists. Once the box is correctly
created, the argument joins the pattern every other argument already had.

So the branch removes a miscompile, and its victim then exhibits a pre-existing leak. Fixing that
leak means releasing argument temporaries on the landing-pad edge, which is a codegen change
affecting every call that can throw — not something to fold into a one-line boxing fix. Filed as
#1232 with the repro and both measurement tables.

Issue #1125, opened 2026-09-20
Copying a variable holding an eval-declared object gives two scope cells one reference ($x = $y; $y = null; frees it)


Guikingone:
Copying a variable that holds an eval-declared object gives two scope cells one reference. Clearing either one frees the object while the other still names it.

## Repro

```php
<?php
eval('
class OBag { public int $k = 5; }
$y = new OBag();
$x = $y;
$y = null;
echo ($x instanceof OBag) ? "yes" : "no", "|", $x->k, "|";
');
```

```
$ php repro.php
yes|5|

$ elephc -q repro.php && ./repro
no||
```

`$x` still names the object; PHP prints `yes|5|`. elephc has already freed it.

## Mechanism

`EvalExpr::LoadVar` returns the stored handle with no retain
(`crates/elephc-magician/src/interpreter/expressions.rs:173`), and
`EvalStmt::StoreVar` stores what it is handed as `ScopeCellOwnership::Owned`
(`interpreter/statements/dispatch.rs`). So after `$x = $y;` both cells claim `Owned` over the
one reference `new OBag()` produced. `$y = null;` makes `set_scope_cell` release the replaced
cell (`scope.rs` → `owned_cell_except`), the refcount reaches zero, and `$x` dangles.

This is adjacent to #982 but a **different invariant**. #982 was a cell claiming a reference
nobody held — `$a = $this;` over a `Borrowed` cell — and is fixed by recording `Borrowed`
instead, so the alias owns nothing and releasing it is a no-op. Here both cells genuinely
should own a reference, and only one exists. The predicate #982 added
(`value_is_an_unretained_scope_borrow`) deliberately requires a `Borrowed` source and so does
not fire for this.

## Why it was left out of #982

Fixing it means choosing between two designs, and the choice is entangled with a third
pre-existing gap:

1. **The alias retains.** `$x = $y;` takes a reference when the value came from a
   pass-through read of an *Owned* cell. Simple, but nothing gives that reference back: a
   method or function scope is dropped without `drain_owned_cells` (`dynamic_method_execution.rs`
   creates the scope, runs, and drops it; there is no `Drop` impl on `ElephcEvalScope`), so
   every alias inside a function would leak. That is exactly what was measured when the same
   approach was tried for the `Borrowed` case: three blocks per call.
2. **The scope tracks aliasing**, so a released cell only frees when it is the last claim.
   Correct without a drain, but a larger change.

[...]

Issue #1124, opened 2026-09-20
An eval-declared class extending Exception cannot be constructed


Guikingone:
An eval-declared class that `extends Exception` cannot be constructed. It is not a `throw` or `catch` problem — `new` alone is enough.

## Repro

```php
<?php
eval('
class VErr extends Exception {}
$e = new VErr("x");
echo ($e instanceof Exception) ? "yes" : "no", "|", $e->getMessage(), "|";
');
```

```
$ php repro.php
yes|x|

$ elephc -q repro.php && ./repro
Fatal error: eval() runtime failed
```

## Perimeter, measured against PHP 8.5.10

The builtin exception classes work; it is subclassing them from inside `eval()` that does not.

| fragment | PHP | elephc |
| --- | --- | --- |
| `try { throw new Exception("x"); } catch (Exception $e) { … }` | `caught\|ok\|` | `caught\|ok\|` |
| `try { throw new Exception("x"); } catch (Exception) { … }` | `caught\|ok\|` | `caught\|ok\|` |
| `$e = new VErr("x");` where `class VErr extends Exception {}` | `yes\|x\|` | **fatal** |
| `try { throw new VErr("x"); } catch (VErr $e) { … }` | `caught\|ok\|` | **fatal** |
| `try { throw new VErr("x"); } catch (Exception $e) { … }` | `caught\|ok\|` | **fatal** |

So a variable-less `catch`, a `catch` with a variable, and `throw` are all fine on the builtin
class. The three failing rows all contain `extends Exception`, and the third of them never
throws anything — the constructor is where it dies.

Declaring an eval class that extends another **eval-declared** class works, so it is specifically
inheriting from a builtin/AOT exception class. I did not narrow it further than that.

## Why it matters beyond the obvious

`class MyException extends \RuntimeException {}` is how essentially every library and
application declares its errors, so this closes off eval for most real code that raises anything
of its own.

It also hides other defects. While fixing #982 I needed to check that `throw $param;` does not
destroy the caller's object, and the natural repro —

```php
class MyErr extends Exception {}
function rethrow(MyErr $e): void { throw $e; }
```

— fatals here for this reason, not for the ownership reason under test. Rewriting it against the
builtin `Exception` showed the ownership bug was real and let it be fixed. Anyone probing eval
exception behaviour should reach for the builtin class until this is resolved.

## Verified at

`89a923f4b4` and on top of the #982 branch (`60dff73c85`) — unaffected by that work either way.

[...]

Issue #1123, opened 2026-09-20
Writing $this into a property of an eval-declared class stores the handle without retaining it (bad refcount)


Guikingone:
Writing `$this` — or any borrowed scope cell — into a property of an eval-declared class stores the handle without taking a reference. Under `--heap-debug` the program dies with `bad refcount`; without it the refcount is silently one too low.

## Repro

```php
<?php
eval('
class NSelf {
    public ?NSelf $me = null;
    public function __construct() { $this->me = $this; }
    public function k(): int { return 1; }
}
$t = 0;
for ($i = 0; $i < 200; $i++) { $o = new NSelf(); $t = $t + $o->k(); }
echo $t, "\n";
');
```

```
$ php repro.php
200

$ elephc -q repro.php && ./repro
200

$ elephc -q --heap-debug repro.php && ./repro
Fatal error: heap debug detected bad refcount
```

PHP 8.5.10 prints `200`. Without `--heap-debug` elephc also prints `200`, so the defect is silent
in a normal build — the instrumentation is what names it.

## Perimeter

- **eval only.** The same class compiled AOT is correct.
- It is the *self-reference* that matters, not the property write. With the property holding a
  freshly constructed object instead of `$this`, the same program is clean:

  ```php
  public function __construct(bool $seed) { if ($seed) { $this->me = new NSelf(false); } }
  ```

  `allocs=5621 frees=4013 live_blocks=1608` and no fatal.
- No `return` is involved, and no method is called on the result. The constructor alone is enough.

## Mechanism

`$this` and every by-value parameter are bound into the callee's eval scope as
`ScopeCellOwnership::Borrowed` (`interpreter/statements/dynamic_method_execution.rs`,
`interpreter/reference_writeback.rs`): the caller holds the only reference, and
`drain_owned_cells` deliberately skips them at teardown. `EvalExpr::LoadVar` hands that stored
handle straight back with **no retain** (`interpreter/expressions.rs:173`).

So every site that takes ownership of an expression result is taking ownership of a reference
nobody gave it. #982 fixed two of them — `EvalStmt::Return` and `EvalStmt::StoreVar`, both now
routed through `retain_unretained_scope_borrow` in
`crates/elephc-magician/src/interpreter/statements/dispatch.rs`. The property write
(`execute_property_stmt`) is the third and was left alone, because it is reachable without
either of those statements and wanted its own repro.

The fix is likely the same shape: decide on two axes — the expression must be one that hands a
[...]

PR #1151, opened 2026-09-21
fix(eval): stop a borrowed scope cell from being destroyed by the statement that reads it

### LINKED ISSUE #982: A method declared : self or : static and called as a statement (result discarded) is refused by the eval bridge
A method declared `: self` or `: static` and called as a STATEMENT — the fluent-interface idiom used for its side effect — is refused by the eval bridge with `unsupported MethodCall expression`. The same call with its result USED works, and the same discarded call with any other return type works.

### Measured, one case per file

| probe | method signature | call site | result |
|---|---|---|---|
| `d1` | `add(string $v): void` | `$o->add("a");` discarded | IDENTICAL to php |
| `d2` | `add(string $v): string` | `$o->add("a");` discarded | IDENTICAL to php |
| `d3` | `add(string $v): self` | `$o->add("a");` discarded | **`unsupported MethodCall expression`** |
| `t24` | `add(string $v): static` | `$o->add("a");` discarded | **`unsupported MethodCall expression`** |
| `d4` | `add(string $v): static` | `$x = $o->add("a"); $y = $x->add("b");` result used | IDENTICAL to php |

php prints the expected counts in every case. So the discriminator is exactly: **declared return type `self`/`static` AND the result discarded.**

### Why it matters

This is the fluent/builder idiom, and calling one link for its side effect without chaining is completely ordinary. Symfony's container alone has 260 `return $this;` methods across 52 files on the 404 request path; `ContainerBuilder::addObjectResource($this);` is one such discarded call on the boot path.

### Note on an earlier, different claim

A construct-mapping pass described this area as "a discarded call to a method returning `$this` DESTROYS the receiver". Reduced here, it is not a destruction — the receiver is fine — it is a refusal at the call. Worth stating because the two readings point at different code.

Guikingone:
Closes #982.

## The bug

```php
<?php
eval('
class Bag {
    private array $items = [];
    public function add(string $v): self { $this->items[] = $v; return $this; }
}
$o = new Bag();
$o->add("a");
var_dump($o instanceof Bag);
');
```

elephc: `bool(false)`. PHP 8.5.10: `bool(true)`. The receiver is gone.

## The mechanism

`$this` and every by-value parameter are bound into the callee's eval scope as
`ScopeCellOwnership::Borrowed` — the caller holds the only reference, and `drain_owned_cells`
deliberately skips them. `EvalExpr::LoadVar` hands that stored handle straight back with **no
retain**, so every statement that does something ownership-bearing with an expression result
acts on a reference nobody gave it.

Four statements did. They now share one predicate,
`value_is_an_unretained_scope_borrow`, and answer it differently according to what each does
with the value:

| statement | what it does with the value | answer |
| --- | --- | --- |
| `return $this;` | transfers it out of the frame | **retain** |
| `throw $param;` | hands it to the catch site, which owns or releases it | **retain** |
| `$this;` | discards it *by releasing it* | **skip the release** |
| `$a = $this;` | keeps it in the frame under an `Owned` claim | **record `Borrowed`** |

The last row is the one worth reading twice. Retaining there also works — it makes the `Owned`
claim true — but only if something gives that reference back, and a method scope is dropped
without `drain_owned_cells`. Measured, `$a = $this;` with no later reassignment leaked **three
blocks per call** that way (2208 vs 1608 over 200 iterations). Recording the truthful label
costs nothing: `owned_cell` gates both the release of a replaced cell and the drain, so the
alias is free rather than merely balanced. Same loop now holds 1408.

## Two axes, because either alone is wrong

- a **syntactic** filter selects only expressions that hand a subexpression's value back
  unchanged — `LoadVar`, both `Ternary` shapes including `?:`, `NullCoalesce`, `Match` arms.
  It keeps the predicate off `return $this->me;`, where a *retaining* property read can hand
  back the very same handle `$this` holds; a handle-only check would act twice and leak.
- a **handle** check requires the value to be the exact cell a visible borrowed entry holds.
  It keeps the predicate off the branch that did not run: in `$c ? $this : new Bag()` only one
  arm produces the value.

Both halves are pinned by heap fixtures — 1008 live blocks for the fresh-object arm, the
borrowed arm, and no ternary at all.

## The issue's perimeter was wrong on both counts

The issue reads this as "a refusal at the call", discriminated by a declared `self`/`static`
return type, and explicitly dismisses an earlier reading that called it a destruction of the
receiver. Measured, that earlier reading was right:

- the call **succeeds** — a fixture echoing before and after prints everything; only the *next*
  use of the receiver fails;
- the return type is irrelevant — `self`, `static`, a plain class name and **no return type at
  all** fail identically;
- it is broader than reported — a plain function returning its own **parameter** destroys the
  caller's object the same way, and no `return` is needed at all: `$a = $this; $a = 1;` and the
  bare statement `$this;` each destroy the receiver from inside the method;
- it is eval-only — every one of these compiled AOT was already correct.

## Tests

`tests/codegen/eval_borrowed_returns.rs` (7) and `tests/codegen/runtime_gc/eval_borrowed_return.rs`
(6). Proven by mutation at each step: with only the first commit in place exactly the four
fixtures added by the second fail and its own five pass; with the `throw` retain alone reversed
exactly the two releasing catch shapes fail; the never-replaced alias fixture is what rejected
the retain-at-store design.

The GC fixtures assert **differences**, not `leak summary: clean` — an eval fragment does not end
clean today, leaving roughly a dozen blocks per call whatever the callee returns. There is
precedent for that shape in `tests/codegen/misc.rs:423`.

## Reviewed by Kimi K3, GLM 5.3 and DeepSeek

All three found the bare-statement site independently (fixed in `fc447c3931`, found in parallel
here). GLM found the store-site leak, which is what produced the ownership-label redesign. Two
further reported leaks were measured and did **not** reproduce as regressions: a coerced return
leaks one block per call identically with and without the retain (9.0 vs 8.0 blocks/iteration
for an owned local, 8.0 vs 7.0 for a pass-through read — the coercion path's own gap), and the
mixed-arm ternary needs `$b->me === $b`, a refcount cycle whose simpler form already fails
`--heap-debug` on #1123.

## Filed rather than fixed

- **#1123** — property writes, instance and static, store without retaining. Same shape,
  reachable without any of the statements here.
- **#1124** — an eval-declared class extending `Exception` cannot be constructed. This masked
  the `throw` bug entirely through the obvious repro, which is why its fixture uses the builtin
  `Exception`.
- **#1125** — `$x = $y; $y = null;` frees the object while `$x` still names it. A *different*
  invariant — two cells claiming one reference — entangled with the missing frame drain.

Probed and found already correct, so untouched: array literal elements (`[$param]`) and by-value
argument binding both materialize an independent owner.


Guikingone:
Measured; it does not reproduce. The predicate has two axes, not one.

`scope_read_passthrough_names` does collect from every arm — that is the syntactic half, and it is
deliberately over-broad. The second half is a HANDLE check: the value must be the exact cell a
visible `Borrowed` entry holds. The review assumes `$holder->item` hands back the same handle as
`$x` when they name the same object; it does not. A property read returns its own reference, so
the handle check fails and the predicate correctly answers false.

Measured with `--heap-debug` at two scales, with `$this->self = $this` established before each
call so the coincidence the review describes is actually present:

| body | blocks per iteration |
| --- | --- |
| `return $this->self;` (no ternary at all) | 11.00 |
| `return true ? $this->self : $this;` | 15.00 |
| `return false ? $this->self : $this;` | 15.00 |
| `return true ? $this->self : null;` — control, no borrowed arm | **15.00** |

The control is the point. Its inactive arm names nothing borrowed, so the predicate cannot fire on
it, and it costs exactly the same 15.00 as the two shapes the review flags. The +4.00 over the
plain read is the cost of the ternary itself, not of a misclassification. All four print the right
answer.

If the handle check were the only guard and it were coinciding, the `Return` site would retain an
already-owned value and the flagged shapes would sit above the control. They do not.

## Fixed from this review

The `///` comments inside test function bodies, which AGENTS.md:455 forbids — four in this
branch's `eval_borrowed_return.rs`. Greptile flagged one instance of this on #1205; sweeping all
eight of my open branches for the same shape found eleven, and they are being fixed on each.

Guikingone:

Issue #982's repro, including the shapes this branch widened it to:

```
alive|1|alive|alive|alive
```

against the `DESTROYED` the issue reports. Every row matches PHP: the discarded self-return, the
discarded parameter return, the ternary pass-through, and alias-then-reassign.

on it:

```rust
```

Rebasing would layer the narrower mechanism on top of the general one. Nothing here is left to carry over.

The fixtures are in `tests/codegen/runtime_gc/eval_borrowed_return.rs` on this
branch if any of those shapes are worth keeping as coverage against a future regression; they are
not being proposed separately because upstream's own tests should cover the mechanism it owns.

Issue #1204, opened 2026-09-21
test_eval_key_sort_applies_php_sort_flags fails on main: the eval path dies with 'eval() runtime failed'


Guikingone:
`test_eval_key_sort_applies_php_sort_flags` fails on `main`. The compiled fixture dies with `Fatal error: eval() runtime failed`.

## Repro

```
$ RUST_MIN_STACK=33554432 cargo test --release --test codegen_tests test_eval_key_sort_applies_php_sort_flags

test codegen::eval::test_eval_key_sort_applies_php_sort_flags ... FAILED

---- codegen::eval::test_eval_key_sort_applies_php_sort_flags stdout ----
thread '…' panicked at tests/codegen/support/runner.rs:1545:5:
binary exited with status exit status: 1
stdout:
stderr: Fatal error: eval() runtime failed

test result: FAILED. 0 passed; 1 failed; …
```

Reproduced at `c53a9df18d` and at `25280c16bd`, on macOS aarch64. Every other filter I ran on
the same tree is green (`runtime_gc` 320, `calls` 241, `regressions` 523, `closures` 119,
`arrays` 799, `objects` 287, lib 1701 — all 0 failed), so this is the only red test in that set.

## Attribution

The test arrived with #1059 (`feat(arrays): ksort() and krsort() take PHP's sort flags`), merged
2026-09-21. The AOT side of that feature presumably works — it is the `eval()` path that fails,
which matches the shape of the already-filed follow-up #1157/#1161/#1162 family: a feature
landed on the AOT side and the Magician interpreter was not brought along.

I did not investigate further; I hit it while verifying an unrelated change (#992) and confirmed
by reverting that change that it is not mine.

## Why it is worth its own issue

The follow-ups #1161–#1164 describe gaps in the feature. This is different: it is a test in the
repository that does not pass, so anyone running the eval filter gets a red suite and has to
re-derive that it is pre-existing. Either the interpreter needs the flags, or the fixture needs
to be marked for the AOT path only until it does.

Issue #1203, opened 2026-09-21
A class-typed parameter returned to the caller is released as if owned (bad refcount), because the argument load reads as a provisional owner


Guikingone:
A function whose parameter is declared as a CLASS type and which returns that parameter makes the caller release storage it only lent. Under `--heap-debug` the program dies with `bad refcount`; without it the refcount is silently one too low.

The `mixed`-typed form of this is #992 and is fixed. This one survives that fix, and for a different reason: the summary is right, and the consumer never asks it.

## Repro

```php
<?php
class Tag { public string $s = "tag"; }

function pick(int $i): Tag { return new Tag(); }
function ident(Tag $value): Tag {
    if ($value->s === "tag") { return $value; }
    return $value;
}

for ($i = 0; $i < 5; $i++) {
    $tmp = pick($i);
    $value = ident($tmp);
}
echo "ok\n";
```

```
$ php repro.php
ok

$ elephc -q --heap-debug repro.php && ./repro
Fatal error: heap debug detected bad refcount
```

Two call sites are required — a single call is inlined, which removes the boundary. The argument
must reach the callee through a local; nested directly (`ident(pick($i))`) it is an owning
temporary and takes the other path.

## Mechanism

The summary is correct. `apply_expr_effects` only recurses into the object for a
`PropertyAccess`, so the condition invalidates nothing, both branches return the parameter, and
`ReturnArgAlias` for `ident` is `Parameters({0})` — before and after #992, since the body
contains no call.

The loss is at the consumer. `value_is_borrowed_user_call_result`
(`src/ir_lower/context.rs`) requires `!self.value_is_owning_temporary(argument)` *before* it
consults the summary, and `value_is_owned_unboxed_local_load` classifies **any** `LoadLocal` of a
`PhpLocal` slot whose value type is `Array | AssocArray | Object | Iterable` as a provisional
owner. So a `Tag`-typed argument load reads as owning, the borrowed-result test is rejected, the
result falls through to the `Op::Call` arm of the owning match, and the caller releases what the
callee handed back at `+0`.

A `Mixed`-typed load matches neither clause of that predicate and stays `maybe_owned`, which is
exactly why #992's fix works for `mixed` and not here.

The provisional-owner answer exists so a later store can widen the slot, and the *argument's*
own release is repaired at builder finalization — but the *result's* release decision is
consumed eagerly, and `finalize_value_ownership_metadata` only ever promotes `MaybeOwned →
Owned`, never removes.

[...]

PR #1205, opened 2026-09-21
fix(checker): a call does not erase what the summary knows about a by-value argument

### LINKED ISSUE #992: A mixed function returning its parameter from a conditional return corrupts an object payload's refcount
## Summary

A `mixed` function that returns its own `mixed` parameter from a **conditional** `return`, alongside a second `return`, corrupts the refcount of an object payload. The same function with a single trailing `return $param;` is clean, so the defect is in how the return-alias summary treats a conditional parameter return rather than in parameter returns as such.

Found while adding heap-debug coverage for the `(object)` cast (#836 / PR #988), but it reproduces with no cast involved.

## Reproduction

```php
<?php
class Tag { public string $s = "tag"; }

function pick(int $i): mixed {
    if ($i > 100) { return "never"; }
    return new Tag();
}

// Two returns, one of them conditional, both naming the parameter.
function ident(mixed $value): mixed {
    if (is_object($value)) { return $value; }
    return $value;
}

for ($i = 0; $i < 5; $i++) {
    $tmp = pick($i);
    $value = ident($tmp);
}
echo "ok\n";
```

```text
$ target/debug/elephc --heap-debug repro.php && ./repro
```

PHP prints `ok`.

## The clean counterpart

Exactly the same function with one trailing return is fine:

```php
function ident(mixed $value): mixed {
    return $value;
}
```

```text
HEAP DEBUG: allocs=31 frees=31 live_blocks=0 live_bytes=0 peak_live_bytes=240
HEAP DEBUG: leak summary: clean
```

So does converting in place and returning once, which is the shape
`object_cast_prelude`'s dynamic helper is written in to stay out of this:

```php
function ident(mixed $value): mixed {
    if (!is_object($value)) { $value = convert($value); }
    return $value;
}
```

## Notes

- The payload matters: an **array** payload through the same conditional-return
  shape is clean; an **object** payload is what trips the check.
- Without `--heap-debug` the program does not report anything, so this is a
  silent refcount corruption in a release build rather than a diagnosed error.
- Suspected site: the `ReturnArgAlias` summary in `src/types/return_alias.rs`
  and its consumer `release_owned_call_arg_temporaries_with_signature` in
  `src/ir_lower/expr/function_calls.rs` — the caller appears to release the
  argument while the result still aliases it.

## Environment

- Elephc 0.27.0, `main` at `ded8961806`
- macOS aarch64; PHP 8.5 reference

Guikingone:
Closes #992.

## The bug

```php
<?php
function ident(mixed $value): mixed { is_object($value); return $value; }
for ($i = 0; $i < 5; $i++) { $tmp = $i > 100 ? "never" : 7; $value = ident($tmp); }
echo "ok\n";
```

`--heap-debug` → `Fatal error: heap debug detected bad refcount`. PHP 8.5.10 prints `ok`, and so
does elephc without the flag — this is a silent corruption in a release build.

## The mechanism, one word wide in the EIR

The caller lends the argument and releases the result:

```
v18: Heap(Mixed) own=maybe_owned = load_local slot[2]   ; $tmp, BORROWED
v19: Heap(Mixed) own=owned       = call v18 data[0]
release v19                                             ; releases what it never acquired
```

while the callee hands back the parameter's own cell without acquiring it. The clean control
differs by exactly one word — `own=maybe_owned` on the `call`, and no `release`.

That word comes from `value_is_borrowed_user_call_result`, which asks the summary for a PROOF.
`Unknown` is not a proof, so **every provenance `ReturnArgAlias` drops becomes a release of lent
storage**. Precision in that pass is therefore load-bearing, not cosmetic.

## What changed

Four rules, all in `src/types/return_alias.rs`. Nothing in lowering, nothing at the call site.

| rule | why |
| --- | --- |
| a by-value argument is not invalidated | the callee can only rebind what it takes by reference — builtins answer from the registry's `ref_params`, source functions from a new pre-pass |
| a known source function does not wipe every provenance | only `eval`, `extract` and `callback` builtins reach locals the call does not name |
| a by-reference CAPTURE poisons the name for the whole body | the ref cell outlives the closure's creation, and a later assignment must not re-establish a provenance |
| a binary operator answers `None` | concatenation builds a fresh string, arithmetic and comparison produce numbers, `&&`/`||` yield a bool — never an operand |

## The issue's perimeter was wrong on all three of its claims

It reports a **mixed** function returning its parameter from a **conditional** return, corrupting
an **object** payload. Measured at `c53a9df18d`:

- `$x = is_object($value); return $value;` — **no branch at all** — fails;
- a bare `int` payload fails; a declared class type fails;
- what is required is that the body hand the parameter to a CALL and return it, that the argument
  reach the callee through a LOCAL (nested directly it is an owning temporary and takes the other
  path), and that there be TWO call sites — one is inlined.

The smallest failing program has no class, no object, no producer function and no conditional
return.

## The alternative that was tried and rejected

Widening the call site to `may_alias_parameter` fixes every corruption I found — and leaks
wherever the callee does return something fresh, which cost four ownership fixtures, all of the
shape `$value = trim($value); return $value;` where "may alias" is honest. Measured, not
reasoned. The answer belongs in the summary.

## Review

Kimi K3 and GLM 5.3 reviewed the first commit; both found things I had not.

**GLM found a regression this branch introduced.** Once a source call no longer wipes the state,
a name captured by reference survives a call it must not — the closure's ref cell is invalidated
at creation, but a later assignment re-establishes a provenance the closure can still write
through. Measured 800 live blocks against 400 before, i.e. two leaked blocks per call. Fixed by
the capture rule above; back to 400.

**Kimi found a hole, and was right against GLM's reading of it.** Kimi said the `BinaryOp` rule
(which I had written and dropped for failing mutation) closes a real corruption; GLM predicted it
would double-release through compound assignment, which parses to a `BinaryOp` inside an `Assign`.
Both shapes measured:

| shape | without the rule | with |
| --- | --- | --- |
| `if ($c) return $v; return $v . "x";` | **bad refcount** | clean |
| `$value .= "x"; return $value;` | 200 blocks | 200 blocks, unchanged |

The rule is back, and this time it is pinned — it had failed mutation because the fixture was
wrong, not the rule.

**DeepSeek could not be obtained for this issue.** Three attempts: two produced empty reports
through a bug in my own harness (an answer with no tool calls past step 3 was accepted as the
review — now fixed), and the third timed out repeatedly against the API.

## Tests

`tests/codegen/runtime_gc/param_return_alias.rs`, seven fixtures. Six fail with their rule
reversed; `test_by_reference_argument_still_loses_its_provenance` passes either way by design,
pinning that the new precision does not reach the by-reference case.

runtime_gc 320, calls 241, regressions 523, closures 119, arrays 799, objects 287, lib 1701 — all
0 failed. One unrelated test fails on this branch and on `c53a9df18d` alike
(`test_eval_key_sort_applies_php_sort_flags`, filed as #1204, confirmed pre-existing by
reverting).

## Filed rather than fixed

- **#1203** — a class-typed parameter returned to the caller is still released as if owned. The
  summary is right here; the consumer never asks it, because `value_is_owned_unboxed_local_load`
  classifies an object-typed local load as a provisional owner and rejects the borrowed-result
  test first. Located independently by both reviewers.
- **#1204** — the pre-existing `eval` sort-flags test failure.

Two shapes remain undecidable rather than missed, and are documented as such: a by-reference
argument really may have been rebound, and `Parameters({0})` merged with `Unknown` across two
branches really does lose the proven path. Closing those needs a lattice that remembers WHICH
parameters an unknown value might still be.


Guikingone:
All four findings confirmed and fixed. The two P1s are not precision losses — they reproduce
issue #992's own `bad refcount` fatal — and measuring them found a leak this branch had
introduced that the review did not name.

## The two P1s are the original defect, respelled

Issue #992's exact repro, with the by-value call written three other ways:

| spelling | before | PHP |
| --- | --- | --- |
| `is_object($value)` (what this branch fixed) | `ok`, clean | `ok` |
| `peek_free(x: $value)` — named argument | **`Fatal error: heap debug detected bad refcount`** | `ok` |
| `P::peek($value)` — static method | **`bad refcount`** | `ok` |
| `$p->look($value)` — instance method | **`bad refcount`** | `ok` |

So a lost provenance is not the conservative direction: the call site asks for a proof, and
`Unknown` makes the caller release storage it never acquired.

`ParameterModes` now carries the declared parameter names, so a named argument resolves to its
position. Methods go in the same map under `class::method` for a static call written with its
class, and `->method` for an instance call — the latter recorded only when every declaration of
that name agrees, since the receiver's class is not knowable in this syntactic pass.

Writing the fixture turned up a fourth: `(new P())->look($v)` still failed, because the inline
`new` wiped every provenance before the method call was considered — and `new P()` takes no
arguments. Constructors are now resolved like any other call, and a declared class with no
constructor records empty modes instead of reading as an unknown callee.

## A leak this branch introduced

Checking the findings against the merge-base showed a regression in the shape the branch was
written to improve:

```php
function ident(mixed $v): mixed { peek_free($v); return $v; }
$s = str_repeat("y", 4);
$r = ident($s);
```

| | blocks per call | answer |
| --- | --- | --- |
| merge-base `c53a9df18d` | 0.00 | correct |
| this branch | **2.00** | correct |
| with the fix | 0.00 | correct |

The IR shows the result acquired and stored with no matching release, because the summary proves
`ident` returns parameter 0. But `$s` is a bare `Str` slot and the parameter is `mixed`, so the
boundary boxes it: the cell the callee hands back is a fresh allocation, not the argument's
storage. The proof is about the PHP value, not its representation.

`arg_and_result_types_can_alias` admits `(_, Mixed)` for exactly the fresh-box case it was written
to allow (#604) — right for the may-alias path, wrong for deciding nobody releases. The precision
work made that path reachable rather than creating the hole.
`value_is_borrowed_user_call_result` now answers false when the result is boxed and the argument
is not. The existing fixtures missed it because they reach `ident` through a `mixed`-returning
producer, so the argument is already boxed and no fresh cell is made.

## The two P2s

Both correct, both fixed. The docblock did state the ownership consequence backwards — an
undecidable summary makes the caller treat the result as OWNED and release it, not as borrowed
and leak it — and that inversion is what made the boxed-result hole easy to miss. The `///` inside
the test body is an AGENTS.md:455 violation; sweeping my other branches for the same thing found
ten more, being fixed with their own PRs.

## What is left open, with numbers

Fixing the resolvable spellings does not change the rule that makes them fail. Measured after
this commit, every callee this pass CANNOT resolve still dies on the same repro:

| inner call | elephc | PHP |
| --- | --- | --- |
| `$f($value)` — a closure in a local | `bad refcount` | `ok` |
| `$name($value)` — a dynamic function name | `bad refcount` | `ok` |
| `(new $c())->look($value)` | `bad refcount` | `ok` |
| `self::peek($value)` | `bad refcount` | `ok` |
| `parent::look($value)` | `bad refcount` | `ok` |

Filed as #1227 with both candidate designs and the regression counts that ruled each out during
#992. This is also why the by-name method key deliberately stays class-only: widening it adds
disagreements, a disagreement drops the key, and a missing key invalidates everything — which is
the direction that fatals rather than leaks.

runtime_gc 322, calls 241, objects 287, closures 119, regressions 523, oop 843, types 418,
lib 1701: all 0 failed.

Guikingone:

Issue #992's repro, in all four call spellings — the one the issue
reports and the three this branch added:

| inner call | this branch fixed | on `c75c75f15d` |
| --- | --- | --- |
| `is_object($value)` | ok | **ok** |
| `peek_free(x: $value)` — named argument | ok | **ok** |
| `P::peek($value)` — static method | ok | **ok** |
| `$p->look($value)` — instance method | ok | **ok** |

And the five spellings this branch could NOT resolve, filed as #1227, are `ok` as well — closure,
dynamic function name, `new $c()`, `self::`, `parent::`. That issue is closed for the same reason.

The second half of this PR — a proven alias must not suppress the release of a BOXED call result —
is also moot. `$s = str_repeat("y", 4); ident($s)` measured 2.00 blocks per call on this branch
before the fix and 0.00 after; on `c75c75f15d` it is 0.00 with nothing applied.

on it:

```rust
```

Rebasing would layer the narrower mechanism on top of the general one. It would also re-introduce the boxed-result question, which upstream no longer
poses.

`tests/codegen/runtime_gc/param_return_alias.rs` on this branch holds nine
fixtures, including the four-spelling table and the boxed-result slope, if any are worth keeping.

PR #1210, opened 2026-09-21
fix(eval): a property store acquires the borrowed value it keeps

### LINKED ISSUE #1123: Writing $this into a property of an eval-declared class stores the handle without retaining it (bad refcount)

Guikingone:
Closes #1123.

> **Stacked on [#1151](https://github.com/illegalstudio/elephc/pull/1151)** (issue #982), which is
> not merged — the predicate this fix calls was introduced there. Review that one first.

## The bug

```php
<?php
eval('
class NSelf {
    public ?NSelf $me = null;
    public function __construct() { $this->me = $this; }
    public function k(): int { return 1; }
}
for ($i = 0; $i < 200; $i++) { $o = new NSelf(); $o->k(); }
echo "ok";
');
```

`--heap-debug` → `bad refcount`. Without the flag it prints the right answer with the receiver's
refcount one too low.

The overwrite form is louder — it destroys the lent object outright:

```php
function sstat(SBag $b): int { SHold::$slot = $b; SHold::$slot = 1; return 9; }
$o = new SBag(); sstat($o);
var_dump($o instanceof SBag);   // PHP: true. elephc: false.
```

## The rule

`$this` and every by-value parameter are bound `ScopeCellOwnership::Borrowed`, and
`EvalExpr::LoadVar` hands that handle back with no retain. #982 answered this for the statements
that hand such a value **out** of the frame (`return`, `throw` — they retain), **discard** it
(`$this;` — it skips the release), and **keep** it under a label (`$a = $this;` — it records
`Borrowed`).

A property store is the fourth site and needs a fourth answer: a slot has no ownership label to
record and it keeps the value, so it must genuinely **acquire**.

## Where the retain goes

This is the part the review corrected, and it is the interesting half.

A property write has several exits and only ONE consumes the value the transferring way:

| exit | contract |
| --- | --- |
| a `set` hook, `__set` | receives it as a by-value argument, bound `Borrowed`; its own body stores it |
| the generated setter for an AOT-declared slot, the native bridge | takes a reference of its own |
| the stdClass store (eval-declared classes) | keeps the handle verbatim, releases what it replaces |

Only the last needs one. The first commit retained at the statement arm, before that dispatch,
which handed the other two a reference nobody gives back. Measured over 200 stores:

| path | before #1123 | retain at the arm | retain at the exit |
| --- | --- | --- | --- |
| `__set` | 8.0 blocks/store | **13.0** | 8.0 |
| AOT-declared object slot | 6.0 | **8.0** | 6.0 |

`eval_property_set_result` now takes whether the value is a borrowed scope read and retains at
that one exit; the statement arm computes the predicate and hands down the answer rather than
acting on it.

## The perimeter, measured — wider than the issue

The issue reports the instance property write. All of these fail the same way:

| shape | before | after |
| --- | --- | --- |
| `$this->me = $this` in a constructor | `bad refcount` | clean |
| `C::$slot = $param;` then overwrite | `eval() runtime failed` | clean |
| `$h->slot = $param;` then overwrite | `eval() runtime failed` | clean |
| `$h->slot = $this;` from a method | `eval() runtime failed` | clean |
| `$h->$name = $param;` (dynamic name) | destroyed, no fatal | clean |

Five `*Set` arms. Four neighbouring families are deliberately untouched, each probed: compound
assignments (the operator yields a fresh value the slot then owns), the property ARRAY paths
(`$h->slots[] = $param;` and friends already retain before storing — five probes match on both
sides, and a fixture pins them), reference binds, and inc/dec.

The indirect shapes at a store were probed too and are correct: `$h->slot = $c ? $b : $b;`, `??`,
`?:`, a `match` arm, and an alias bound earlier.

## Review

**Kimi K3** found the AOT-slot half by reading the emitted setter's `__rt_incref`. **GLM 5.3**
found the same half independently and added the hook/`__set` half. Both converged on the same
structural answer — retain at the consuming store, not at the arm — which is what this branch
now does.

My own first measurement said neither existed. I had compared against a control that allocates one
more object per iteration, so the surplus and the extra allocation cancelled. Measuring the same
program with the commit reversed is what settled it, and is why both findings are now pinned as
slopes across 100 and 300 iterations rather than against a matched control.

**DeepSeek produced nothing for this issue**: four attempts, all API timeouts. Two of its earlier
empty reports on other issues came from a bug in my own review harness — an answer with no tool
calls past step 3 was accepted as the review — which is fixed; these are not that.

## Tests

`tests/codegen/eval_property_borrow.rs`, six fixtures. Four fail with their commit reversed; the
array-paths fixture passes either way by design, pinning what must not change.

The self-reference fixture asserts the program survives instrumentation rather than ending clean:
`$this->me = $this` is a refcount cycle, so a clean heap is not the bar — what must not happen is
`bad refcount`.

runtime_gc 307, eval 1297, calls 238, regressions 520, closures 119, arrays 785, objects 285, lib
1685, magician 1199 — all 0 failed.


Guikingone:
Both findings measured. One does not reproduce, the other is real but sits in a different place
than the review puts it — and measuring it caught a false claim in my own commit message.

## "AOT static stores double-retain"

Measured on four shapes with `--heap-debug` at two scales, comparing the branch as written
against the same programs with the statement arm's retain removed:

| shape | arm retains (branch) | retain removed |
| --- | --- | --- |
| eval-declared static, fresh store each iteration | 5.00 blocks/store | 0.00, and latently over-released |
| eval-declared static, slot overwritten in scope | 5.00, correct | **fatal: `eval() runtime failed`** |
| AOT static through the native bridge | 7.00 | 7.00 |
| a write that throws (private static from outside) | 17.00 | 17.00 |

The retain is not surplus — it is load-bearing. `context.set_static_property` stores the handle
verbatim and releases whatever it replaces, so a borrowed value must be acquired before it lands
there; removing the retain makes the second shape fail outright. And the AOT bridge path is
unchanged by it in both directions, so nothing is retained twice there.

I did restructure the static path anyway, because the shape the review points at is a real
inconsistency even though it is currently unobservable: `eval_static_property_set_result` now
takes whether the value is a borrowed scope read and acquires it at the one exit that transfers,
exactly as `eval_property_set_result` does. Measured neutral on all four shapes — it states the
contract rather than fixing a defect, and the docblock says so.

## What the measurement did find

Two things this branch did not cause.

**The AOT bridge path leaks 7 blocks per store**, against 5.00 for the identical program with the
holder declared inside the eval. Both programs allocate the same objects per iteration; the only
difference is which store the write lands on. Filed as #1228 with the two programs.

**The previous commit's message is wrong.** It says "an AOT static is not reachable from eval
at all today (`Access to undeclared static property`)". It is reachable, and all three spellings
match host PHP 8.5.10:

```php
class Holder { public static ?object $slot = null; public static string $name = "start"; }
eval('Holder::$name = "written-by-eval";');          // lands
eval('$t = new Thing(); Holder::$slot = $t;');       // lands
function store(object $o): void { eval('Holder::$slot = $o;'); }  // lands
```

That claim is corrected in the follow-up commit.

## "Inactive branch misclassifies ownership"

Does not reproduce. The predicate has two axes, not one: the syntactic collection over every arm
AND a handle check that the value IS the exact cell a visible `Borrowed` entry holds. The review
assumes `$this->self` hands back the same handle as `$this`; it does not — the property read
returns its own reference, so the handle check fails and the predicate correctly answers false.

Measured, with `$this->self = $this` established before each call:

| body | blocks per iteration |
| --- | --- |
| `return $this->self;` (no ternary) | 11.00 |
| `return true ? $this->self : $this;` | 15.00 |
| `return false ? $this->self : $this;` | 15.00 |
| `return true ? $this->self : null;` (control: no borrowed arm) | **15.00** |

The control is the point: its inactive arm names nothing borrowed, so the predicate cannot fire,
and it costs exactly the same 15.00. The +4 over the plain read is the ternary itself. All four
print the right answer.

## Verification

runtime_gc 307, property 550, static_property 86, eval_property 9, eval_borrowed 13,
eval_class 6, magician lib 1199 — all 0 failed. The full `eval` sweep is not in that list on
purpose: it is 1299 tests running at roughly four a minute here, which is hours for a change
confined to eval's property and static-property write paths. runtime_gc and property cover those
between them; the four filters take the rest by name. Said plainly rather than quoting a suite
that did not finish.

Guikingone:

Issue #1123's repro, instance and static:

```
4|9|alive|4|7|alive|4
```

against the `DESTROYED` the issue reports — the borrowed value survives the slot being overwritten
in both. Matches PHP exactly.

The separate AOT-static accounting gap this branch measured (#1228) is closed too: the AOT path
went from 7.00 blocks per store to 5.00, level with the eval-declared path.

on it:

```rust
```

Rebasing would layer the narrower mechanism on top of the general one. The static-path restructuring this branch added was measured behaviour-neutral
on all four reachable shapes, so there is nothing in it that upstream lacks.

`tests/codegen/eval_property_borrow.rs` holds six fixtures, including the `__set`
and AOT-slot slopes, if any are worth keeping.

PR #1217, opened 2026-09-21
fix(reflection): a union type object prints its members the way PHP does

### LINKED ISSUE #1118: ReflectionUnionType::__toString() keeps the declared order, drops object, and answers empty for a union with false
`ReflectionUnionType::__toString()` answers with the DECLARED member order, drops `object` entirely, and answers an empty string for a union holding `false`. PHP renders a union from its type mask, which has a fixed order.

## Repro

```php
<?php
class C {}
interface I {}

class U {
    public function a(int|string $v) {}
    public function d(bool|array|string $v) {}
    public function f(C|int|string $v) {}
    public function g(int|C|I $v) {}
    public function h(false|int $v) {}
    public function i(array|object|string $v) {}
}

echo "a: ", (string) (new ReflectionMethod('U', 'a'))->getParameters()[0]->getType(), "\n";
echo "d: ", (string) (new ReflectionMethod('U', 'd'))->getParameters()[0]->getType(), "\n";
echo "f: ", (string) (new ReflectionMethod('U', 'f'))->getParameters()[0]->getType(), "\n";
echo "g: ", (string) (new ReflectionMethod('U', 'g'))->getParameters()[0]->getType(), "\n";
echo "h: ", (string) (new ReflectionMethod('U', 'h'))->getParameters()[0]->getType(), "\n";
echo "i: ", (string) (new ReflectionMethod('U', 'i'))->getParameters()[0]->getType(), "\n";
```

| declared | elephc | php 8.5.10 |
| --- | --- | --- |
| `int\|string` | `int\|string` | `string\|int` |
| `bool\|array\|string` | `bool\|array\|string` | `array\|string\|bool` |
| `C\|int\|string` | `C\|int\|string` | `C\|string\|int` |
| `int\|C\|I` | `int\|C\|I` | `C\|I\|int` |
| `false\|int` | **`` (empty)** | `int\|false` |
| `array\|object\|string` | **`array\|\|string`** | `object\|array\|string` |

Two of these are worse than an ordering difference: `false` makes the whole string empty, and `object` renders as nothing, leaving a bare `||`.

## Cause

This is the `ReflectionUnionType` OBJECT's own `__toString`, which is a different surface from the one `ReflectionMethod::__toString()` and `ReflectionProperty::__toString()` use. #1080 gave the latter PHP's order through `reflection_union_member_rank` (`src/codegen/lower_inst/objects/reflection/property_members.rs`), measured row by row:

```
class names (in declared order) < static < callable < object < array < iterable
    < string < int < float < bool < false < true < null
```

The type object's own rendering was not touched, so the two disagree on the same program:

```php
$p = (new ReflectionMethod('U', 'a'))->getParameters()[0];
echo (string) $p;             // Parameter #0 [ <required> string|int $v ]   <- PHP's order
echo (string) $p->getType();  // int|string                                   <- declared order
```

## Expected

`ReflectionUnionType::__toString()` should use the same rank table, and should render `object` and `false` as members rather than dropping them.

Measured on `89a923f4b4` plus the #1080 branch, macOS aarch64.

Guikingone:
Closes #1118.

> **Stacked on [#1120](https://github.com/illegalstudio/elephc/pull/1120)** (issue #1080), which
> is not merged — the rank table this reuses was introduced there. Review that one first.

## The bug

Same program, two answers:

```php
echo (string) $p;             // Parameter #0 [ <required> string|int $v ]   <- PHP's order
echo (string) $p->getType();  // int|string                                   <- declared order
```

| declared | elephc before | php 8.5.10 |
| --- | --- | --- |
| `int\|string` | `int\|string` | `string\|int` |
| `false\|int` | **`` (empty)** | `int\|false` |
| `array\|object\|string` | **`array\|\|string`** | `object\|array\|string` |

## The three causes

All in the `PhpType` mapper. Its `TypeExpr` twin already handled two of them, which is why only
parameters and inferred types were wrong:

1. members were never sorted;
2. `PhpType::False` hit `_ => None`, so `collect::<Option<Vec<_>>>()?` poisoned the whole union —
   that is why `false|int` rendered as the EMPTY string rather than as a union missing a member;
3. a bare `object` hint is `PhpType::Object(String::new())`, and the empty name was copied
   verbatim, leaving a bare `||`.

## One table, both engines

The first commit sorted the compiled path only. All three reviewers independently found the
consequence: the eval bridge rebuilds its own metadata, so the fix moved the disagreement one
layer out instead of closing it.

```
aot-prop:   string|int      eval-prop:   int|string
```

`reflection_union_member_rank` now lives in `elephc-builtin-contract`, which the compiler and the
interpreter both already depend on, and both sort with it. Sharing rather than copying is the
point — a duplicated table is how two surfaces drift apart again.

The sort is stable, which keeps class names in their declared order among themselves (they all
share rank 0), and it runs before the `types.len() == 1` branch, so `int|null` still unwraps to a
nullable named type rather than becoming a one-member union.

`getTypes()` inherits the order, which is correct: measured, PHP hands members back in mask order
there too (`getTypes=string,int` for `int|string`).

## Re-based expectations

Twelve in total — five on the compiled side, seven on the eval side. Every one is the same
`int|string` → `string|int` mask order.

**One correction to my own first commit message.** It claimed every re-based value was read back
from PHP. That is true of the ORDER but not of the `nullableUnion` row's member COUNT:

```
php> function f(int|string|null $x) {}
members: string(builtin=1,null=0), int(builtin=1,null=0), null(builtin=1,null=1),
count:   3
```

PHP reports three members for a nullable union, including `null`; elephc reports two and carries
nullability on `allowsNull()`. That divergence predates this work (`null` is filtered into
`allows_null` when the metadata is built) and is filed separately — but the row was re-based on
elephc's count, and saying otherwise was wrong. GLM 5.3 caught it.

## Review

Three models, all three converging on the eval divergence.

- **Kimi K3** located the eval builder and listed the four eval rows a future fix would need to
  re-base.
- **GLM 5.3** found the same, plus the `nullableUnion` count above, plus a fourth gap filed below.
  One of its predictions did not hold: that PHP's mask prints `float` before `int`. Measured,
  `C3|float|int` renders `C3|int|float`, so the table stands — GLM flagged that row at low
  confidence and said as much.
- **DeepSeek** re-derived all five compiled-side expectations from the rank table rather than
  trusting the diff, and observed they match the *pre-existing* `__toString()` renderer — so they
  are not "the new implementation's answer".

## Filed rather than fixed

Each measured, none introduced here:

| shape | elephc | php 8.5.10 |
| --- | --- | --- |
| `true\|int` | `int\|bool` | `int\|true` |
| `iterable\|int` | `iterable\|int` | `Traversable\|array\|int` |
| `getTypes()` on `int\|string\|null` | 2 members | 3 members, incl. `null` |
| `(A&B)\|int` | renders as absent | `(A&B)\|int` |

The first two are member NAMING rather than order, and both affect the dump and the type object
alike, so those two surfaces still agree. The last is the same poisoning mechanism this branch
fixed for `false`.

## Tests

`tests/codegen/oop/reflection_union_type.rs`, four fixtures. Three fail with the first commit
reversed; the fourth pins that a union reducing to one member after `null` is removed still
answers as a nullable named type.

reflection 308, oop 858, types 409, eval 1281, regressions 520, lib 1685 — all 0 failed.


Guikingone:
Both correct.

**The obsolete limitation.** The bullet said the eval bridge "keeps a union in its declared
order (#1117)" and counted three remaining differences from the compiled path. This branch's
second commit is what removed that one: both paths now sort with the shared `union_member_rank`
table, so the sentence contradicted the behaviour documented immediately below it. Fixed, and the
count corrected to two.

**Builtin signatures losing parameters.** Confirmed, and it is the same finding Greptile raised
on #1120, which this branch is stacked on. `function_by_name` only knows generated functions and
closures, so `(new ReflectionParameter('strlen', 'string'))->getDeclaringFunction()->__toString()`
fell back to the deliberately parameterless metadata and claimed `Parameters [0]` for a function
that has one. `reflection_builtin_function_signature` — already consulted two lines below to
decide the `<internal>` label — is the signature source for exactly those names. Fixed on #1120
and inherited here by rebase, so it is reviewed once rather than twice.

PR #1223, opened 2026-09-21
fix(types): a class name is a valid subject for is_a() and is_subclass_of()

Guikingone:
Refs #1113 — the first of the two pieces that issue scopes.

## The bug

```php
echo is_subclass_of("Derived", "Base") ? "y" : "n";   // elephc: n. PHP: y.
echo is_subclass_of("Impl", "I")       ? "y" : "n";   // elephc: n. PHP: y.
echo is_a("Derived", "Base", true)     ? "y" : "n";   // elephc: n. PHP: y.
```

Silent, no diagnostic. `static_relation_holds` opened with

```rust
let PhpType::Object(object_class) = ctx.value_php_type(object)? else {
    return Ok(false);
};
```

so a string subject returned before the parent-chain and interface walks below it were ever
reached. Those walks were already correct — they need a class name and do not care which operand
shape it arrived in.

## What it took to be right

Five things, each measured against host PHP 8.5.10. Three I found, two the review did.

| | |
| --- | --- |
| **`$allow_string` defaults differ** | `is_subclass_of` accepts a name unless told otherwise, `is_a` only when told to: `is_a("D","B")` is false, `is_a("D","B",true)` is true |
| **An interface name is a legal subject** | reachable only through the string form — there is no instance of an interface to pass — and its parents live in `interface_infos` |
| **`is_a` counts the subject itself** | `is_a("I","I",true)` is true where `is_subclass_of("I","I")` is false |
| **A RUNTIME flag must not be guessed** | the result is exactly `$allow_string && relation`, so when the relation holds the flag IS the answer |
| **An unknown name claims nothing, even against itself** | PHP looks the subject up first, so `is_a("Ghost","Ghost",true)` is false |

## Review

All three models reported, and two of the five rows above come from them.

**The regression.** My first commit substituted the builtin's default when the flag was not a
literal, with a comment calling that "the conservative answer". For `is_subclass_of` it is the
*permissive* one, and it turned a previously-correct `false` into a wrong `true`:

```php
$off = /* false at runtime */;
is_subclass_of("Derived", "Base", $off)   // was: y. PHP: n.
```

Found by **DeepSeek** and **Kimi K3** independently. Neither guessing nor refusing was needed —
the flag governs only a string subject, so a runtime branch gives the exact answer, and
`is_a("Derived","Base",$on)` becomes right too, which it was not before this branch existed.

**The self-check.** The interface branch compared subject to target before establishing the
subject exists, contradicting this branch's own stated invariant. Found by **Kimi K3** and **GLM
5.3**.

**The duplication.** `interface_extends` re-implemented `collect_interface_parents` from the
sibling module. **GLM** called it "exactly the disagreement you avoided in attack point 4" — two
walks over the same tables are how `is_subclass_of("J","I")` and `class_implements("J")` come to
disagree. Using the shared walk removes the copy, and its `lookup_interface` gate closed the
self-check hole for free.

## What is deliberately still `false`

A NON-LITERAL name (`$n = "Derived"; is_subclass_of($n, "Base")`). That is the second piece of
#1113 and needs a name-keyed table the emitted program can consult — it is also what
`ReflectionAttribute::IS_INSTANCEOF` actually needs, since `$attribute->getName()` is never a
literal. The issue stays open for it.

## Filed rather than fixed

`is_subclass_of("Suit", "UnitEnum")` for an enum. The cause is wider than this builtin, and wider
than I first measured: enums take their `implements` clause verbatim with no transitive closure,
while classes get one. So `is_subclass_of("Suit", "Colorful")` is wrong too when `HasColor extends
Colorful` — fixing only the implicit `UnitEnum`/`BackedEnum` set would not be enough. GLM found
that half. Special-casing it here would make `is_subclass_of` and `class_implements` disagree,
which is the disease this PR is treating.

## Tests

`tests/codegen/oop/class_relation_names.rs`, six fixtures: the issue's rows, both
`$allow_string` defaults with explicit overrides, the interface-subject walk, name folding and
unknown names, the runtime flag, and the undeclared-name self-check. All six fail with the first
commit reversed.

oop 849, types 424, eval 1282, regressions 523, callables 598, lib 1703 — all 0 failed.


Guikingone:
Confirmed, and wider than reported. Fixed in the follow-up commit.

The branch's own comment on that line said why it should not happen:

```rust
// A flag that is neither a literal nor a `bool` would need PHP's truthiness conversion; the
// declared signature is `bool $allow_string`, so this is the shape the checker does not produce.
```

It does produce it. Measured against host PHP 8.5.10, three scalar types reach that branch and all
three were wrong:

| flag | elephc before | PHP |
| --- | --- | --- |
| `$flag = 1` | `n` | `y` |
| `$flag = "1"` | `n` | `y` |
| `$flag = 1.0` | `n` | `y` |

The review named the `int` case; the `string` and `float` cases came out of measuring it, along
with `is_a` and the interface-subject spellings, which were wrong the same way.

The relation already holds where that branch runs, so the answer is exactly the flag's PHP
truthiness. Rather than a nonzero test, `lower_is_truthy`'s body is split into
`emit_value_truthiness` — taking a bare `ValueId` — and reused. That is load-bearing rather than
tidy: PHP's string rule makes `"0"` false while `"0.0"` and `" "` are true, and a nonzero test
gets all three wrong. All seven truthiness corners now match:

| flag | elephc | PHP |
| --- | --- | --- |
| `"0"` | `n` | `n` |
| `"0.0"` | `y` | `y` |
| `" "` | `y` | `y` |
| `""` | `n` | `n` |
| `-1` | `y` | `y` |
| `0.0` | `n` | `n` |
| `-0.5` | `y` | `y` |

Two fixtures added, one per table. Both fail with the fix reversed and nothing else in the file
does.

nahime0:
Follow-up from review of this PR:

- Opened #1277 — `docs/php/types.md` says `class_implements()` on an enum is empty; with a direct `implements` clause it returns that interface and still omits transitive parents and implicit `UnitEnum`/`BackedEnum` (same gap as #1224).
- Non-literal subjects remain #1113 (second half of that issue).
- Enum `UnitEnum`/`BackedEnum` and transitive `implements` remain #1224.

No other leftovers noted for this change.

PR #1234, opened 2026-09-22
fix(codegen): a plain local load is not a caller temporary at a call boundary

### LINKED ISSUE #1203: A class-typed parameter returned to the caller is released as if owned (bad refcount), because the argument load reads as a provisional owner

Guikingone:
Fixes #1203.

## The bug

```php
class Tag { public string $s = "tag"; }

function pick(int $i): Tag { return new Tag(); }
function ident(Tag $value): Tag {
    if ($value->s === "tag") { return $value; }
    return $value;
}

for ($i = 0; $i < 5; $i++) {
    $tmp = pick($i);
    $value = ident($tmp);
}
echo "ok\n";
```

```
$ php repro.php
ok

$ elephc --heap-debug repro.php && ./repro
```

Without `--heap-debug` the refcount is silently one too low.

## The summary was right; the consumer never asked it

`value_is_borrowed_user_call_result` refuses when the argument is an owning temporary, and it
must: an operand the caller still owes a release on cannot also be the thing nobody releases
(#486). But `value_is_owned_unboxed_local_load` answers **true** for any `LoadLocal` of a
`PhpLocal` slot holding an object, array, assoc array or iterable.

That answer is **provisional**. It exists so a later store can widen the slot, and the argument's
own release is repaired at builder finalization — but the *result's* release decision is consumed
eagerly, and `finalize_value_ownership_metadata` only ever promotes `MaybeOwned → Owned`, never
removes. The caller owes nothing on the load itself: the local's lifetime covers it. So for this
question it is the wrong answer, and taking it rejected the borrowed-result test and released what
the callee handed back at `+0`.

A `mixed`-typed load matches neither clause of that predicate and stays `maybe_owned`, which is
exactly why the same shape was already correct for `mixed` (#992) and not for a class type.

## Perimeter, measured rather than assumed

The predicate covers `Array`, `AssocArray`, `Object` and `Iterable`, so all four looked equally
exposed. They are not:

| shape | with the fix reverted |
| --- | --- |
| `Tag` (object) | **`bad refcount`** |
| `array` | clean |
| assoc `array` | clean |

Only the object form was observably broken. The array fixture is kept as a guard that the fix did
not disturb the shapes that were already right, and its docblock says so rather than implying it
has teeth it does not.

## Tests

`tests/codegen/runtime_gc/class_param_return.rs`, three fixtures: the issue's repro, the array
shapes, and a nested owning temporary (`ident(pick($i))`) that must keep refusing — that argument
genuinely is a temporary the caller owes a release on, and treating the result as borrowed there
would leave nobody to release it. Only the first fails with the commit reversed.

runtime_gc 939, calls 256, objects 458, regressions 590, lib 2319 + crates: 0 failed from this
change.

Three failures in that sweep are not from this branch, each checked directly on plain
`origin/main` at `c75c75f15d` and failing there too:
`test_core_eval_callable_and_nonempty_metadata_results_balance_owners`,
`test_eval_first_class_callables_are_php_closure_objects`, and
`eval_curlfile_and_curl_file_create_construct_working_objects`.

## Credit

Located independently by Kimi K3 and GLM 5.3 while reviewing #992; both named the same predicate.


nahime0 (approved):
lgtm

PR #1235, opened 2026-09-22
fix(json): a tagged nullable int encodes its value, not null

Guikingone:
Refs #1121, the first of the three mechanisms that issue separates — and the one that is **silently wrong** rather than loudly refused.

## The bug

    function nint(int $i): ?int { return $i >= 0 ? $i : null; }
    echo json_encode(nint(5));    // elephc: null. PHP: 5.

No diagnostic, for every value.

`emit_json_encode_loaded_value` dispatches on the argument's `codegen_repr()`,
and `int|null` is the one union that repr does NOT map to `Mixed`: under the
default tagged null representation it stays an unboxed two-word `{payload, tag}`
pair. So it fell past every named arm to the `_` fallback, which calls
`__rt_json_encode_null`.

A `TaggedScalar` is an int or a null and nothing else, so the tag decides between
the two encoders that already exist. Nothing is boxed, deliberately: boxing would
allocate on every call, and it changes what the runtime hands back — the failure
#1121 records for `abs`, whose result type comes from a hook that reads the
argument's type. That problem does not have to be created here.

Measured against host PHP 8.5.10, both polarities and zero:

    5 -> 5 | 0 -> 0 | null -> null | (null ?? -7) -> -7

`0` is in there because a branch that tested the payload instead of the tag would
answer `null` for it. The value also survives two encodings and a later read
(`42|42|42`), which pins that reading the payload in place does not consume it,
and `--heap-debug` reports `leak summary: clean`.

Two fixtures in `tests/codegen/null_sentinel/tagged.rs`, one per property. Both
fail with this commit reversed and nothing else in the file does.

json 450, null_sentinel 36, nullable 100, lib 2319 + crates: 0 failed.
`elephc-pcntl` fails only in the parallel workspace sweep and passes 26/26 when
its crate runs single-threaded, which is its known race and not this change.

The other two mechanisms the issue names — the six backend refusals, and the
array-element form where the elements stay `TaggedScalar` behind a boxed array —
are untouched and the issue stays open for them.


nahime0 (approved):
lgtm

Issue #1244, opened 2026-09-22
throw refuses an eval-declared class instance at compile time, even throw new EvalClass()


Guikingone:
`throw` refuses an instance of an eval-declared class at compile time: `Type error: throw requires an object value`. It refuses even `throw new EvalDeclaredClass("boom")`, where the operand is a literal `new` of a class the program itself just declared. `new` on its own compiles — assigning the instance to a local and calling methods on it is fine. Only `throw` refuses it.

`class MyException extends \RuntimeException {}` declared in eval and thrown from compiled code is the shape this closes off, so it blocks the pattern #1124 was about from the AOT side even now that #1124's declaration defect is fixed.

Found while fixing #1124, and independent of it: the refusal is a checker diagnostic, and it is unchanged by that fix.

## Repro

Three files, because a compile-time refusal takes the whole file with it.

`t_aot.php` — control, AOT-declared subclass:

```php
<?php
class TA extends Exception {}
$e = new TA("boom");
try { throw $e; } catch (Exception $c) { echo "caught|", $c->getMessage(), "|"; }
```

`t_eval.php` — eval-declared subclass, thrown from AOT through a local:

```php
<?php
eval('class TB extends Exception {}');
$e = new TB("boom");
try { throw $e; } catch (Exception $c) { echo "caught|", $c->getMessage(), "|"; }
```

`t_direct.php` — eval-declared subclass, thrown directly with no local in between:

```php
<?php
eval('class TD extends Exception {}');
try { throw new TD("boom"); } catch (Exception $c) { echo "caught|", $c->getMessage(), "|"; }
```

## Measured

Host PHP 8.5.10, elephc at `c75c75f15d` plus the #1124 fix.

| file | elephc | PHP |
|---|---|---|
| `t_aot.php` | `caught\|boom\|` | `caught\|boom\|` |
| `t_eval.php` | `error[5:7]: Type error: throw requires an object value` | `caught\|boom\|` |
| `t_direct.php` | `error[5:5]: Type error: throw requires an object value` | `caught\|boom\|` |

Throwing the same class from **inside** eval works — issue #1124's own table rows for
`try { throw new VErr("x"); } catch (VErr $e)` and `catch (Exception $e)` both pass now. So the
runtime handles these objects correctly; it is the compiled-side checker that refuses them.

## Note on the right shape of the fix

PHP does not require a statically proven object here: `throw $x` on a non-Throwable is a runtime
[...]

Issue #1243, opened 2026-09-22
__get is not dispatched across the eval boundary: eval reads give "", AOT reads of an eval-written property warn


Guikingone:
`__get` is not dispatched across the eval boundary. The `__set` half works — the value reaches the backing property — but reading the same magic property back gives the wrong answer in two different ways depending on which side reads it:

- **eval reads** a magic property: the result is the empty string instead of `__get`'s return value. It does not matter which side wrote it.
- **AOT reads** a magic property that eval wrote: `Warning: Undefined property` instead of a `__get` call.

Only the all-AOT path is correct.

Found while fixing #1124. It is unrelated to that defect: the matrix below is byte-identical with #1124's fix reverted.

## Repro

```php
<?php
class MagicSink {
    public array $seen = [];
    public function __set(string $name, $value): void { $this->seen[$name] = $value; }
    public function __get(string $name) { return $this->seen[$name] ?? "absent"; }
}

$mode = $argv[1] ?? "evalset";

if ($mode === "evalset") {            // write from eval, read from AOT
    $o = new MagicSink();
    eval('$o->colour = "red";');
    echo "count=", count($o->seen), "|aot_read=", $o->colour, "|direct=", $o->seen["colour"] ?? "missing", "|";
} elseif ($mode === "aotset") {       // control: write from AOT, read from AOT
    $o = new MagicSink();
    $o->colour = "red";
    echo "count=", count($o->seen), "|aot_read=", $o->colour, "|direct=", $o->seen["colour"] ?? "missing", "|";
} elseif ($mode === "evalboth") {     // write and read both from eval
    $o = new MagicSink();
    eval('$o->colour = "red"; echo "count=", count($o->seen), "|eval_read=", $o->colour, "|";');
} elseif ($mode === "evalread") {     // write from AOT, read from eval
    $o = new MagicSink();
    $o->colour = "red";
    eval('echo "count=", count($o->seen), "|eval_read=", $o->colour, "|";');
}
echo "\n";
```

## Measured

Host PHP 8.5.10, elephc at `c75c75f15d` plus the #1124 fix. Reverting that fix changes nothing
in this table, so the two defects are independent.

| mode | elephc | PHP |
|---|---|---|
| `evalset` | `count=1\|aot_read=Warning: Undefined property: MagicSink::$colour\|direct=red\|` | `count=1\|aot_read=red\|direct=red\|` |
[...]

Issue #1241, opened 2026-09-22
Exception's seven final methods are declared non-final: an override PHP refuses is accepted


Guikingone:
PHP declares seven of `Exception`'s methods `final`, so a subclass that tries to override one is a fatal error at declaration time. elephc declares all nine non-final, so it accepts the override and `ReflectionMethod::isFinal()` answers `false` where PHP answers `true`.

The declarations are built by `concrete_throwable_method` in `src/types/checker/builtin_types/exception.rs`, which hardcodes `is_final: false` for every method it builds. The same list backs `Error` and, through inheritance, `RuntimeException`, `TypeError`, `ValueError` and the rest.

Found while fixing #1124.

## Repro

```php
<?php
eval('class FinalOverride extends Exception {
    public function getMessage(): string { return "over"; }
}
echo "declared|";');

eval('$r = new ReflectionMethod("Exception", "getMessage");
echo "getMessage final=", $r->isFinal() ? "y" : "n", "|";');
echo "\n";
```

## Measured

Host PHP 8.5.10, elephc at `c75c75f15d`.

| | elephc | PHP |
|---|---|---|
| declaring the override | `declared\|` | `Fatal error: Cannot override final method Exception::getMessage()` |

PHP's own answer for the whole surface, for the fix to match:

| method | PHP `isFinal()` |
|---|---|
| `getMessage` | true |
| `getCode` | true |
| `getFile` | true |
| `getLine` | true |
| `getTrace` | true |
| `getTraceAsString` | true |
| `getPrevious` | true |
| `__toString` | false |
| `__construct` | false |

`__toString` and `__construct` must stay overridable — that is how user exception classes customise their rendering.

Note this is a *tightening*: setting the flag makes elephc refuse programs it currently compiles. Any fixture in the tree that overrides one of the seven will need updating, which is also the measure of how much user code the current laxity was letting through.

Issue #1240, opened 2026-09-22
An eval class that adds its own property to a builtin parent cannot be instantiated


Guikingone:
An eval-declared class that extends a builtin (checker-injected) class and adds a property of its own cannot be instantiated: `new` fatals with `eval() runtime failed`. Dropping the property makes the same class construct fine, and the same property over an *AOT user* parent constructs fine — so the failing combination is `own property` × `builtin parent`.

Found while fixing #1124; it is the next wall behind that one for a class with any state, and a different mechanism.

## Repro

```php
<?php
$mode = $argv[1] ?? "builtin_prop";

if ($mode === "builtin_prop") {          // builtin parent + own property, inherited ctor
    eval('class BP1 extends Exception { public int $n = 4; }
          $o = new BP1("m"); echo "n=", $o->n, "|", $o->getMessage(), "|";');
} elseif ($mode === "builtin_prop_ctor") { // builtin parent + own property + own ctor
    eval('class BP2 extends Exception { public int $n = 4; public function __construct() {} }
          $o = new BP2(); echo "n=", $o->n, "|";');
} elseif ($mode === "builtin_no_prop") {   // control: builtin parent, no own property
    eval('class BP3 extends Exception { public function __construct() {} }
          $o = new BP3(); echo "ok|";');
} elseif ($mode === "aot_prop") {          // control: AOT user parent + own property
    eval('class BP4 extends AotBase { public int $m = 1; public function __construct() { $this->m = 3; } }
          $o = new BP4(); echo "m=", $o->m, "|";');
}
echo "\n";

class AotBase { public int $k = 1; public function __construct(int $v) { $this->k = $v; } }
```

## Measured

Host PHP 8.5.10, elephc at `c75c75f15d` plus the #1124 fix (without it every builtin-parent row fatals earlier, at the declaration).

| mode | elephc | PHP |
|---|---|---|
| `builtin_prop` | `Fatal error: eval() runtime failed` | `n=4\|m\|` |
| `builtin_prop_ctor` | `Fatal error: eval() runtime failed` | `n=4\|` |
| `builtin_no_prop` | `ok\|` | `ok\|` |
| `aot_prop` | `m=3\|` | `m=3\|` |

[...]

PR #1247, opened 2026-09-22
fix(codegen): derive method abstractness from the declaration, not from emission

### LINKED ISSUE #1124: An eval-declared class extending Exception cannot be constructed

Guikingone:
Closes #1124.

An eval-declared class could not extend any runtime-backed parent:

```php
<?php
eval('class VErr extends Exception {} $e = new VErr("x");
      echo ($e instanceof Exception) ? "yes" : "no", "|", $e->getMessage(), "|";');
```

PHP 8.5.10 prints `yes|x|`; elephc printed `Fatal error: eval() runtime failed`.

Two corrections to the issue text, both measured:

- The **declaration** fails, not the constructor. `eval('class VErr extends Exception {}')`
  with nothing after it already fataled, where the issue says "the constructor is where it dies".
- It is **any runtime-backed parent**, not just throwables. `Exception`, `RuntimeException`,
  `Error`, `ArrayObject` and `SplStack` all fataled; an AOT user-class parent and
  `implements Countable` were fine.

## Mechanism

`validate_concrete_class_aot_parent_requirements` refuses a concrete eval class that leaves an
abstract parent method unconcretized, reading `EVAL_REFLECTION_MEMBER_FLAG_ABSTRACT` from the
emitted reflection table. Instrumented, `Exception` answered `abstract=true` for eight of its
nine concrete methods — every one but `__construct`.

Those flags come from `eval_reflection_instance_method_flags`, which decided "abstract" as
`!class_info.method_impl_classes.contains_key(method_name)`. That map does not mean that:
`runtime_class_infos` (`src/codegen/runtime_metadata/classes.rs:24`) trims it down to the methods
whose body symbol the EIR backend actually **emitted**, so a concrete method nothing calls loses
its entry. At the emission site `Exception` carried `impl_keys=["__construct"]`.

Declaration-time instrumentation confirmed all nine were inserted (`is_abstract=false
has_body=true`), that neither removal site in `schema/classes/{methods,interfaces}.rs` fires for
them, and that reachability's `prune_class_methods` is not called for these classes here. The
emission trim is the only narrowing.

## Fix

Abstractness is a property of the declaration, so `ClassInfo` records it directly:
`abstract_methods` / `abstract_static_methods` beside the existing `abstract_properties`. Written
at the four sites that already decide it (both `declare_method` halves, both
`validate_interface_method` abstract-class branches), inherited in `inherit_methods` /
`inherit_static_methods`, retained by reachability with the same key sets as `final_methods`, and
never trimmed by emission. The producer tests that set.

## Measured

Host PHP 8.5.10. Every fragment now declares, matching PHP; the first five fataled before.

| eval fragment | before | after | PHP |
|---|---|---|---|
| `class A extends Exception {}` | fatal | declared | declared |
| `class A extends RuntimeException {}` | fatal | declared | declared |
| `class A extends Error {}` | fatal | declared | declared |
| `class A extends ArrayObject {}` | fatal | declared | declared |
| `class A extends SplStack {}` | fatal | declared | declared |
| `class A extends AotUserClass {}` | declared | declared | declared |
| `class A implements Countable {…}` | declared | declared | declared |
| `class A {}` | declared | declared | declared |

Issue #1124's own repro prints `bad address|422|tagged|exception|throwable`, PHP's same string,
and every row of the issue's own table now matches PHP — including the two `throw`/`catch` rows,
which were fataling for the same declaration reason rather than for anything to do with throwing:

| issue #1124 table row | elephc | PHP |
|---|---|---|
| `throw new Exception("x")` / `catch (Exception $e)` | `caught\|ok\|` | `caught\|ok\|` |
| `throw new Exception("x")` / `catch (Exception)` | `caught\|ok\|` | `caught\|ok\|` |
| `$e = new VErr("x")` | `yes\|x\|` | `yes\|x\|` |
| `throw new VErr("x")` / `catch (VErr $e)` | `caught\|ok\|` | `caught\|ok\|` |
| `throw new VErr("x")` / `catch (Exception $e)` | `caught\|ok\|` | `caught\|ok\|` |

## Tests

Four in `tests/codegen/eval.rs`, including the sharpest form of the defect —
`(new ReflectionMethod("Exception", "getMessage"))->isAbstract()` — and the opposite direction,
a genuinely abstract method that must stay abstract. Mutating the instance producer line alone,
back to `!method_impl_classes.contains_key(...)`, fails three of the four.

## What is not proven

The static half is changed for the same reason, but **no fixture covers it**: mutating
`eval_reflection_static_method_flags` alone breaks nothing, and I could not build a program where
an uncalled static loses its entry (`DateTime::createFromFormat`,
`DateTimeImmutable::createFromFormat` and an uncalled user static all still report
`isAbstract()=false` under that mutation). Kept rather than leave one half deriving abstractness
from emission next to the fixed half.

## Left for follow-ups

Five defects were found and measured while narrowing this one. Each is a different mechanism and
has its own repro with controls. Their relationship to this commit, stated exactly:

- #1239 and #1240 were **masked** by #1124 — the declaration fataled before either could be
  reached — so they become visible only once this lands.
- #1241 is independent of it and reproduces either way.
- #1243 was measured with this fix reverted and the matrix is byte-identical.
- #1244 is a checker diagnostic; this commit is in codegen and does not touch that path.

- #1239 — `parent::__construct()` from an eval class into an AOT parent fatals. The same call
  into an eval-declared parent works, and a constructor over an AOT parent that does not call the
  parent works.
- #1240 — an eval class that adds its **own property** on top of a builtin parent cannot be
  instantiated. Without the property it constructs; with the same property over an AOT user
  parent it constructs.
- #1241 — elephc declares all nine `Exception` methods non-final where PHP declares seven
  `final`, so it accepts an override PHP refuses.
- #1243 — `__get` is not dispatched across the eval boundary: an eval read yields `""`, and an
  AOT read of an eval-written property warns `Undefined property`. `__set` and the backing store
  are correct in every case.
- #1244 — `throw` refuses an eval-declared class instance at compile time, even
  `throw new EvalClass("boom")`. Throwing the same class from inside eval works, so this is a
  checker refusal, not a runtime limit.


nahime0:
Follow-ups from leftover gaps after this change:

- #1257 — Lock static `ReflectionMethod::isAbstract` against emission trimming (static flags path is not covered by a revert-style regression today).
- #1258 — Record enum method abstractness in the enum `ClassInfo` builder (`enums.rs` never reads `method.is_abstract`).

Neither blocks this PR.

